PR #91, opened 2014-08-10
Add _setup method

### LINKED ISSUE #86: Adding custom method to service
Hi,
How can I add custom method to service like "count" method? or as loopback does ( http://docs.strongloop.com/display/DOC/Exposing+models+over+a+REST+API ).

Thanks

daffl:
This pull request adds a `service._setup(app, path)` method that allows registering conflicting routes as reported in and closes #86. Works as described in the documentation added:

`setup` will be called after all REST routes have been set up. This means that you
can't register custom middleware with potentially conflicting paths.
In this case, implement `_setup(app, path)` instead which will run before the REST
middleware is registered. An example would be implementing a `todos/count` route
which returns the number of todos (instead of the todo with the id count):

``` js
var todoService = {
  todos: [
    {
      id: 0,
      description: 'Learn Feathers'
    },
    {
      id: 1,
      description: 'Do dishes'
    }
  ],

  find: function (params, callback) {
    callback(null, this.todos);
  },

  _setup: function (app, path) {
    var self = this;

    app.get('/' + path + '/count', function (req, res) {
      self.find({}, function (error, todos) {
        res.json({
          count: todos.length
        });
      });
    });
  }
};

feathers()
  .use('todos', todoService)
  .listen(8000);
```

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -34,6 +34,10 @@ module.exports = {
       fn.call(self, protoService);
     });
 
+    if(typeof protoService._setup === 'function') {
+      protoService._setup(this, location);
+    }
+
     // Run the provider functions to register the service
     _.each(this.providers, function (provider) {
       provider(location, protoService, options);
--- a/readme.md
+++ b/readme.md
@@ -368,6 +368,48 @@ feathers()
 
 You can see the combination when going to `http://localhost:8000/my/test`.
 
+`setup` will be called after all REST routes have been set up. This means that you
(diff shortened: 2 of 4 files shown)


elixiao:
can I get "params" or "params.query" in custom service methods?

daffl:
@elixiao Yes, you always do in any service if you call it through REST or websockets. You will have to pass `query` yourself on the server.

PR #101, opened 2015-01-08
Clone Proto to modify and use the copy.

marshallswain:
As mentioned in feathersjs/feathers-mongodb#9, this creates a clone of Proto to be modified.

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -1,17 +1,17 @@
 'use strict';
 
-var Proto = require('uberproto');
 var _ = require('lodash');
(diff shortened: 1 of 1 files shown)


daffl:
That's a great idea. Tests are failing, probably because they need `Proto.create` now though.

marshallswain:
I don't think this is a good idea.  Even after removing the Proto.isPrototypeOf() test, EVERY test after the first 10 fails.

daffl:
It is the right approach though I think. I'll give it a try as well and see what is happening.

Issue #96, opened 2014-10-18
Should socket calls require a callback?


marshallswain:
This feels like a bug to me.  I've been trying to do a one-off update to the server like this:

```js
socket.emit("api/preferences::update", pref._id, pref, {});
```

Which results in a server crash on this error:

```
/Users/marshallthompson//Sites/MMA/node_modules/feathers/node_modules/rubberduck/lib/rubberduck.js:67
        throw e;
              ^
TypeError: Property 'callback' of object #<Object> is not a function
    at hookCallback (/Users/marshallthompson//Sites/MMA/node_modules/feathers-hooks/lib/before.js:39:35)
    at Object.exports.requireAuth (/Users/marshallthompson//Sites/MMA/server/hook-library.js:30:12)
    at Object.fn (/Users/marshallthompson//Sites/MMA/node_modules/feathers-hooks/lib/before.js:49:32)
    at Object.self.(anonymous function) (/Users/marshallthompson//Sites/MMA/node_modules/feathers/node_modules/uberproto/lib/proto.js:61:21)
    at hookCallback (/Users/marshallthompson//Sites/MMA/node_modules/feathers-hooks/lib/before.js:44:22)
    at Object.exports.setUserID (/Users/marshallthompson//Sites/MMA/server/hook-library.js:120:10)
    at Object.fn (/Users/marshallthompson//Sites/MMA/node_modules/feathers-hooks/lib/before.js:49:32)
    at Object.self.(anonymous function) (/Users/marshallthompson//Sites/MMA/node_modules/feathers/node_modules/uberproto/lib/proto.js:61:21)
    at Object.wrapped [as update] (/Users/marshallthompson//Sites/MMA/node_modules/feathers/node_modules/rubberduck/lib/rubberduck.js:63:12)
    at Socket.<anonymous> (/Users/marshallthompson//Sites/MMA/node_modules/feathers/lib/providers/socket/commons.js:28:23)
```

If I put the callback in place like this:

```js 
socket.emit('api/preferences::update', pref._id, pref, {}, function(error, data) {
	console.log(data);
});
```

It works as expected.  That doesn't seem right to me.  Do socket calls always require the callback in place?

marshallswain:
This is fixed in #102.

Issue #82, opened 2014-06-14
Dynamic services


grendo:
Hi,
  like the library did not suit exactly what I wanted.  I want exactly the realtime/rest interface you have but want to be able to set up services at runtime.  As you are using app.use this is not possible this without restarting the server. 

Not a bug but just a feature I would find useful.

Issue #67, opened 2014-05-31
setup service method not called in async plugin


ksmth:
Even though the `setup` method isn't called, the services work properly.

```JavaScript
// app.js
require('feathers')()
    .configure(function () {
        var app = this;
        app.use('/setup', {
            get : function (id, params, callback) {
                callback(null, {});
            },
            setup : function (app) {
                console.log('setup');
            }
        });

        setTimeout(function () {
            app.use('/nosetup', {
                get : function (id, params, callback) {
                    callback(null, {});
                },
                setup : function (app) {
                    console.log('no setup :(');
                }
            });
        }, 100);
    })
    .listen(8080);
```

```
$ node app.js
setup
```

PR #102, opened 2015-01-09
Dynamic services

marshallswain:
Creates the ability to add services at runtime.  This is going to take a bit more effort to test out properly, but I wanted to get it out   in the wild to get some feedback on it while I'm in the process of further testing.  Maybe you'll see something that I haven't noticed, yet.  I'll report back what I find.

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -48,7 +48,7 @@ module.exports = {
     });
 
     this.services[location] = protoService;
-    return this;
+    return protoService;
   },
 
   use: function () {
@@ -76,7 +76,7 @@ module.exports = {
       this.addService(svc, location);
     }
 
-    return svc;
+    return this;
   },
 
   setup: function() {
--- a/lib/application.js
+++ b/lib/application.js
@@ -73,7 +73,7 @@ module.exports = {
 
     // If already _setup, just add this single service.
     if (this._setup) {
-      this.addService(svc, location);
+      this.addService(svc, stripSlashes(location));
     }
 
     return this;
--- a/lib/application.js
+++ b/lib/application.js
@@ -1,7 +1,7 @@
 'use strict';
 
 var _ = require('lodash');
-var Proto = _.clone(require('uberproto'));
+var Proto = require('uberproto');
 // We do not want to support Uberproto's create functionality
 // Since our service methods have a method with the same name
 Proto._create = Proto.create;
--- a/lib/application.js
+++ b/lib/application.js
@@ -1,11 +1,7 @@
 'use strict';
 
 var _ = require('lodash');
-var Proto = require('uberproto');
-// We do not want to support Uberproto's create functionality
-// Since our service methods have a method with the same name
-Proto._create = Proto.create;
-delete Proto.create;
+var Proto = require('./uberproto');
 
 var mixins = require('./mixins');
 var stripSlashes = function (name) {
--- a/lib/application.js
+++ b/lib/application.js
@@ -1,17 +1,17 @@
 'use strict';
 
-var Proto = require('uberproto');
 var _ = require('lodash');
+var Proto = _.clone(require('uberproto'));
+// We do not want to support Uberproto's create functionality
+// Since our service methods have a method with the same name
+Proto._create = Proto.create;
+delete Proto.create;
+
 
 var mixins = require('./mixins');
 var stripSlashes = function (name) {
   return name.replace(/^\/|\/$/g, '');
 };
-
-// We do not want to support Uberproto's create functionality
-// Since our service methods have a method with the same name
-Proto._create = Proto.create;
-delete Proto.create;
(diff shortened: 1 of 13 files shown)


marshallswain:
Oops.  I left that other commit in there.  I'll look into the test for that to see what's going on.

marshallswain:
@daffl can you help me understand what's going on in this test?  There's only one argument passed to Proto.isPrototypeOf().  What is being compared here? https://github.com/feathersjs/feathers/blob/master/test/application.test.js#L47

and also here:
https://github.com/feathersjs/feathers/blob/master/lib/application.js#L32
How/what is Proto extending when there's only one argument?

marshallswain:
Heh.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/isPrototypeOf

No matter how much time passes, every day I discover that I'm still an amateur.

marshallswain:
OK.  It looks like that test is failing because when you clone Uberproto it breaks the prototype "link" so that isPrototypeOf no longer works.  
```js

var _ = require('lodash');

// Clone Uberproto
var Proto = _.clone(require('uberproto'));
var dummyService = {
  create: function (data, params, callback) {
    callback(null, data);
  }
};

dummyService = Proto.extend(dummyService);
console.log(Proto.isPrototypeOf(dummyService)); // true

// Clone again and compare
var Proto = _.clone(require('uberproto'));
console.log(Proto.isPrototypeOf(dummyService)); // false

```
That test won't work after cloning Uberproto.

marshallswain:
CRUD tests are passing.  Event-related tests are all failing.

Almost there.

marshallswain:
Victory!  @daffl, we now have dynamic services.  This takes care of #82.

marshallswain:
I just noticed this comment: https://github.com/feathersjs/feathers/issues/67#issuecomment-44757986
>The problem with setting up services right away when calling .use() is that they won't be able to use any other service that has been registered later on.

I added a test for looking up a service added after listen().

marshallswain:
Those last two commits fix #96.  The server will no longer crash when a socket client fails to pass a callback.  It basically creates a dummy callback function that does nothing unless there is an error, in which case it will log the error to the server console.  That should help the developer to not pull as much hair out... if they have access to the server logs.

daffl:
This looks great, thank you! I will have a close look and then merge it. Now I just have to somehow figure out the Proto._create problem. But that should be solvable.

marshallswain:
Here's another attempt at a fix that actually passes all of the tests.  I put a custom version of uberproto.js in the libs directory and have replaced all require() references with a reference to the custom one.

marshallswain:
I just don't know what effect this will have on feathers-[service] modules, yet.

marshallswain:
I simplified the file by removing the extra factory function, et al.  Other than that, I renamed the create to _create and added a comment to the top... and a JSHint directive to ignore making functions inside loops.

I did some testing.  If you make two **completely identical** copies of uberproto.js and extend an object with one, the extended object knows its parent. (Mommy issues or something, I guess.)
```js
var Proto = require('./uberproto');
var proto = require('uberproto');
var cage = {
  animal: 'moose'
};
var moose = Proto.extend(cage);
console.log(proto.isPrototypeOf(moose)); // false
console.log(Proto.isPrototypeOf(moose)); // true
```
Maybe it tracks the register in memory that contains the original?  I don't remember.  Anyway, other services will have to figure out a solution... unless we put a uberproto-feathers module in NPM.  It shouldn't be much of an issue, other than in tests.  Actually, to solve that and maybe improve the quality of custom feathers services, maybe we could put the tests in their own NPM module.  We could then create an NPM module that's ready to test.

marshallswain:
Woot!  I'm on the team!  

Ok.  I'm super tempted to merge this, but for the sake of checks and balances, I'd prefer that somebody else does it.

daffl:
Actually first, could you push to a branch on this repository so that I can check it out?

marshallswain:
Done.  https://github.com/feathersjs/feathers/tree/dynamic-services

ekryski:
@marshallswain :boom: nice work man. That is suweet!

ekryski:
@daffl and I have been talking about creating a test harness for the crud adapters for a while. I'll cook something up this week. I have a bit more free time now.

daffl:
I didn't forget about this just didn't have the time to go over everything yet, sorry.

daffl:
Allright, I think #104 and #105 should make for a new patch release which I think also caused some issues in this PR. @marshallswain Could you clean this one up one more time when those are in?

I think it looks good besides that but maybe we can discuss if there is a way to consolidate functionality even more.

marshallswain:
@daffl @ekryski Done.  I've created a cleaner PR for dynamic services in #107.

PR #104, opened 2015-02-04
Make sure that mixins are specific to each new app

daffl:
This pull request fixes a bug that cause weird issues. The reason was that an applications `mixins` array was shared through *every* Feathers application.


daffl:
Looks like this depends on #105 - which has to be merged first - for the tests to pass.

PR #105, opened 2015-02-04
Use Uberproto extended instance when creating services

daffl:
This pull request fixes the Uberproto hack from #100 and corrects the promise mixin to only work with functions.


ekryski:
:shipit:

PR #107, opened 2015-02-04
Dynamic services 2

### LINKED ISSUE #67: setup service method not called in async plugin

### LINKED ISSUE #82: Dynamic services

marshallswain:
This is Dynamic Services on top of the recent 1.0.2 patch.  I've also made it no longer a requirement to pass a callback (#96).  If the request fails, it logs a message to the console for a developer to track down.

As for consolidating the code a bit more.  This is basically creating a second path in the code to setup a service.  The first way is through the setup methods, and now this for sure works after the server has started.  I'm not sure if it would be possible to just use this new way BEFORE the server starts.  I would have to check to see if there is anything that the setup methods put in place that the services require as prerequisites.

**Update:** We for sure have to wait until the providers are set in place, which isn't really guaranteed until after the server starts (since providers are setup with configure).  There may be other optimizations that can be made?

Closes #67 and closes #82

### DIFF
--- a/lib/providers/socket/commons.js
+++ b/lib/providers/socket/commons.js
@@ -1,6 +1,7 @@
 'use strict';
 
 var _ = require('lodash');
+var eventsMixin = require('../../mixins/event').Mixin;
 
 // The position of the params parameters for a service method so that we can extend them
 // default is 1
@@ -10,8 +11,19 @@ var paramsPositions = {
   patch: 2
 };
 
-// Set up the service method handlers for a service and socket.
-exports.setupMethodHandler = function setupMethodHandler (emitter, params, service, path, method) {
+exports.addService = function addService(service, path){
+  // Add handlers for the service to connected sockets.
+  _.each(this.info.connections, function (spark) {
+    this.setupMethodHandlers(service, path, spark);
+  }, this);
+
+  // Setup events for the service.
+  eventsMixin.applyEvents.call(service);
+  exports.setupEventHandlers.call(this, service, path);
+};
+
+// Set up the service method handler for a service and socket.
+exports.setupMethodHandler = function setupMethodHandler(emitter, params, service, path, method) {
   var name = path + '::' + method;
   var position = typeof paramsPositions[method] !== 'undefined' ? paramsPositions[method] : 1;
 
@@ -26,6 +38,15 @@ exports.setupMethodHandler = function setupMethodHandler (emitter, params, servi
       args[position] = _.extend({ query: args[position] }, params);
       service[method].apply(service, args);
     });
+  }
+};
+
+exports.setupEventHandlers = function setupEventHandlers(service, path){
+  // If the service emits events that we want to listen to (Event mixin)
+  if (typeof service.on === 'function' && service._serviceEvents) {
+    _.each(service._serviceEvents, function (ev) {
+      exports.setupEventHandler(this.info, service, path, ev);
+    }, this);
   }
 };
 
@@ -54,4 +75,3 @@ exports.setupEventHandler = function setupEventHandler (info, service, path, ev)
     });
   });
 };
-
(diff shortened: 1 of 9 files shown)


marshallswain:
For some reason I CAN create a branch on the main Feathers repo, but I'm not able to push to it, so there's a straggling branch called dynamic-services-2 hanging out right now.

daffl:
I just double checked and you should have push access to this repository. Did you use `git@github.com:feathersjs/feathers.git`?

marshallswain:
Ahh.  I enabled two-factor auth.  I forgot to use the personal access token.  I pushed the branch.

marshallswain:
>I think it looks good besides that but maybe we can discuss if there is a way to consolidate functionality even more.

@daffl What did you have in mind?

lib/providers/socket/commons.js
```diff
@@ -24,6 +24,19 @@ exports.setupMethodHandler = function setupMethodHandler (emitter, params, servi
         args.splice(position, 0, {});
       }
       args[position] = _.extend({ query: args[position] }, params);
+
+      // If there was no callback supplied by a socket client...
```
daffl: Can we put this into a separate PR since it belongs to some different functionality?

daffl:
Well it looks that for example the code from https://github.com/feathersjs/feathers/pull/107/files#diff-a89e33317f0c782705345838f29cf003R57 and https://github.com/feathersjs/feathers/pull/107/files#diff-dbb6eace96490fa4b721a67e6fd9340dR57 can be maybe consolidated into https://github.com/feathersjs/feathers/blob/master/lib/providers/socket/commons.js.

I'm also wondering if we could somehow hook into [app.service](https://github.com/feathersjs/feathers/blob/master/lib/application.js#L24) and make sure that the provider methods can be called at any time? I think you a currently doing this in `.use` (at https://github.com/feathersjs/feathers/pull/107/files#diff-5372f626ee15242f1e2c6eb31655b4faR73) right?

marshallswain:
OK.  I moved most of it into the commons.  You'll notice that in the `this.info` part of each provider I put an attribute called connections, which is used in commons.js [here](https://github.com/feathersjs/feathers/blob/dynamic-services-2/lib/providers/socket/commons.js#L16).  I put it in `this.info` because I didn't want to pollute the `app`'s global space. hmm... Actually, the way it is set up, `this.info` gets mixed into the `app`, now.  That makes it so that Socket.io can't be used at the same time as Primus.  That's not really a concern, is it?

I left the `setupMethodHandlers()` method in each one since the methodHandlers depend on the connected socket and use different parameter names.  It can be done, but it felt a little hacky moving it into commons.js.

marshallswain:
Oh.  I also took that other commit regarding socket callbacks out.

marshallswain:
Oops.  I somehow got it back in there.  Let me fix that.

marshallswain:
I'll make a PR for it after we get this one merged.

daffl:
Sweet! I will check it out tonight and then merge it in. This will make for a great 1.1 release. Debating if [ZeroRPC](http://zerorpc.dotcloud.com/) support would be something worth considering for that, too (and writing up how use this as microservices).

Oh and I'd like to polish the website a little, too.

daffl:
Merged! Great job!

PR #110, opened 2015-02-12
Make sure .setup() runs on dynamic services.

marshallswain:
Setup() wasn't being run on dynamic services.  I've included a fix and updated one of the tests to make sure setup runs.

Oh, I also moved `stripSlashes(location)` to the top of app.service() and replaced all duplicate instances of it with `location`.

PR #115, opened 2015-02-17
Fix duplicate events in dynamic services.

marshallswain:
After fixing setup() for dynamic services in the last PR (#110), the events mixin's setup() function was adding duplicate events.  This PR wraps `applyEvents` with an `if (!this._serviceEvents) { ... }` check to assure it isn't adding events twice.

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -45,13 +45,9 @@ module.exports = {
       provider(location, protoService, options || {});
     });
 
-    // If already _setup, just add this single service.
-    if (this._setup) {
+    // If we ran setup already, set this service up explicitly
+    if (this._isSetup) {
       protoService.setup(this, location);
-      // If we're using a socket provider, register the service on it.
-      if (this.addService) {
-        this.addService(protoService, location);
-      }
     }
 
     this.services[location] = protoService;
@@ -89,7 +85,7 @@ module.exports = {
       }
     }.bind(this));
 
-    this._setup = true;
+    this._isSetup = true;
 
     return this;
   },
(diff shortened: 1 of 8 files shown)


daffl:
Are you sure this is the best solution? I would think we just don't need to call `applyEvents` anymore since it will be mixed in and called with `.setup`, no?

marshallswain:
Yeah.  Alternatively we can remove applyEvents and move this line: https://github.com/feathersjs/feathers/blob/master/lib/application.js#L54
just before this.addService() (a few lines up).   

You think that's a better solution?

marshallswain:
See #116.

daffl:
You could just update this PR by pushing to the `fix-duplicate-events` branch again instead of opening a new one. Or was there a reason?

marshallswain:
Just so you could see the diff.  Not trying to insult your intelligence.  :)

marshallswain:
I was thinking of leaving it the other way in preparation for allowing providers to be registered at any time and making a single pathway for registering services instead of the duality going on right now with setup() and addService().  It's not that big of a change though, so it doesn't really matter.

daffl:
I will have another look. I feel that there there is a good single way of doing this. Maybe you are right and mixins are not the way to go here.

daffl:
I did some refactoring in https://github.com/feathersjs/feathers/blob/ffb603a4d8be2b309f26ad2e449611f8f94f9993/lib/providers/socket/commons.js that should handle both kinds of services properly and also made the socket handlers a little shorter and moved the dynamic service tests into nested describes. Can you double check if that solves all the issues and also makes sense?

marshallswain:
Ahh...  Yes.  That is so much cleaner and solves the issue.

Issue #113, opened 2015-02-15
Server hardening with socket message validation.


marshallswain:
Working on #112 & #96, I realized that it is **really easy** to take the server down by sending a malformed message.  It's not very resilient.  (and we don't want to test the limits of the [forever module](https://github.com/foreverjs/forever-monitor) )

Really, Feathers shouldn't be responsible for a lot of validation, but it should do anything it can to keep itself running.  It seems to me like we ought to at least validate the number of arguments as well as the presence (& maybe lightly the type) of each argument.

```js
// Service methods arranged by number of arguments
var myService = {
  find:   function(params, callback) {},
  get:    function(id, params, callback) {},
  remove: function(id, params, callback) {},
  create: function(data, params, callback) {},
  update: function(id, data, params, callback) {},
  patch:  function(id, data, params, callback) {}
};
```

If we do validate the type of each argument, is there any case where id would be anything other than a string or number?  MongoDB usually uses ObjectIDs, but that's something that the MongoDB service would need to validate.  I can't imagine a need for using an object as an id, but maybe I need a better imagination.  I think we're already checking for the presence of at least id in other places.  We do it in [utils.js](https://github.com/feathersjs/feathers-hooks/blob/75afb12b72fdc412f9cc2d8d2da788c16840ac29/lib/utils.js) `feathers-hooks` and it seems like I've seen something like this elsewhere.

The right place to be handling this is probably in [commons.js](https://github.com/feathersjs/feathers/blob/master/lib/providers/socket/commons.js#L35). We are already creating a `params` object if it's not provided. Combined with #112, which does the same for the `callback`, I think all that's left is validating `id` and `data` as well as the number of arguments.  

I think that, ideally, if the request is malformed, we would immediately send back an `error 400 / Bad Request`.  Developers making services wouldn't have to worry about handling the same basic validation, but can focus on validation unique to the module they're developing.

@feathersjs/contributors While we're on the topic, maybe you've run into other places that the server could be hardened a bit.  Let me know what you think.

daffl:
That is true, you just have to emit some invalid data to make the server crash. The best place is probably in the socket argument validation at https://github.com/feathersjs/feathers/blob/master/lib/providers/socket/commons.js#L31. We want to make sure that there always is the minimum of the required parameters for Socket calls.

I don't think this is a problem for REST calls since they'll always look the same.

marshallswain:
[...]

PR #120, opened 2015-02-27
Always run stripSlashes() in use().

### LINKED ISSUE #119: Regression in location/path name with latest refactor (for sockets)
We now have two `service` methods.  
One here: https://github.com/feathersjs/feathers/blob/master/lib/application.js#L25
The other gets mixed in here: (or in primus.js): https://github.com/feathersjs/feathers/blob/master/lib/providers/socket/socketio.js#L16

`stripSlashes()` is only called in the first one.  Since the second `service` method gets called directly from `app.use()`, is the proper fix to simply add a `stripSlashes()` here: https://github.com/feathersjs/feathers/blob/master/lib/application.js#L76

marshallswain:
This fixes #119.  Paths weren't being escaped inside the socket commons.js.  In `application.js`, Running `stripSlashes()` inside both `use()` and `service()` fixes the issue.  I thought it best to go here so the individual providers don't need to do it.

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -22,17 +22,17 @@ module.exports = {
     });
   },
 
-  service: function(location, service, options) {
-    location = stripSlashes(location);
+  service: function(path, service, options) {
+    path = stripSlashes(path);
 
     if(!service) {
-      return this.services[location];
+      return this.services[path];
     }
 
     var protoService = Proto.extend(service);
     var self = this;
 
-    debug('Registering new service at `' + location + '`');
+    debug('Registering new service at `' + path + '`');
 
     // Add all the mixins
     _.each(this.mixins, function (fn) {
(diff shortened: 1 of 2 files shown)


marshallswain:
For consistency, I've renamed all instances of `location` to `path`, instead of having two names for the exact same variable contents.  I've also modified the test for leading/trailing slashes to use the socketio plugin, so it should catch this if ever in the future...

If everything looks good I'll squash/merge.

marshallswain:
Actually, that test modification is useless.  The problem only happens in the socket provider when the events get fired with the leading/trailing slashes intact.  I'll come up with something else when I get another minute.

daffl:
Good catch though. I'm wondering if this is the right time to introduce a set of utility functions (e.g. `app.feathers.stripSlashes`). I'm thinking we might need those for argument normalization (see #113) anyway (like `var args = app.feathers.getArguments('create', arguments)`)

daffl:
Fixed via https://github.com/feathersjs/feathers-commons/commit/f05b9502160d2c3c7326222027956d1cedf54476

PR #123, opened 2015-03-06
Tests for socket message validation and errors.

### LINKED ISSUE #96: Should socket calls require a callback?

### LINKED ISSUE #113: Server hardening with socket message validation.

daffl:
This pull request adds tests for the latest [feathers-commons](https://github.com/feathersjs/feathers-commons) functionality which allows Socket message validation and to omit parameters and calls back with an error when wrong arguments are submitted. Closes #96 and closes #113. /cc @marshallswain

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: node_js
 node_js:
   - "0.10"
-  - "0.11"
+  - "node"
+  - "iojs"
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "lib": "lib"
   },
   "scripts": {
-    "test": "grunt"
+    "test": "grunt --stack"
   },
   "engines": {
     "node": ">= 0.10.0",
@@ -37,8 +37,8 @@
     "body-parser": "^1.0.2",
     "debug": "^2.1.1",
     "express": "^4.0.0",
-    "feathers-commons": "^0.1.0",
-    "feathers-errors": ">=0.2.0",
+    "feathers-commons": "^0.2.0",
+    "feathers-errors": "^0.2.5",
     "lodash": "^3.1.0",
     "primus": "^2.4.0",
     "primus-emitter": "^3.0.2",
--- a/test/providers/primus.test.js
+++ b/test/providers/primus.test.js
@@ -93,8 +93,15 @@ describe('Primus provider', function () {
   });
 
   describe('Services', function() {
+    it('invalid arguments cause an error', function (done) {
+      socket.send('todo::find', 1, {}, function(error) {
+        assert.equal(error.message, 'Too many arguments for \'find\' service method');
+        done();
+      });
+    });
 
     describe('CRUD', function () {
+
       it('::find', function (done) {
         socket.send('todo::find', {}, function (error, data) {
           verify.find(data);
@@ -120,6 +127,20 @@ describe('Primus provider', function () {
           verify.create(original, data);
 
           done(error);
+        });
+      });
+
+      it('::create without parameters and callback', function (done) {
+        var original = {
+          name: 'creating'
+        };
+
+        socket.send('todo::create', original);
+
+        socket.once('todo created', function(data) {
+          verify.create(original, data);
+
+          done();
         });
       });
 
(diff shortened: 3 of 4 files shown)


ekryski:
Looks all good!

marshallswain:
I love the validation!  It's good to see the server still running after doing something stupid.

daffl:
I might refactor this as a general mixin that doesn't just do this for sockets but for all service method calls all the time (so you can also use this with `app.service`, hooks etc.).

PR #124, opened 2015-03-07
Service method call normalization

daffl:
#123 added method normalization to Socket calls. As discussed, this pull request instead adds a mixin that always normalizes all method calls. It will always be added as the last mixin (and therefore the first to run) which also makes sure that plugins like [feathers-hooks](https://github.com/feathersjs/feathers-hooks) won't have to worry about normalizing the parameters.

### DIFF
--- a/lib/mixins/index.js
+++ b/lib/mixins/index.js
@@ -3,6 +3,7 @@
 module.exports = function() {
   return [
     require('./promise'),
-    require('./event')
+    require('./event'),
+    require('./normalizer')
   ];
 };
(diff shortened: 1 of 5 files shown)


daffl:
This means you can now always do things like:

```js
app.use('/todos', {
  update: function(id, data, params, callback) {
    callback({
      id: id,
      data: data,
      params: params
    })
  }
});

var todoService = app.service('todos');

// Call `update` with just id, callback and data
todoService.update(10, { type: 'new data' }, function() {});

// Call `update` with just id and data
app.service('todos').update(10, { type: 'new data' });
``

PR #141, opened 2015-07-24
Handle advanced rest routes for collection in resource

ebourmalo:
Handle new routes for updating a collection within a resource:

- Find an item within the collection:
[GET] /:id/:collection 
--> service.findInCollection(id, collection, params, cb)

- Get an item of the collection
[GET] /:id/:collection/:documentId 
--> service.getInCollection(id, collection, documentId, params, cb)

- Add an item to the collection
[POST] /:id/:collection 
--> service.addToCollection(id, collection, data, params, cb)

- Remove an item from the collection
[DELETE] /:id/:collection/:documentId 
--> service.removeFromCollection(id, collection, documentId, params, cb)

Linked with these PRs:
- feathers-commons: https://github.com/feathersjs/feathers-commons/pull/2
- feathers-hooks: https://github.com/feathersjs/feathers-hooks/pull/21

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
   "dependencies": {
     "debug": "^2.1.1",
     "express": "^4.12.3",
-    "feathers-commons": "^0.2.0",
+    "feathers-commons": "ebourmalo/feathers-commons",
     "feathers-errors": "^0.2.5",
     "lodash": "^3.1.0",
     "primus": "^3.0.2",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "feathers",
   "description": "Build Better APIs, Faster than Ever.",
-  "version": "1.1.0",
+  "version": "1.1.1",
   "homepage": "http://feathersjs.com",
   "repository": {
     "type": "git",
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
   "dependencies": {
     "debug": "^2.1.1",
     "express": "^4.12.3",
-    "feathers-commons": "ebourmalo/feathers-commons",
+    "feathers-commons": "ebourmalo/feathers-commons#feat-advanced-routes",
     "feathers-errors": "^0.2.5",
     "lodash": "^3.1.0",
     "primus": "^3.0.2",
--- a/lib/application.js
+++ b/lib/application.js
@@ -12,7 +12,9 @@ var mixins = require('./mixins');
 module.exports = {
   init: function () {
     _.extend(this, {
-      methods: ['find', 'get', 'create', 'update', 'patch', 'remove'],
+      methods: ['find', 'get', 'create', 'update', 'patch',
+        'remove', 'findInCollection', 'addToCollection',
+        'getInCollection', 'removeFromCollection'],
       mixins: mixins(),
       services: {},
       providers: [],
--- a/lib/mixins/event.js
+++ b/lib/mixins/event.js
@@ -7,8 +7,11 @@ var eventMappings = {
   create: 'created',
   update: 'updated',
   remove: 'removed',
-  patch: 'patched'
+  patch: 'patched',
+  addToCollection: 'updated',
+  removeFromCollection: 'updated'
 };
+var debug = require('debug')('feathers:event');
 
 /**
  * An event mixin that emits events after a service method calls its callback.
@@ -31,15 +34,23 @@ var EventMixin = {
     _.each(eventMappings, function (event, method) {
       var alreadyEmits = self._serviceEvents.indexOf(event) !== -1;
 
-      if (typeof self[method] === 'function' && !alreadyEmits) {
+      // dont'check already emits because new collection methods
+      // (addToCollection and removeFromCollection) also emits an 'update' event
+      if (typeof self[method] === 'function') {
         // The Rubberduck event name (e.g. afterCreate, afterUpdate or afterDestroy)
         var eventName = 'after' + method.charAt(0).toUpperCase() + method.substring(1);
+
+        debug('event propagate:' + eventName);
+
         self._serviceEvents.push(event);
         // Punch the given method
         emitter.punch(method, -1);
         // Pass the event and error event through
         emitter.on(eventName, function (results) {
           if (!results[0]) { // callback without error
+
+            debug('Emit:' + eventName);
+
             self.emit(event, results[1]);
           } else {
             self.emit('serviceError', results[0]);
--- a/lib/providers/rest/index.js
+++ b/lib/providers/rest/index.js
@@ -45,16 +45,16 @@ module.exports = function (config) {
 
       debug('Adding REST provider for service `' + path + '` at base route `' + uri + '`');
 
-      // GET / -> service.find(cb, params)
+      // GET / -> service.find(params, cb)
       baseRoute.get.apply(baseRoute, middleware.concat(app.rest.find(service)));
-      // POST -> service.create(data, params, cb)
+      // POST / -> service.create(data, params, cb)
       baseRoute.post.apply(baseRoute, middleware.concat(app.rest.create(service)));
 
       // GET /:id -> service.get(id, params, cb)
       idRoute.get.apply(idRoute, middleware.concat(app.rest.get(service)));
       // PUT /:id -> service.update(id, data, params, cb)
       idRoute.put.apply(idRoute, middleware.concat(app.rest.update(service)));
-      // PATCH /:id -> service.patch(id, data, params, callback)
+      // PATCH /:id -> service.patch(id, data, params, cb)
       idRoute.patch.apply(idRoute, middleware.concat(app.rest.patch(service)));
       // DELETE /:id -> service.remove(id, params, cb)
       idRoute.delete.apply(idRoute, middleware.concat(app.rest.remove(service)));
--- a/lib/providers/rest/index.js
+++ b/lib/providers/rest/index.js
@@ -61,6 +61,20 @@ module.exports = function (config) {
       // DELETE /:id -> service.remove(id, params, cb)
       idRoute.delete.apply(idRoute, middleware.concat(app.rest.remove(service)));
 
+      // GET /:id/:collection -> service.findInCollection(id, collection, params, cb)
+      collectionRoute.get.apply(collectionRoute,
+        middleware.concat(app.rest.findInCollection(service)));
+      // POST /:id/:collection -> service.addToCollection(id, collection, params, cb)
+      collectionItemRoute.post.apply(collectionItemRoute,
+        middleware.concat(app.rest.addToCollection(service)));
+
+      // GET /:id/:collection/:itemId -> service.getInCollection(id, collection, itemId, params, cb)
+      collectionItemRoute.get.apply(collectionItemRoute,
+        middleware.concat(app.rest.getInCollection(service)));
+      // DELETE /:id/:collection/:itemId -> service.getInCollection(id, collection, itemId, params, cb)
+      collectionItemRoute.delete.apply(collectionItemRoute,
+        middleware.concat(app.rest.removeFromCollection(service)));
+
       app.use(uri, handler);
     });
   };
--- a/lib/providers/rest/index.js
+++ b/lib/providers/rest/index.js
@@ -64,8 +64,8 @@ module.exports = function (config) {
       // GET /:id/:collection -> service.findInCollection(id, collection, params, cb)
       collectionRoute.get.apply(collectionRoute,
         middleware.concat(app.rest.findInCollection(service)));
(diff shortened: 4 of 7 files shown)


package.json
```diff
@@ -43,7 +43,7 @@
     "body-parser": "^1.0.2",
     "debug": "^2.1.1",
     "express": "^4.12.3",
-    "feathers-commons": "^0.2.0",
+    "feathers-commons": "ebourmalo/feathers-commons#feat-advanced-routes",
```
ebourmalo: Should be replaced with `feathers-commons": "^0.3.0"` if the corresponding PR is validated and merged (https://github.com/feathersjs/feathers-commons/pull/2)

package.json
```diff
@@ -42,7 +42,7 @@
   "dependencies": {
     "debug": "^2.1.1",
     "express": "^4.12.3",
-    "feathers-commons": "^0.2.0",
+    "feathers-commons": "ebourmalo/feathers-commons#feat-advanced-routes",
```
ebourmalo: Should be replaced with `feathers-commons": "^0.3.0"` if the corresponding PR is validated and merged (https://github.com/feathersjs/feathers-commons/pull/2)

daffl:
Hi Emmanuel!

Thank you so much for contributing! I think URL routing is one of my few pet-peeves (as in that it is very easy to make it too complicated - so I apologize if I'm a stickler about it ;). Here are some of my questions:

0) Could you elaborate a little more on the use cases? I just realized that there really should be a way to update entire resource collections (e.g. PATCH /todos) but it looks like your routes are supposed to handle something different.

1) What is the difference to implementing a service like

```js
var service = {
  removeFromCollection(id, collection, documentId, params, callback) {

  },

  remove: function(documentId, params, callback) {
    var collection = params.collection;
    var id = params.mainId;

    if(collection && documentId) {
      this.removeFromCollection(id, collection, documentId, params, callback);
    }
  }
};

app.use('/todos/:mainId/:collection', service);
```

2) Could it be implemented as a plugin? We're always careful to add anything to core if it can also be implemented as a plugin. Maybe something that allows us to alias parameters to other service methods?

3) From a quick look at the code I think the methods can be called via websockets but does it send the `created`, `removed` etc events?

ekryski:
@ebourmalo Yes thanks for the awesome PRs! I agree with @daffl here. I'd probably prefer to see this as a plugin. Something like `feathers-collections`. However could you please elaborate on the use case? 

You should already be able to modify resources in a collection with a flatter route structure. For example, let’s say it’s posts and comments. Instead of  `/posts/:id/comments` you just do `/comments?postId=:id`. 

I realize that most frameworks out there support deeply nested routes, and we have seen this come up a few times with Feathers however @daffl and I have realized that you don't need them. We may totally be wrong but we'll need a bit of convincing :wink:

ebourmalo:
Thank you for you feedbacks :) I really trust on this library so I'm really motivated for helping you on that work (if you allow me to naturally :)

@daffl 
0) Let's take a concrete use case: there is a resource `team` containing `name`, `description` and `users`. 

With the existing routes, you can automatically find, get, update and delete a team. If we focus on `users` (collection of resource user), we can only update the hole thing: if we want to add just one user, we have to pass all the existing users + the new one.

With these new routes, we can automatically for any collection in a resource, find, get, add or remove an item from that collection. Let's consider each case:

- GET  /users/1/projects : Find all projects (or filtered projects if params) the user 1 is linked to.
- GET /users/1/projects/123 : Check if the user 1 is linked to the project 123 (return the project if OK, not found otherwise).
- POST /users/1/projects/ : Link the user 1 to a new project. The project can either be an existing project (we'll find an id in the passed data) or a new one (no id in the passed data). In the later case, we can create a new record in the project resource before linking the user to it.
- DELETE /users/1/projects/123 : Unlink the user 1 to the project 123.

1) I didn't think about it, that's a good idea but: 
- you'll have to custom and 'complicate' routes as you did, I think (IMO) the point of feathersjs is to stay simple, light and clear as it is: `app.use('/todos', TodoService);`
- you can do it if you create your service manually, not if you use a plugin for generating your service automatically (i'll come back on that point just after)

2) I think it could be done as a plugin, do you? I think these "advanced" routes are really requested on an API and it's not wrong to me to integrate it to the core but it can also be seen as a plugin. I'm not sure if it is possible as it needs modifications on 3 modules (feathers, feathers-hooks, feathers-commons), you may have a better overview of the feasibility?

3) For the websocket thing and the events, I first wanted your feedbacks :) So right now, it's not (yet) implemented.

@ekryski 
Your example with comments and posts could be write in the only case they are 2-way linked (`posts` have a collection `comments` and `comments` have a property `post`). It doesn't work if the comment doesn't have the property `post` (doesn't make lots of sense in this case but it was for keeping the same idea). 

I think it's really convenient to update a collection in the context of a resource so: 
- we don't have to retrieve all the data and then just select the requested collection
- we avoid to develop the same blocks of code for repeating the basics operations for every collection in a resource
- we can easily and with the same input add an item to the collection, whether this item already exist (an existing project with an id for instance) or a brand new one that will be created and added to the collection.


Now, for exposing my real use case on top of that:
The main idea behind that is for using feathers with a mongoose plugin. I developed one here: https://github.com/ebourmalo/feathers-mongoose-rest because I couldn't make the original one (feathers-mongoose) work (pb of dependencies versions and some other extra pbs) :( 

From my existing mongoose models, I can expose them directly on my API automatically, with 3 lines of code. It's very convenient to just declare `app.use('/projects', ProjectService);` and to see the resource exposed, even with its different collection, without having a thing to do. 

And then, the use case of the collection: in my resource `project`, I have `users`. At the very beginning, without the new routes, I had to update the whole collection of users in the project, do a diff process for figuring out the new users to add or remove, and then update it. No easy way for adding or removing a user to/from a project. I think that's because this kind of problems that Sails added `add to` and `remove from` methods natively.

IMO, that's pretty common to have these behaviours on an API and would be awesome to integrate it to the core but I can totally understand your point of view for proposing it as a plugin.

What do you think about it?

daffl:
It sounds similar to what [feathers-association](https://github.com/feathersjs/feathers-associations) is supposed to do. The reason why I'd rather see this as a plugin as opposed to core is because it is a rather specific use-case. In most relational databases you would have a `/users` and `/projects` service (and DB table) which is where you'd do what @ekryski suggested or pull it in through associations.

The only time this isn't the case is in document oriented database where you can embed collections that do not have their own service within other resources (at least that's what I understood from your explanation).

Would it be ok to move this functionality into a plugin or maybe even make it part of the existing feathers-associations? I can give you some pointers on how to hook that functionality in.

ebourmalo:
I didn't know about [feathers-association](https://github.com/feathersjs/feathers-associations) :). 
The idea is of association is there but this plugin adds a complexity not needed IMO:
- You have to manually declare the routes you want to associate
- You break the `loosely coupled` principle of `feathers` by inserting an `associate method` to the `app`
- You'll have to create a service if there isn't an existing one for this resource
- The service you'll use with the association don't reflect the notion of `addToCollection` or `removeFromCollection`. So what (http verb and method of the service) will be associated for these actions? new methods ? Using the existing ones (create) but checking in it if we are in the classical context (POST /users) of in the embed collection context (POST /users/123/comments) and having 2 behaviour in a same method?

#### The associated collection case
If we are in the context of a relationnel model for the DB, we have a `/users` and `/projects` service. A user can work on `many` projects, a project can have `many` users (so `many-to-many` relationship). I agree with @ekryski for finding all the users of a project by doing `/users?projectId=:id` instead of `/projects/id/users`. 

But what about: 
- associating an existing project to a user? 
- associating an existing user to a project? 
- removing a user from a project? 
Currently with these services and routes, you can't do that easily right?

We can find this context of an `associated collection` in: 
- a relational DB (i.e mysql with a `many-to-many` relationship through an association table containing the fields `user_id` and `project_id`)
- a document oriented DB. If we take mongo, we can have an associated collection with a field `users` defined as an array of `ObjectId`, referencing the existing model `user`.

#### The embed collection case
In a context of a real embed collection, just a document oriented DB can do it as we can see it [here for example](http://mongoosejs.com/docs/2.7.x/docs/embedded-documents.html) with the `comments` field. 
Here we are in a specific case exposing the same problems / limits I described before. We can put this use case aside for now to simplify the discussion (even if the advanced routes also match this one).

#### The point
The associated collection within a resource (many-to-many relationship) remains commonly used in an API IMO. Lots of proper APIs now propose these new advanced routes to keeps things simple and avoid useless process (to add just an item to an associated collection, you'll have to pass the whole array (extra data for the body request), process it to determine the item to add, do some checks for validation and finally update the resource.)

I.e [github API](https://developer.github.com/v3/): `/orgs/:name:/repos`
[GET] /orgs/octokit/repos --> `fetch the list of repositories owned by the octokit organization`

These new advanced routes are shortcuts and are proposed to make services simpler, more focused on what they do and delegate the actions on an associated collection to the corresponding methods (`addToCollection`, `removeFromCollection`, `getInCollection`, `findInCollection`), not to the 'first-level' methods anymore (`update` for the 2 first methods, `get` for the 2 last ones).

I'm good to move this feature into a new plugin `feathers-collections` if I still didn't convince you to integrate it to the core ^^ I think the existing plugin `feathers-association` is a bit confusing and doesn't match 100% what is proposed here. 

If you agree with that, I'd be interested to have some tracks for hooking it properly. 
Naturally, it would be an honour for me to have access to the repo, thank you :)

If you have misunderstandings, things unclear or don't exactly get the points covered here (this is a complex subject and I may not take the easiest way to do it =/), we can do a call to speak about it :)

ebourmalo:
Just noticed I didn't ping you guys on that last reply so you may not have received notification about it. @daffl @ekryski :)

daffl:
I'm closing this PR as we are moving this functionality into https://github.com/feathersjs/feathers-collections as discussed.

devotis:
@daffl I just went through a lot of questions and answers about getting nested collections like this to work: `GET /users/1/projects/123`  

https://github.com/feathersjs/feathers-collections is a 404
https://github.com/feathersjs/feathers-associations is a 404

Common hooks explain me how to get user 1 and also see his project 123 by doing a `GET /users/1`

But I'd really like to get stuff like this to work:

GET /users/1/projects 
GET /users/1/projects/123
POST /users/1/projects/
etc

How do I do that in 2017?

marshallswain:
@devotis I'd recommend taking another look at whether it's **really** important to do it.  Sure you can do it with Express, but not with WebSockets.   If you're never going to use WebSockets, I guess there's not a problem, but then you don't need Feathers for this.  You could just read Express tutorials on how to implement this.  If you **do** plan on using WebSockets, then you're creating a disconnect between the REST and WebSocket APIs for your server.  The recommended way to query projects by user is by setting up your `/projects` endpoint and either using query params on it.  `/projects?userId=1` or, just taking advantage of functionality that you get for free with `feathers-authentication` and the rest of the authentication suite, and once a user is logged in you can trim that query down to `/projects` and on the server use `hook.params.user` to limit to only the projects for that user.  You will save yourself a lot of future headache by keeping things simple and keeping your URLs decoupled.

devotis:
Almost every rest api out there is exposing stuff like /things/:id/other-things

It just seems natural that I expose an api that is similar to that.

marshallswain:
Right.  I get it.  I'm just suggesting that the tight coupling doesn't really help the consumer of the API, and definitely doesn't help the developer of the API.  I'm also suggesting that it's probably not worth it to use that common practice (which is not a best practice, in my opinion) with a multiple-transport API.  You end up with two APIs that work differently but have to be maintained simultaneously.  You lose one of the benefits of using Feathers in the first place.

I no longer consider those URLs to be "pretty".  For me, they represent lost time.  I was caught up on them in the beginning, too.

devotis:
That makes sense in every way. You've changed my mind. I'm going to drop my 'soft' requirement for nested routes for now.

daffl:
I am 💯 with what @marshallswain said (which is partly why all the links are 404ing, sorry about that 😉).

I updated the [nested routing FAQ](https://docs.feathersjs.com/faq/readme.html#how-do-i-do-nested-or-custom-routes) with some more content and also added an example how it is still possible to create nested routes if you consider them critical. The currently recommended way looks like this:

```js
app.use('/posts', postService);
app.use('/users', userService);

// re-export the posts service on the /users/:userId/posts route
app.use('/users/:userId/posts', app.service('posts'));

// A hook that updates `data` with the route parameter
function mapUserIdToData(hook) {
  if(hook.data && hook.params.userId) {
    hook.data.userId = hook.params.userId;
  }
}

// For the new route, map the `:userId` route parameter to the query in a hook
app.service('users/:userId/posts').hooks({
  before: {
    find(hook) {
      hook.params.query.userId = hook.params.userId;
    },
    create: mapUserIdToData,
    update: mapUserIdToData,
    patch: mapUserIdToData
  }  
})
```

PR #146, opened 2015-09-21
Fix 404 not being properly thrown by REST provider

loris:
The way the routes are created by the REST provider prevent 404 errors being properly thrown.
To reproduce, just run the feathers todo-list demo and browse to an URL like `/todos/foo/bar`, it won’t return a 404 as it should but a 200 without any content (because of the `app.use(uri, handler);` statement catching any URLs starting with `/todos`)


daffl:
That makes sense, thank you! I will release a 1.1.1 with this fix later today.

PR #178, opened 2015-12-01
Handle middleware passed after the service to app.use

### LINKED ISSUE #176: Feature request: (req, res, next) handler for *after* the service
Looks like this works:
```js
  app.use(endpoint, function (req, res, next) {
    console.log(`${endpoint} before 1`);
    next();
  }, function (req, res, next) {
    console.log(`${endpoint} before 2`);
    next();
  }, Stream);
```

But this doesn't: (`TypeError: Router.use() requires middleware function but got a Object`)

```js
  app.use(endpoint, function (req, res, next) {
    console.log(`${endpoint} before 1`);
    next();
  }, function (req, res, next) {
    console.log(`${endpoint} before 2`);
    next();
  }, Stream, function (req, res, next) {
    console.log(`${endpoint} after`);
    next();
  });
```

I also tried this, but the `after` handler simply gets ignored:
```js
  app.use(endpoint, function (req, res, next) {
    console.log(`${endpoint} before 1`);
    next();
  }, function (req, res, next) {
    console.log(`${endpoint} before 2`);
    next();
  }, Stream);
  app.use(endpoint, function (req, res, next) {
    console.log(`${endpoint} after`);
    next();
  });
```

dbkaplun:
Closes #176

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -56,12 +56,22 @@ module.exports = {
     return protoService;
   },
 
-  use: function () {
-    var args = _.toArray(arguments);
-    var location = args.shift();
-    var parts = _.partition(args, _.isFunction);
-    if (parts[1].length > 1) throw new Error('more than one service passed to app.use');
-    var service = parts[1].pop();
+  use: function (location) {
+    var service, middleware = _(arguments)
+      .slice(1)
+      .reduce(function (middleware, arg) {
+        if (typeof arg === 'function') {
+          middleware[service ? 'after' : 'before'].push(arg);
+        } else if (!service) {
+          service = arg;
+        } else {
+          throw new Error('invalid arg passed to app.use');
+        }
+        return middleware;
+      }, {
+        before: [],
+        after: []
+      });
     var hasMethod = function(methods) {
       return _.some(methods, function(name) {
         return (service && typeof service[name] === 'function');
@@ -75,7 +85,7 @@ module.exports = {
 
     this.service(location, service, {
       // Any arguments left over are other middleware that we want to pass to the providers
-      middleware: parts[0]
+      middleware: middleware
     });
 
     return this;
--- a/lib/application.js
+++ b/lib/application.js
@@ -59,7 +59,9 @@ module.exports = {
   use: function () {
     var args = _.toArray(arguments);
     var location = args.shift();
-    var service = args.pop();
+    var parts = _.partition(args, _.isFunction);
+    if (parts[1].length > 1) throw new Error('more than one service passed to app.use');
+    var service = parts[1].pop();
     var hasMethod = function(methods) {
       return _.some(methods, function(name) {
         return (service && typeof service[name] === 'function');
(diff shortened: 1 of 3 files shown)


lib/application.js
```diff
@@ -59,7 +59,9 @@ module.exports = {
   use: function () {
     var args = _.toArray(arguments);
     var location = args.shift();
-    var service = args.pop();
+    var parts = _.partition(args, _.isFunction);
+    if (parts[1].length > 1) throw new Error('more than one service passed to app.use');
```
daffl: Small style nit, I'd make it

```js
if (parts[1].length > 1) {
  throw new Error('more than one service passed to app.use');
}
```

lib/application.js
```diff
@@ -73,7 +75,7 @@ module.exports = {
 
     this.service(location, service, {
       // Any arguments left over are other middleware that we want to pass to the providers
-      middleware: args
+      middleware: parts[0]
```
daffl: I think this would just add everything before the service happens right? Maybe we need a `middleware: { before: [], after: [] }`?

daffl:
Awesome, thank you! I made some comments and it would be nice to have some tests for this functionality (probably similar to https://github.com/feathersjs/feathers/blob/master/test/providers/rest.test.js#L337).
After that we can definitely get it into the new release!

dbkaplun:
@daffl fixed and added tests

test/application.test.js
```diff
@@ -128,21 +128,25 @@ describe('Feathers application', function () {
     var app = feathers()
       .configure(feathers.rest())
       .use('/todo', function (req, res, next) {
-        req.feathers.stuff = 'custom middleware';
+        req.feathers.preService = 'pre-service middleware';
         next();
-      }, todoService)
+      }, todoService, function (req, res, next) {
+        res.set('post-service', res.data.id);
```
daffl: I think we could set and then check for `res.data.postService` instead since the handler should run last.

daffl:
This is great! Thank you! One more small comment for the test. Nothing major but then we can get it in and ready for a version 1.3.

test/application.test.js
dbkaplun: Sorry I don't understand, can you rephrase please?

daffl: What I mean is to instead set something like:

```js
res.data.postService = true;
```

and then in the test check for

```js
assert.ok(data.postService, 'Post-service middleware updated response');
```

instead of `assert.equal(response.headers['post-service'], 'dishes', 'Post-service middleware updated response');`.

dbkaplun: I see! I did it this way because I wanted to actually ensure that the post-service handler actually runs after the service, but I can change it if you like.

daffl: Good point. Lets leave it then. I will merge shortly. Thanks again!

ekryski:
Oh nice! This PR snuck by me but I've run into this before. Thanks @dbkaplun!

PR #188, opened 2016-01-03
Migration to ES6 and API providers in separate modules

### LINKED ISSUE #156: Split feathers transports into their own modules
We should do this so that people don't need to include compile engine.io when they don't need to. It also helps with our mantra of keeping things small and modular.

Modules that should be created:

- [x] [feathers-rest](https://github.com/feathersjs/feathers-rest)
- [x] [feathers-socketio](https://github.com/feathersjs/feathers-socketio)
- [x] [feathers-primus](https://github.com/feathersjs/feathers-primus)

### LINKED ISSUE #183: Convert Feathers core to ES6
Since we are converting everything to ES6 we should convert core as well. This should probably happen in tandem with #156.

daffl:
This pull request migrates Feathers Core to ES6 using Babel and removed the API providers which are now in separate modules (#156).

Closes #156 and closes #183.


src/mixins/event.js
```diff
@@ -0,0 +1,57 @@
+import _ from 'lodash';
```
ekryski: Now that we are ES6 we should be able to remove lodash pretty easily.

src/feathers.js
```diff
@@ -0,0 +1,25 @@
+if(!global._babelPolyfill) { require('babel-polyfill'); }
+
+import _ from 'lodash';
```
ekryski: Now that we are ES6 we should be able to remove lodash pretty easily.

ekryski:
@daffl Nice work man. This makes the core so small. I :heart: it so much!

:shipit:

ekryski:
I guess once we roll this out we should update the examples in the database adapters to use the ES6 syntax or possibly add an ES6 example alongside the old ones.

marshallswain:
So what's on the todo list before we merge this?

daffl:
Documentation but I don't think that should block the merge. I started them in the https://github.com/feathersjs/feathers/tree/gh-pages-2.0 branch but didn't get that far. We need to do the reoganization we talked about, update all code samples to ES6 and add the topics that we'd like to cover and that are still missing.

Just for migration purposes we can just create a Gist that describes what is necessary (which should mostly just be replacing `app.configure(feathers.rest())` etc. with `var rest = require('feathers-rest'); app.configure(rest());`).

ekryski:
Yes, and also how you initialize the database adapters. Let's merge this PR and then we'll do a whole swath of documentation updates.

PR #193, opened 2016-01-09
Make Feathers universal

### LINKED ISSUE #191: Include Feathers client core for universal Feathers applications
As mentioned by @ahdinosaur in #156, what is currently in [feathers-client](https://github.com/feathersjs/feathers-client/) can actually be moved into the main repository and the respective transports. Using the `browser` field, this will allow for a fully isomorphic Feathers that can be used the same way on the client and the server, including feathers-hooks (related issue https://github.com/feathersjs/feathers-client/issues/20).

daffl:
This pull request adds a client side shim for Express, allowing to use Feathers in non-Node environments like the Browser and React Native to build fully universal applications with Feathers best practises like services and hooks. The Browserified build is around 8k minified and gzipped.

Also adds a `.defaultService` functionality. This will be needed when initializing non-existing services on the client which should then just connect to their configured endpoint as it is already done in [feathers-client](https://github.com/feathersjs/feathers-client).

Closes #191

### DIFF
--- a/client.js
+++ b/client.js
@@ -1,0 +1,1 @@
+module.exports = require('./lib/client');
--- a/package.json
+++ b/package.json
@@ -69,5 +69,10 @@
     "q": "^1.0.1",
     "request": "^2.x",
     "socket.io-client": "^1.0.0"
+  },
+  "browser": {
+    "express": "./lib/client/express",
+    "./lib/feathers": "./lib/client/index",
+    "babel-polyfill": false
   }
 }
--- a/src/application.js
+++ b/src/application.js
@@ -20,11 +20,17 @@ export default {
     });
   },
 
-  service(location, service, options) {
+  service(location, service, options = {}) {
     location = stripSlashes(location);
 
     if(!service) {
-      return this.services[location];
+      const current = this.services[location];
+
+      if(typeof current === 'undefined' && typeof this.defaultService === 'function') {
+        return this.service(location, this.defaultService(location), options);
+      }
+
+      return current;
     }
 
     let protoService = Proto.extend(service);
@@ -40,7 +46,7 @@ export default {
 
     // Run the provider functions to register the service
     this.providers.forEach(provider =>
-      provider.call(this, location, protoService, options || {}));
+      provider.call(this, location, protoService, options));
 
     // If we ran setup already, set this service up explicitly
     if (this._isSetup && typeof protoService.setup === 'function') {
@@ -48,8 +54,7 @@ export default {
       protoService.setup(this, location);
     }
 
-    this.services[location] = protoService;
-    return protoService;
+    return (this.services[location] = protoService);
   },
 
   use(location) {
--- a/src/client/express.js
+++ b/src/client/express.js
@@ -1,0 +1,47 @@
+import { EventEmitter } from 'events';
+import Proto from 'uberproto';
+
+export default function() {
+  const app = {
+    settings: {},
+
+    get(name) {
+      return this.settings[name];
+    },
+
+    set(name, value) {
+      this.settings[name] = value;
+      return this;
+    },
+
+    disable(name) {
+      this.settings[name] = false;
+      return this;
+    },
+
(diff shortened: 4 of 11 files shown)


ekryski:
@daffl can you spec out a quick example of how this would be used? Is it in conjunction with `feathers-client` or is it a replacement for now?

daffl:
`feathers-client` would still exist (and also be backwards compatible) but just re-export the respective clients (`feathers/client`, `feathers-socketio/client`, `feathers-primus/client` and `feathers-rest/client`) to provide a non-CommonJS distributable.

Importing `feathers` directly only works with CommonJS compatible module loaders like Webpack, Browserify or Steal that use the `browser` field in `package.json`. With those it should now be possible to architect client side Feathers apps the same way as on the server though:

```js
import feathers from 'feathers';
import socketio from 'feathers-socketio';
import hooks from 'feathers-hooks';
import localStorage from 'feathers-localstorage';
import React from 'react';

const app = feathers()
  .configure(socketio('http://chat.feathersjs.com'))
  .configure(hooks())
  // You can register your own services to manage application state
  .use('/store', localStorage('storage'));

// Add hooks e.g. for client-side validation
app.service('messages').before({
  create(hook, next) {
    if(typeof hook.message !== 'string') {
      return next(new Error('A message needs to be provided'));
    }
    next();
  }
});

class TodoList extends React.Component {
  componentDidMount() {
    app.service('messages').find().then(messages => 
      this.setState({ messages }));
  }

  render() {
    if(!this.state.messages) {
      return <div>Loading...</div>;
    }

    return <ul>{this.state.messages.map(message => return <li>{message.text}</li>)}</ul>;
  }
}
```

To use the client in Node you'd `import feathers from 'feathers/client'`, `import socketio from 'feathers-socketio/client'` etc.

ekryski:
Man this looks awesome!! You'll probably do the same thing in react native: `import feathers from 'feathers/client'`.

Let's merge this sucker. I want to play!! :smile:

On another note, you didn't write `feathers-localstorage` yet did you?

daffl:
Not yet. It was just an idea and as an example that you could use services and other Feathers best practises to also manage application state on the client. Working on migrating the REST, SocketIO and Primus clients now.

ahdinosaur:
wow this is great @daffl, really keen to play around with this, thanks. :)

daffl:
@ahdinosaur Thanks for bringing that up! It definitely turned out even better than I expected. Everything should be released now in `feathers` 2.0.0-pre.1 and 1.1.0 in the `feathers-rest`, `feathers-socketio` and `feathers-primus` modules. If you and/or @ekryski want to take it for a spin. Please let me know if you are seeing any issues (I'm hoping I got the `browser` field right but there wasn't an easy way to test it).

ekryski:
Yay! I'll check it out today! I guess this means that we now no longer need https://github.com/feathersjs/feathers-client.

daffl:
We still need a non CommonJS/NPM build. The question is if they should be provided as distributables in their respective repos or consolidated in feathers-client as a dist that includes all the provider clients and probably hooks.

The first option makes it a little annoying to get set up in a non CommonJS/NPM environment because you'll have to include a whole bunch of dependencies manually. Namespacing is also trickier than having  the `feathers-client` module that simply does something like

```
// src/client.js
import feathers from 'feathers/client'
import socketio from 'feathers-socketio/client';
import primus from 'feathers-primus/client';
import rest from 'feathers-rest/client';
import hooks from 'feathers-hooks';

Object.assign(feathers, { socketio, primus, rest, hooks });

export default feathers;
```

And then Browserifies that as an UMD module or a global `feathers` variable. That way we also won't break any existing feathers-client users too badly. Depending on how it goes I think we can definitely think about deprecating the client in v3.

ekryski:
@daffl Ya I'm for option 2: `feathers-client` just being a simple wrapper for all the feathers libs. It will make it easier if we need to make any modifications to it (or create a separate client) for React Native.

PR #205, opened 2016-01-18
Appending `nsp check` to test script.

marshallswain:
This adds a Node Security Project scan to the end of the tests.


daffl:
Oh nice. That was easy :) So our CI should fail if it finds any know vulnerabilities. We may even want to add this to all the other plugins, too.

ekryski:
definitely should add this to the other plugins if it is this easy

PR #208, opened 2016-01-18
Explicitly import /index files to work with Steal.

marshallswain:


daffl:
:+1: I'll make another pre shortly which will also include the event filtering.

PR #246, opened 2016-02-29
Fix mixing execution order by reversing it

daffl:
Because of [Uberproto's](https://github.com/daffl/uberproto) mixin order, the last mixin gets always executed first which is why we had to had the normalized to always be on the end.

The correct fix is to reverse the mixin array when a service is registered so that they run in the expected order. This should also fix a bug with service event data that were being sent before hooks are executed.

### DIFF
--- a/src/application.js
+++ b/src/application.js
@@ -38,7 +38,7 @@ export default {
     debug(`Registering new service at \`${location}\``);
 
     // Add all the mixins
-    this.mixins.forEach(fn => fn.call(this, protoService));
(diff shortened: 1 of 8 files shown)


daffl:
I'm going to close this one but keep the branch and the issue around. This may require some breaking changes so we will table it for version 3.

ekryski:
Why don't we keep it open but start a v3 milestone?

ekryski:
Nvm. We should get into the habit of creating a new major milestone branch. So this should be merged into that one. So we'll need to close this an do a new PR.

PR #268, opened 2016-03-17
Fix link to docs

lepiaf:
Hi,

There is no link on first button. I added it.


daffl:
Thank you! I saw this earlier but then obviously forgot about fixing it.

ekryski:
It's meant to scroll down to the next section. Maybe that wasn't working.

ekryski:
Either way, thanks @lepiaf this is a much better fallback.

PR #306, opened 2016-04-17
Remove JSON loading from the client version

daffl:
Some loaders (like Webpack) can not require JSON by default so it is best to take out requiring the package to get the version. This will also make built versions a little smaller. The drawback is that we will have to update those versions manually (maybe remove them entirely in 3.0).


ekryski:
:shipit:

ekryski:
It's a bummer to add a manual step but it is required and slims the build so 👍

PR #308, opened 2016-04-22
Test and fix for allowing services with only a setup method

### LINKED ISSUE #285: Cryptic error message on "setup-only" service.
Just noticed if you create a service with only a Service#setup(...) function and `app.use(...)` it. 
You get a cryptic error message ( my guess - it's coming from express ), "Router.use() requires middleware function but got a Object".

Example:
```
// Error
const SomeService = {
	setup(app, path) {}
}
​
// No Error
​
const AnoterService = {
	setup(app, path) {},
	create(data, params, cb) {}
}
​
​
app.use('/some', SomeService); // cryptic error
app.use('/another', AnotherService); // Fine
```

daffl:
Closes #285

### DIFF
--- a/src/application.js
+++ b/src/application.js
@@ -46,7 +46,8 @@ export default {
 
     // Run the provider functions to register the service
     this.providers.forEach(provider =>
-      provider.call(this, location, protoService, options));
+      provider.call(this, location, protoService, options)
+    );
 
     // If we ran setup already, set this service up explicitly
     if (this._isSetup && typeof protoService.setup === 'function') {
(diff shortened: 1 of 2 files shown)


src/application.js
```diff
@@ -75,10 +76,11 @@ export default {
       });
 
     const hasMethod = methods => methods.some(name =>
-      (service && typeof service[name] === 'function'));
+      (service && typeof service[name] === 'function')
+    );
 
     // Check for service (any object with at least one service method)
-    if(hasMethod(['handle', 'set']) || !hasMethod(this.methods)) {
+    if(hasMethod(['handle', 'set']) || !hasMethod(this.methods.concat('setup'))) {
```
ekryski: Do we want to throw an error or a debug warning in case the service doesn't have a `setup` method?

test/application.test.js
```diff
@@ -326,6 +326,20 @@ describe('Feathers application', () => {
     assert.equal(otherApp.mixins.length, 4);
   });
 
+  it('setup only services work (#285)', done => {
```
ekryski: I don't really know what this means? Maybe we can rephrase to "initializes a service with only a setup method"?

src/application.js
daffl: What do you mean? Services without a `setup` method shouldn't be a problem.

ekryski: After I made that comment I realized it didn't make sense. PROCEED! 😄

ekryski:
:shipit:

PR #412, opened 2016-09-21
expose the service name to hooks and internal service methods

ekryski:
We need this functionality for checking permissions. @daffl let me know if you think this the right spot.

### DIFF
--- a/src/application.js
+++ b/src/application.js
@@ -94,6 +94,7 @@ export default {
     // Setup each service (pass the app so that they can look up other services etc.)
     Object.keys(this.services).forEach(path => {
       const service = this.services[path];
+      service.name = path;
 
       debug(`Setting up service for \`${path}\``);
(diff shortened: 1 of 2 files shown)


daffl:
It is possible to use the same service object on two different paths. I think it should happen in https://github.com/feathersjs/feathers-hooks/blob/master/src/hooks.js#L36 but that might require us to add the path to mixin (https://github.com/feathersjs/feathers/blob/master/src/application.js#L41). That basically makes them the same as a provider (which is called a couple of lines after). For V3 we should really consolidate them.

daffl:
I think we can close this as per our discussion today? I think hooks should include the path but we may have to wait until V3 to make providers and mixins the same.

ekryski:
After some discussion. We've found we no longer need this and this isn't the correct spot to do this in case the same service is used for multiple paths.

PR #425, opened 2016-10-12
Switch from jshint to semistandard

marshallswain:


ekryski:
Looks good. Looks like the build is just failing on old node versions due to the eslint config. Does eslint autoconfigure based on the node version being used?

daffl:
It looks like ESLint does not work in older Node versions which is unfortunate. We can probably drop support for anything other than the LTS and the latest Node in V3 but not before.

marshallswain:
I created a utility that lets us run eslint conditionally: https://www.npmjs.com/package/eslint-if-supported

It checks the node version and if it's less than 4.0, it just exits with a success code which skips eslint and moves on to the tests.

To use it, you just install the package and prepend it to the "eslint" script like this:
```
"eslint": "eslint-if-supported semistandard --fix"
```

ekryski:
@marshallswain you able to resolve conflicts and ship this?

daffl:
Or should we defer it to when we start working on v3?

ekryski:
I'm good to ship it now and update all the plugins over time. Updating them all is a pain. But don't really care either way. Just don't want this PR to become too stale if it's actually good to merge.

PR #426, opened 2016-10-12
body-parser, feathers-* and socket.io-client are dependencies

StevenBlack:
... not devDependencies.

I hope I have this right  😄


daffl:
Thank you for the pull request. Those modules are not required dependencies of Feathers core however. They are only used in the tests (which is why they are devDependencies).

StevenBlack:
@daffl yeah I figured that might be the case.

Closing.

PR #430, opened 2016-10-17
jshint —> semistandard

marshallswain:
This installd and implements semistandard for linting.  Uses eslint-if-supported for tests.


ekryski:
:shipit: All good homie!

Issue #435, opened 2016-10-17
Add issue and PR templates to all repos


ekryski:
We should add a `.github` folder to each repo with issue and PR templates. Let's figure out what the wording should be or if there is a good example we can copy and then I think we can write a quick script to add this to each repo.

### Repos
- [x] feathers
- [x] feathers-memory
- [x] feathers-mongodb
- [x] feathers-mongoose
- [x] feathers-nedb
- [x] feathers-query-filters
- [x] feathers-errors
- [x] feathers-hooks
- [x] feathers-hooks-common
- [x] feathers-commons
- [x] feathers-rethinkdb
- [x] feathers-rest
- [x] feathers-socketio
- [x] feathers-primus
- [x] feathers-authentication
- [x] feathers-authentication-oauth2
- [x] feathers-authentication-local
- [x] feathers-sendgrid
- [x] feathers-mailgun
- [x] feathers-mailer
- [x] feathers-stripe
- [x] feathers-twilio
- [x] feathers-reactive
- [x] feathers-blob
- [x] feathers-levelup
- [x] feathers-permissions
- [x] generator-feathers-plugin
- [x] feathers-cli
- [x] feathers-generator
- [x] feathers-socket-commons
- [x] feathers-localstorage
- [x] feathers-sequelize
- [x] feathers-knex
- [x] feathers-reactive
- [x] feathers-waterline
- [x] feathers-configuration
- [x] feathers-batch
- [x] feathers-sync
- [x] feathers-bootstrap
- [x] feathers-service-tests?
- [x] feathers-client

corymsmith:
This has been completed.

PR #442, opened 2016-10-22
adding .github folder and templates

ekryski:
@feathersjs/core-team please review and add feedback. We'll iterate in this PR and then can add a script to add this folder to all the other repos as outlined in #435.


.github/issue_template.md
```diff
@@ -0,0 +1,20 @@
+### Steps to reproduce
+
+(First please check that this issue is not already solved as [described
+here](https://github.com/feathersjs/feathers/blob/master/.github/contributing.md#report-a-bug))
+
+- [ ] Tell us how you broke things. The more detailed the better.
```
daffl: Maybe less accusing? Something like "Tell us what broke."

ekryski: lol true dat

.github/issue_template.md
```diff
@@ -0,0 +1,20 @@
...
+
+### Actual behavior
+Tell us what happens instead
+
+### System configuration
+**Module versions** (especially the part that's not working):
+
+**NodeJS version**:
+
+**Operating System**:
```
daffl: Should we ask for the client configuration here, too? Browser version and module loader (if applicable)

ekryski: Ya probably. At least browser version and module loader

corymsmith:
Once we agree on this I can sort the script to add it to all repos, have a good idea on that, should be pretty simple using something like https://github.com/github/hub since I already have all the repos in a folder.

ekryski:
@corymsmith That was my thinking as well. The quick and dirty would be if you already have the folders and write access to simply make a script that rips through all of them and copies the folder to each one, adds it to the end of `.npmignore`, commits and pushes straight to master.

corymsmith:
Exactly what I was thinking, though I was thinking of PR but thats probably overkill

ekryski:
For this stuff I think so. It doesn't have any impact on code and it's going to be the same in every repo.

corymsmith:
I've got a script sorted for this so let me know when we're good to go

Issue #381, opened 2016-08-07
TypeScript support


Hongbo-Miao:
http://docs.feathersjs.com/frameworks/angular2.html
Now the demo uses

```
const feathers = require('feathers/client');
const socketio = require('feathers-socketio/client');
const io = require('socket.io-client');
const localstorage = require('feathers-localstorage');
const hooks = require('feathers-hooks');
const rest = require('feathers-rest/client');
const authentication = require('feathers-authentication/client');
```

Please support TypeScript definitions to use `import`!

daffl:
The best way to move this forward would be to start a repository with some basic Typescript definitions (in #64 I posted some pointers on where to start) and continue
 discussion there.

I can leave this issue open for another month or two but the other issue was open for almost a year with nobody picking it up. As mentioned there, the current core team is not using Typescript or Angular so it isn't really on our radar (and we'd prefer only keeping issues open if there is a chance that they'll be addressed in the foreseeable future).

daffl:
I think @harangue has been working on some basic definitions. Maybe we can move the discussion we had on Slack about where they would live over here?

Hongbo-Miao:
sure

rollymaduk:
It will be really nice if typescript can be supported.. how can I help?

harangue:
@rollymaduk I've made some decent progress on definitions for Feathers, but we have yet to establish the best way to distribute them (and things are a little transitional with TypeScript 2 about to release). If you'd like to have something to work off until then, you can [see this gist](https://gist.github.com/harangue/9d4ed79118e2656f5e3d2d699296ed36) I've worked on.

rollymaduk:
@harangue really nice work infact it looks quite ready.. I guess some areas that may require some attention would be incorporating some basic plugins like authentication, providers (rest,socke.io,primus) and some of the db adapters right?  please what do you mean by the best way to distribute them?

daffl:
I'm not sure if we discussed this yet, but I think it'll be fine if we distribute the Typescript definitions in the main repository as long as there is someone to maintain and keep them up to date. So once they are ready, we can make it part of a Feathers 2.1 release (which I think might also ship with some warnings that e.g. callbacks in services will be deprecated in v3).

j2L4e:
Is there any fork of `generator-feathers` for TypeScript? @harangue How are you generating your TS projects? Interested in working on a TS fork of the generator?

daffl:
[...]

PR #457, opened 2016-11-09
Created TS definitions

thosakwe:
Feel free to check them out for yourselves. I plan to add definitions for the other core Feathers libraries, so that they can be easily used with Angular 2.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,6 @@ tmp*
 
 # The compiled/babelified modules
 lib/
+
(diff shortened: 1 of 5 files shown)


thosakwe:
Actually, I just saw the definitions by @harangue, never mind.

marshallswain:
@thosakwe where are the ones by @harangue?

thosakwe:
See #381 :)

PR #506, opened 2017-01-31
ES2015 module support for rollup/webpack/...

ForsakenHarmony:
[Read more](https://github.com/rollup/rollup/wiki/pkg.module)

I probably should not have done this on github but it's too late now

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -4,7 +4,6 @@
 .istanbul.yml
 .babelrc
 .idea/
-src/
 test/
 !lib/
 coverage


daffl:
Looks good and simple enough. Did you get a chance to try it out with Rollup? If it can do tree shaking and export UMD builds I'd be curious how much of a size difference there would be compared to the Browserify feathers-client build we are using right now (it's about ~21k right now minified + gzipped so it's not super important).

ForsakenHarmony:
no, sorry didn't get around to testing anything

daffl:
Thanks again for the PR @ForsakenHarmony. I'm closing it in favour of #608 though.

PR #573, opened 2017-05-09
Fix typescript defnition of Service. All the service methods should be

harish2704:
optional

### Summary

As per the [current type definition of Service ](https://github.com/feathersjs/feathers/blob/master/index.d.ts#L23), all the service methods are required methods. 

But according to documentation, service methods are optional . 
See 
image

So, service definition should be something like below
```typescript
  interface Service<T> extends events.EventEmitter {
    find?(params?: Params, callback?: any): Promise<T[] | Pagination<T>>;
    get?(id: number | string, params?: Params, callback?: any): Promise<T>;
    create?(data: T | T[], params?: Params, callback?: any): Promise<T | T[]>;
    update?(id: NullableId, data: T, params?: Params, callback?: any): Promise<T>;
    patch?(id: NullableId, data: any, params?: Params, callback?: any): Promise<T>;
    remove?(id: NullableId, params?: Params, callback?: any): Promise<T>;
    setup?(app?: Application, path?: string): void;
  }
```

I think, This pull request don't need any sort of test cases because it won't affect any other part of the application.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -26,41 +26,41 @@ declare namespace feathers {
      * Retrieves a list of all resources from the service.
      * Provider parameters will be passed as params.query
      */
-    find(params?: Params, callback?: any): Promise<T[] | Pagination<T>>;
+    find?(params?: Params, callback?: any): Promise<T[] | Pagination<T>>;
 
     /**
      * Retrieves a single resource with the given id from the service.
      */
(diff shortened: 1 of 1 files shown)


eddyystop:
Thanks for contributing! @daffl could you review?

daffl:
Done and released, thank you! To mention it once again, we would still appreciate it very much if someone could take on submitting the definitions to the DefinitelyTyped repository.

PR #602, opened 2017-06-23
Return types needed

Creiger:
### Summary

I am not sure if it is only my problem, but i got an compilation erorr without types defined.


eddyystop:
The community maintains the TS typings, so thanks!

daffl:
Released as v2.1.4

Issue #589, opened 2017-06-01
hook.path is not set in filter function


mxgr7:
### Steps to reproduce

```
app.service('example').filter(function (data, connection, hook) {
  if (hook.path !== 'example')
    throw new Error(`hook.path is ${hook.path}`)
  return data;
})
```

### Expected behavior
`hook.path` should be set (to `'example'` in the above example), therefore the filter above should not throw an error.

### Actual behavior
`hook.path` is `undefined`, so the above filter throws an error.

### System configuration

Tested with feathers version 2.1.3.

daffl:
This will be fixed in the next version via #408 

A workaround for right now should be

```js
app.service('example').filter(function (data, connection, hook) {
  const path = Object.keys(app.services).find(current => app.services[current] === this);
  
  return data;
})
```

Issue #588, opened 2017-05-31
Using Rollup throws errors about dependency Rubberduck using eval


colwilson:
### Steps to reproduce

- Rollup throws "Use of eval is strongly discouraged, as it poses security risks and may cause issues with minification" - https://github.com/rollup/rollup/wiki/Troubleshooting#avoiding-eval
- see repo at https://bitbucket.org/colwilson/feathers-rollup-rubberduck-issue/src 
- deprecation of rubberduck is also mentioned in **Make feathers-hooks part of core and service events a hook** https://github.com/feathersjs/feathers/issues/408

```sh
$ yarn
$ yarn dev
yarn dev v0.21.3
$ cross-env NODE_ENV=development rollup -c 
preferring built-in module 'events' over local alternative at '/home/col/dev/play-valued-stud/node_modules/events/events.js', pass 'preferBuiltins: false' to disable this behavior or 'preferBuiltins: true' to disable this warning
preferring built-in module 'events' over local alternative at '/home/col/dev/play-valued-stud/node_modules/events/events.js', pass 'preferBuiltins: false' to disable this behavior or 'preferBuiltins: true' to disable this warning
preferring built-in module 'events' over local alternative at '/home/col/dev/play-valued-stud/node_modules/events/events.js', pass 'preferBuiltins: false' to disable this behavior or 'preferBuiltins: true' to disable this warning
⚠️   Use of eval is strongly discouraged, as it poses security risks and may cause issues with minification
https://github.com/rollup/rollup/wiki/Troubleshooting#avoiding-eval
node_modules/rubberduck/lib/rubberduck.js (33:6)
31: 
32:     if (strict) {
33:       eval('wrapped = ' + utils.addArgs(wrapped.toString(), fn.length));
          ^
34:     }

⚠️   Use of eval is strongly discouraged, as it poses security risks and may cause issues with minification
https://github.com/rollup/rollup/wiki/Troubleshooting#avoiding-eval
node_modules/rubberduck/lib/rubberduck.js (76:6)
74: 
75:     if (strict) {
76:       eval('wrapped = ' + utils.addArgs(wrapped.toString(), fn.length));
          ^
77:     }

Done in 2.89s.

```

### Expected behavior
Should compile

### Actual behavior
rollup fails with "Use of eval is strongly discouraged, as it poses security risks and may cause issues with minification" - https://github.com/rollup/rollup/wiki/Troubleshooting#avoiding-eval

paulocoghi:
[...]

Issue #566, opened 2017-05-02
Re-using services


Mattchewone:
In ref to feathers-hooks issue [#154](https://github.com/feathersjs/feathers-hooks/issues/154)

There was an issue when using a service under a different route path, ie:
```
  app.use('/projects', createService(options));

  const service = app.service('/projects');

  app.use('/me/projects', service)
```
The __hooks object was being overwritten, which has since been fixed but now in the console there is the following output:
```
express:router <anonymous>  : /projects +8ms
feathers-errors NotFound(404): Page not found +1ms
express:router <anonymous>  : /projects +0ms
finalhandler default 404 +1ms
finalhandler cannot 404 after headers sent +0ms
Error: Can't set headers after they are sent.
  at ServerResponse.OutgoingMessage.setHeader (_http_outgoing.js:356:11)
  at ServerResponse.header
....
```

It seems as though it is trying to route it again and not finding the route and trying to throw NotFound error but a response has already been set with headers. 

As per @daffl suggestion to post this in here.

Thanks

daffl:
Thank you! This is definitely an issue. I really thought I tested it somewhere but it doesn't seem so.

Mattchewone:
@daffl Been having a little dig into this and with some debugging, I can see that after the `/me/projects` call that the not-found handler is called which I believe is then trying to set headers even though they are already set from the service call. 

Here is a little output I have put together:
```
  feathers:rest getHandler
  feathers:rest REST handler calling `find` from `/projects` +5s
  service:call lookup
  info: after: projects - Method: find
  feathers-rest formatter..
```
```
  feathers:rest getHandler
  feathers:rest REST handler calling `find` from `/me/projects` +5s
  service:call lookup
  info: after: me/projects - Method: find
  feathers-rest formatter..
  feathers-errors not-found
```

I can't quite work out where a next() is being called to trigger the `not-found` handler to catch the request.

I found a work-a-round but would be good to know what is triggering the next call.
```javascript
import errors from './index';

export default function() {
	return function(req, res, next) {
		if (!res.headersSent) {
			next(new errors.NotFound('Page not found'));
		}
	};
}
```

If you have any suggestions that would be awesome.

Issue #515, opened 2017-02-11
Universal/Isomorphic Express


Download:
I just wanted to point you guys to a discussion happening in the Express project that, in a roundabout way, was actually triggered in part by Feathers JS.

Feathers JS has an application object (Feathers) that mimics the Express API. I really liked that aspect of Feathers so I looked how you did it and it inspired me to try to take it a step further, by also enabling the `use` method to work on the client... I ended up building [uexpress](https://github.com/download/uexpress). 

I asked [a question](https://github.com/expressjs/express/issues/3198) about the Express source code on their issue tracker and it was seen by Wesley Todd, who, with his project [nighthawk](https://github.com/wesleytodd/nighthawk), was doing something similar. He somehow got it scheduled on an [Express TC meeting](https://github.com/expressjs/discussions/issues/51) and they [discussed](http://youtu.be/Ol-gr_OQuHg) what it would take in the core of Express 5 to allow projects like nighthawk, uexpress and Feathers JS to assemble their own Express-compatible servers, using the same parts used by Express core (factored out into separate packages).

Since you guys have a lot of experience in writing isomorphic code with Express, and are (as I understand it) looking at generalizing this concept over to multiple server types (Koa, etc), I think you would be able to weigh in on this discussion with your insights.

eddyystop:
@daffl

daffl:
What we are planning for the next version is basically turning our current client (https://github.com/feathersjs/feathers/blob/master/src/client/express.js) into core and then mixing in framework specific functionality as plugins (`app.configure(feathersExpress())`). Our client looks very similar to what you did in https://github.com/Download/uexpress/blob/master/uexpress.js#L40.

Some random thoughts:

- From our experience, sub-apps with websockets and authentication are hard (and annoying). I'm not entirely sure what we want to do about that yet.
- Feathers itself does not really need any routing functionality
- I'd like to discourage using a view rendering engine in Feathers. The better way to use it on the server would be importing the Feathers app (and potentially authentication) into a separate Express application and use it that way.

Anyway, if there is anything we can do to help we'd be more than happy to. Currently I am not entirely sure how exactly that would look like though - without sub-apps, routing or view rendering it's basically only `.get`, `.set` and the event emitter that Feathers is relying on.

Download:
> Our client looks very similar to what you did

Lol that is no coincidence... I stole a lot from you guys. Thanks for that BTW!

> I'd like to discourage using a view rendering engine in Feathers. 

I agree.

[...]

PR #605, opened 2017-06-28
Initial changes for Feathers v3

daffl:
This pull request does quite a couple of things that couldn't easily be split up:

- Makes Feathers framework independent and removes the dependency on Express (#258, #515)
- Pulls feathers-hooks into core and removes dependency on Rubberduck (#408, #376, #589, #588)
- Remove auto import of babel-polyfill (#510)
- Makes sure that services can be re-used (#566)


kristianmandrup:
When can we expect the v3 to be released which decouples feathers from express? 

I'd like to use feathers with [Fastify](https://thenewstack.io/introducing-fastify-speedy-node-js-web-framework/) or koa :)

daffl:
You can install it with `npm install feathers@pre` but the framework independent release does not necessarily mean that it will automatically work with Koa right away. @christopherjbaker started that discussion in https://github.com/feathersjs/feathers-express/issues/3 though.

paulocoghi:
Even that the new Feathers (v3) will be framework independent, it will be interesting to learn the techniques  used in Fastify :)

PR #607, opened 2017-06-29
Allow flag to return the hook object

daffl:
This will be used by the updated providers (primarily to retrieve `hook.dispatch` which will contain the "safe" data to send to a client). Allows to do something like:

```js
app.service('dummy').find({
  __returnHook: true
}).then(hook => {
  // hook is the actual hook object instead of just the result
});
```

### DIFF
--- a/src/hooks.js
+++ b/src/hooks.js
@@ -67,7 +67,9 @@ export function hookMixin (service) {
           });
         })
         // Make a copy of hookObject from `before` hooks and update type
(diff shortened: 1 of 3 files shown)


ekryski:
Looks good @daffl do we have a test somewhere that actually makes sure the hook object is returned when the flag is set? The default is falsey I'm assuming.

daffl:
It's at the bottom of the diff: https://github.com/feathersjs/feathers/pull/607/files#diff-3c6456101c2eef8b9a21f4595301eebeR157

Issue #608, opened 2017-06-29
Remove Babel Compilation - Drop Support For Node v4


ekryski:
Starting with the next release (Buzzard) we would like to remove Babel compilation from all server side and React Native modules where appropriate and for modules that can be used in the browser  swap over to using Webpack in order to leverage hoisting.

Client side versions of modules are already being moved into their own repos in this release (https://github.com/feathersjs/feathers-client/issues/137), so now seems like the best time to do this.

As a result the next Buzzard release will drop support for Node v4 a bit early. Even though there is still 6 months of official LTS left for v4 from Node, we don't believe many people are using Feathers with Node v4.x. Feathers works without compilation on Node v6+ and Node v8.x is the current stable version.

Node has a good migration path so it shouldn't be a blocker for many people if they do need to upgrade and because we use semver Feathers Auk will continue to work on Node v4+ with Babel compilation so migrating current Feathers apps onto new versions of Node is a non-issue.

**As a result you should probably start now if you are on Node v4.x.**

## Why Change This

- We will be able to install modules from a github repo. This will make it easier for people to have private forks and contribute fixes. 🎉 
- Reduced dependencies are always great 🙌
- Easier debugging and less code bloat
- Less tooling to maintain
- No need for people to compile and/or `watch` files when developing, testing modules, etc.
- We're currently restructuring some modules and breaking out client builds into their own repos

## What Changes

- The next major version of all the modules below will not be supported on NodeJS versions older than v4. You will need to use babel to compile them yourself.
- Switch to webpack for client side builds
- For most modules it is simply the `import` and `export`s that will change and updating the tests

---

The @feathersjs/core-team has decided that this is totally worth but if you feel this is cause for concern please let us know! Feel free to comment or simply give a 👍 or 👎 .

### Core Modules

- [x] feathers https://github.com/feathersjs/feathers/pull/614
- [ ] feathers-query-filters
- [x] ~~feathers-hooks~~ (this is moved into core)
- [x] feathers-commons https://github.com/feathersjs/feathers-commons/pull/53
- [ ] feathers-socket-commons
- [ ] feathers-primus
- [ ] feathers-socketio
- [ ] feathers-rest
- [ ] feathers-service-tests?
- [ ] feathers-hooks-common (need a client build?)
- [ ] feathers-authentication
- [ ] feathers-authentication-local
- [ ] feathers-authentication-oauth1
- [ ] feathers-authentication-oauth2
- [ ] feathers-authentication-ldap
[...]

PR #614, opened 2017-07-07
Update to new plugin infrastructure

daffl:
Update v3 to the latest plugin infrastructure. See https://github.com/feathersjs/feathers/issues/608


package.json
```diff
@@ -27,27 +27,26 @@
     "lib": "lib"
   },
   "scripts": {
-    "prepublish": "npm run compile",
+    "update-version": "node -e \"console.log('module.exports = \\'' + require('./package.json').version + '\\';')\" > lib/version.js",
+    "version": "npm run update-version && git add lib/version.js && git commit -am \"Updating version\"",
     "publish": "git push origin && git push origin --tags",
```
marshallswain: No changelog here?

daffl: We had it in the docs but apparently it disappeared. There is so much noise in the main repo though that I don't think an automated changelog makes a lot of sense in this case.

marshallswain: 👍

PR #619, opened 2017-07-12
typings: properly overload .create()

j2L4e:
done.


j2L4e:
Scratch that. Fixing the overload should do for now. done.

PR #621, opened 2017-07-14
Allows error hooks to swallow error by setting the result

daffl:
This allows error hooks to normally resolve by setting `hook.result` instead of always throwing the error.

Closes https://github.com/feathersjs/feathers-hooks/issues/151


marshallswain:
Looks great!

PR #624, opened 2017-07-18
Add a missing configure() method in typescript definition

jansel369:
## Summary

A little problem with the typescript definition without configure() method in Application interface


eddyystop:
Thanks! We depend on the TS community to keep these up to date.

PR #627, opened 2017-07-21
[WIP] TypeScript: Add ability to define and retrieve properly typed services

j2L4e:
Grants the ability to type static services.

Services returned from app.service() will have the correct type now. Dynamic services will obviously still need casting. Examples:

```ts
import * as feathers from 'feathers';

import {
  GetService, // only implements .get()
  CreateService, // .create()
  FullService // implements all service methods
} from 'feathers';

interface MyInterface {
  id: number
}

interface Services {
  full: FullService<MyInterface>
  getty: GetService<MyInterface> & CreateService<MyInterface> // service implements .get() and .create()
}

const app = feathers<Services>();

app.service('full').find();   // OK
app.service('getty').get(1).then((result: MyInterface) => {});  // OK

app.service('getty').find();                                // TS2339:Property 'find' does not exist on type 'GetService<MyInterface> & CreateService<MyInterface> & ServiceAddons'.
app.service('getty').get(1).then((result: string) => {});   // TS2345: [...] Type 'MyInterface' is not assignable to type 'string'.

// Getting an untyped Service will return type Service<{}> 
// and needs casting (unchanged behaviour)
app.service('anotherService').find();                                  // TS2532:Object is possibly 'undefined'.
```

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,8 +1,8 @@
 import * as express from 'express';
-import * as expressCore from "express-serve-static-core";
+import * as expressCore from 'express-serve-static-core';
 import * as events from 'events';
 
-declare function feathers(): feathers.Application;
+declare function feathers<ServiceListInterface = {}>(): feathers.Application<ServiceListInterface>;
 
 declare namespace feathers {
   export var static: typeof express.static;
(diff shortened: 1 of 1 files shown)


Sieabah:
Would it be possible to define what functions are available on the service through a decorator in that case? All default functions unless specified?

j2L4e:
```
interface Services {
  [key: string]: FullService<any>
}
```
Would be a good way.

I don't see how decorators could achieve what you're looking for.

Sieabah:
Instead of that you can do. I'm pretty sure you could use it to modify the definition of what is available.

`
  interface Services {
    @FeatherService('get', ...)
    getty: Service<MyInterface>

    @FeatherService()
    full: Service<MyInterface>
  }
`

j2L4e:
Decorators and types are totally unrelated. This won't work.

Sieabah:
Apparently I didn't submit my comment after I realized you can't modify types with decorators as they're only evaluated at compile time.

bisubus:
@j2L4e It looks to me like `before` and `after` were deprecated but weren't substituted with `hooks`.

georgeedwards:
OK, so I am getting issue: 
> All declarations of 'Application' must have identical type parameters
, changing [this](https://github.com/feathersjs/feathers/pull/627/commits/b3faa52257487b8f72d564c5aeaa491582cbc1ee#diff-b52768974e6bc0faccb7d4b75b162c99R100) to `_Application` and its references then it all compiles correctly.

But I then have the issue that `Property 'hooks' does not exist on type '_Application<Services>'`?

j2L4e:
Other packages extending the Applicatiin interface would need to get these changes too. That's why it kinda works when you change the name. 

Am 21. September 2017 19:05:30 MESZ schrieb georgeedwards <notifications@github.com>:
>OK, so I am getting issue: 
>> All declarations of 'Application' must have identical type parameters
>, changing
>[this](https://github.com/feathersjs/feathers/pull/627/commits/b3faa52257487b8f72d564c5aeaa491582cbc1ee#diff-b52768974e6bc0faccb7d4b75b162c99R100)
>to `_Application` and its references then it all compiles correctly.
>
>But I then have the issue that `Property 'hooks' does not exist on type
>'_Application<Services>'`?
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/feathersjs/feathers/pull/627#issuecomment-331220038

-- 
Diese Nachricht wurde von meinem Android-Gerät mit K-9 Mail gesendet.

j2L4e:
This should be reimplemented using TypeScript's `Pick` and `Partial` keywords. I'm closing this as I had a talk with daffl about typings and how to do it right with v3.

PR #682, opened 2017-09-27
Stop relying on "this" as the only option in app.configure()

nikolakanacki:
Starting with ES6 it seams weird and unnatural to rely on `this` outside the class methods. I remember contributing to `feathers-hooks` repo on the same issue by making the hook callbacks not rely on `this` as well by putting `hook.service` into use.

No breaking changes are introduced here.

```js
// in app.configure implementation, this:
fn.call(this);
// becomes:
fn.call(this, this);
// so callbacks can become arrow functions like this:
app.configure(app => /* some magic */);
// which is now impossible
```

Wanted to seek others opinions on it, but I see no reason this should be a problem.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1871,9 +1871,9 @@
       "dev": true
     },
     "etag": {
-      "version": "1.8.0",
-      "resolved": "https://registry.npmjs.org/etag/-/etag-1.8.0.tgz",
-      "integrity": "sha1-b2Ma7zNtbEY2K1F2QETOIWvjwFE="
+      "version": "1.8.1",
+      "resolved": "https://registry.npmjs.org/etag/-/etag-1.8.1.tgz",
+      "integrity": "sha1-Qa4u62XvpiJorr/qg6x9eSmbCIc="
     },
     "event-emitter": {
       "version": "0.3.5",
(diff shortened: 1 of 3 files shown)


nikolakanacki:
...as far as I can see the failing tests have nothing to do with this PR.

nikolakanacki:
@daffl I think we should fast forward this because the security issues with the current `express` version, do I need a separate pr or can we just proceed here?

daffl:
Thank @nikolakanacki. Could you split this up into two PRs? One for the Express version update against the `master` branch and the other change against v3 in the [major](https://github.com/feathersjs/feathers/tree/major) branch - it should also include a test like this:

```js
it('passes app to configure', done => {
  const app = feathers();

  app.configure(function(application) {
    assert.equal(app, this, 'Set this reference');
    assert.equal(app, application, 'Passed the app');
    done();
  });
});
```

daffl:
I updated the dependencies in #684. Going to close this one but would love to see the `.configure` change against the `major` branch 😄

nikolakanacki:
@daffl Sorry, didn't have time in the last couple of days to tackle this. Any particular reason why not have this one against the `master` branch since it is 100% backwards compatible?

PR #716, opened 2017-11-01
Add placeholders for channel and publish methods

daffl:
This will give a better error message if someone tries to use event channels e.g. on the client or a server that does not have a real-time transport set up.


daffl:
Leaving this out for now. Might revisit later if it actually is a problem.

PR #719, opened 2017-11-03
Rename expressify to express

bertho-zero:
According with https://github.com/feathersjs/express, and because I find it more appropriate, what do you think?


daffl:
That's how I had it before and someone else pointed out that it was confusing (because you'd expect that to be `require('express')`).

I also think we should keep it at `express` (especially since it has all its middleware on it).
Thanks!

Issue #740, opened 2017-12-01
Make sure real-time events are sent for every item in an array


daffl:
This was happening in v2 at https://github.com/feathersjs/feathers/blob/26306a6bdd844a8fe6dbb498b4ea6a1129fad0f0/src/mixins/event.js#L52 but is not anymore in the new event system.

daffl:
Closed via #743

PR #743, opened 2017-12-01
Updated to handle array emit for service results.

superlazycoder:
Fixed emit to use Array.isArray and emit multiples for array results.
Updated tests with array values to ensure the emits were handled properly.
Refactored the tests to bring the value below the testing limit of 5


superlazycoder:
resolves #740

daffl:
Thank you for doing this! Haven't heard back from Codeclimate so just merged and will see if it can be fixed once I hear back.

Released as v3.0.2

Issue #767, opened 2018-01-02
Swallowing Error Hook


Mattchewone:
I have a hook which will set `context.result` from an error hook. This works when called by an internal server call but does not work when called by an external client request.

I have created the following repo that reproduces the issue here - https://github.com/Mattchewone/feathers-buzzard-error-hooks

If you run the mocha tests in VSCode you can see that the service call sets the `context.result` and the response is correct, but with the client calls the `context.result` is set but the request still returns the error.

Not sure what `context.returnHook` does internally but this is `false` for a server call, but `true` for a client call.

Mattchewone:
Looks like here - https://github.com/feathersjs/express/blob/master/lib/rest/wrappers.js#L60 and https://github.com/feathersjs/socket-commons/blob/master/lib/utils.js#L119 are adding a `true` as a last param, which here - https://github.com/feathersjs/feathers/blob/master/lib/hooks.js#L35 is then setting the `returnHook` to true and is then ignoring the `context.result` and just returning the original error.

daffl:
Indeed, definitely faulty logic there. https://github.com/feathersjs/feathers/pull/769 should fix it and will be published shortly.

Mattchewone:
Thanks @daffl much appreciated.

PR #769, opened 2018-01-02
Properly resolve the promise in error hooks if returnHook is set.

daffl:
When returning the hook object previously it was not possible to swallow the error. This should fix it.

Closes https://github.com/feathersjs/feathers/issues/767


Mattchewone:
@daffl Big thanks for looking into this and fixing it so quickly!

daffl:
Thanks for digging into the issue! Also prompted me to clean up how `returnHook` is being used (it really shouldn't be added to the hook object).

PR #792, opened 2018-01-17
Add ability to skip all following hooks

sylvainlap:
Exports SKIP from `(@feathersjs/commons).hooks.SKIP`.


daffl:
Thank you for doing this! Released as v3.2.0. I'll get the docs updated in a bit.

PR #862, opened 2018-04-28
Event emit skipping

TimNZ:
Refer to discussion here:
https://github.com/feathersjs-ecosystem/feathers-sequelize/issues/188

Want granular control over event emitting after service methods.

You'll likely want to tidy up my test which is hacky.
I'm not knowledgeable on how you test for something NOT happening.

### DIFF
--- a/lib/events.js
+++ b/lib/events.js
@@ -5,6 +5,7 @@ const Proto = require('uberproto');
 // used as the very last hook in the chain
 const eventHook = exports.eventHook = function eventHook () {
   return function (hook) {
+    if (hook.skipEvent) return;
     const { app, service } = hook;
     const eventName = app.eventMappings[hook.method];
     const isHookEvent = service._hookEvents && service._hookEvents.indexOf(eventName) !== -1;
--- a/test/events.test.js
+++ b/test/events.test.js
@@ -278,4 +278,31 @@ describe('Service events', () => {
         .then(() => service.emit('created', { message: 'custom event' }));
     });
   });
+
+  describe('event skipping', () => {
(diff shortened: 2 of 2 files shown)


daffl:
Thank you for the PR @TimNZ. Skipping even emitting should however already be possible by returning [feathers.SKIP](https://docs.feathersjs.com/api/hooks.html#returning-feathersskip) in the last hook you want to run.

TimNZ:
I knew about SKIP but didn't test well enough if it could achieve what I wanted.

Have tested nested service calls and a simple hook and modifying context/params achieves desired flow.

Though if any other 'finally' hooks ever get added in the future this will break them.
That's why I prefer the explicit event skipping, doesn't depend on any particular hook ordering

TimNZ:
If you decide not to have explicit event skipping support, 
perhaps a note in the SKIP section and Events page docs that SKIP prevents event emitting as well.
https://docs.feathersjs.com/api/hooks.html#returning-feathersskip

My event logging module is separate from hooks handling, I like the modular micro approach you've taken and don't want to introduce dependencies.

daffl:
Haven't forgotten about this. Just figuring out if and how this will fit into the next release.

TimNZ:
@daffl Thanks.  

No rush.

A context flag to skip events processing, decoupled from hook flow, still makes most sense to me.

daffl:
Allright, so what I'd like to allow for this release would be adding a hook `context.event` which will contain the service event name that will be sent once the hook completes successfully. That way it would be possible to change the event that is emitted or prevent any events from being sent by setting `context.event = null`.

TimNZ:
Does the job.

Issue #863, opened 2018-05-01
It is not possible to use a Symbol as a property key on a service.


bertho-zero:
### Steps to reproduce

```js
const feathers = require('@feathersjs/feathers');
const prop = Symbol('test-it');

const app = feathers()
  .use('users', {
    create() {},
    [prop]: 'ok'
  });

const service = app.service('users')
console.log(service[prop]); // undefined
// should be 'ok'
```

### Expected behavior
It should be possible to recover a value with a Symbol key.

### Actual behavior
The key disappears.

### System configuration

**Module versions** (especially the part that's not working):
`@feathersjs/feathers@^3.1.4`

**NodeJS version**:
v8.11.1

**Operating System**:
Ubuntu 16.04

daffl:
It's because `Object.getOwnPropertyNames` does not include symbols (which is happening at https://github.com/daffl/uberproto/blob/master/lib/proto.js#L81).

daffl:
Closed via https://github.com/feathersjs/feathers/pull/880, https://github.com/daffl/uberproto/pull/13 and https://github.com/feathersjs/feathers/pull/881

PR #864, opened 2018-05-01
Open hook workflow to custom methods

bertho-zero:
**Works with https://github.com/feathersjs/commons/pull/72.**

This PR allows you to add converters to your custom methods, which allows to create the hook corresponding to your arguments and finally to fully use the hook workflow.

I would have liked to use a Symbol but https://github.com/feathersjs/feathers/issues/863.

Example:
```js
const app = feathers().use('/dummy', {
  methods: {
    custom: ['id', 'data', 'params']
  },
  get() {},
  // ...
  custom(id, data, params) {
    return Promise.resolve([id, data, params]);
  }
});

app.service('dummy').hooks({
  before: {
    all: [/**/],
    custom: [/**/]
  },
  after: {
    all: [/**/],
    custom: [/**/]
  }
});

app.service('dummy').custom(28, { bar: 'foo' }, { provider: 'rest' });
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,6 @@ node_modules
 
 # Users Environment Variables
 .lock-wscript
+
+# IDEs
+.idea
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 sudo: false
 language: node_js
+install: npm install
 node_js:
   - node
   - '6'
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ node_js:
   - '6'
 addons:
   code_climate:
-    repo_token: c2aa7db42f71ed803abb928688f20dfe3d2d397fe9b5d6392cf65aacbc6382ed
+    repo_token: 9840b8b56f6d10209b3478d41f7ba5102d02981435df129bb18554a529359c62
 before_script:
   - 'npm install -g codeclimate-test-reporter'
 after_script:
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,12 @@
 # Change Log
 
+## [v3.1.4](https://github.com/feathersjs/feathers/tree/v3.1.4) (2018-03-26)
+[Full Changelog](https://github.com/feathersjs/feathers/compare/v3.1.3...v3.1.4)
+
+**Merged pull requests:**
+
+- Make sure error hooks always have the original context information [\#842](https://github.com/feathersjs/feathers/pull/842) ([daffl](https://github.com/daffl))
+
 ## [v3.1.3](https://github.com/feathersjs/feathers/tree/v3.1.3) (2018-02-16)
 [Full Changelog](https://github.com/feathersjs/feathers/compare/v3.1.2...v3.1.3)
 
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,22 @@
 # Change Log
 
+## [v3.1.7](https://github.com/feathersjs/feathers/tree/v3.1.7) (2018-06-16)
+[Full Changelog](https://github.com/feathersjs/feathers/compare/v3.1.6...v3.1.7)
+
+**Merged pull requests:**
+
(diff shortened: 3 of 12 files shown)


bertho-zero:
And then we can take advantage of authentication, validation, formatting, parsing, etc. without any other effort:

```js
// for async/await in express middleware
const wrap = fn => (req, res, next) => Promise.resolve(fn(req, res, next)).catch(next);

app.use('/custom', wrap(async (req, res) => {
  const result = await app.service('dummy')
    .custom(1, req.body, {
      query: req.query,
      provider: 'rest'
    });
  res.send({ result });
}));
```

bertho-zero:
I just thought of another syntax:
```js
const app = feathers().use('/dummy', {
  get() {},
  // ...
  @enableHooks(['id', 'data', 'params'])
  custom(id, data, params) {
    return Promise.resolve([id, data, params]);
  }
});

// or without decorator

const app = feathers().use('/dummy', {
  get() {},
  // ...
  custom: enableHooks(['id', 'data', 'params'])(
    (id, data, params) => {
      return Promise.resolve([id, data, params]);
    }
  )
});

```

Which do you prefer between this one and that of the PR?

bertho-zero:
I think that all methods can be added to `service.methods`, for example the argument list can be used in `@feathersjs/express` to convert the request into arguments.

https://github.com/feathersjs/express/blob/9bc5541107bfd1ec58d3e38fb968b91832cda377/lib/rest/wrappers.js#L111-L116

daffl:
I like the

```js
const app = feathers().use('/dummy', {
  get() {},
  // ...
  custom: activateHooks(['id', 'data', 'params'])(
    (id, data, params) => {
      return Promise.resolve([id, data, params]);
    }
  )
});
```

Syntax. Now that you got Symbols working, is there anything that would change?

bertho-zero:
I just had to rename `activeHooks` to `activateHooks`, it's all good.
Just publish a version with the latest `@feathers/commons` changes and tests pass.

bertho-zero:
`@feathersjs/express` for example, could use a function like 
```js
function makeArgsGetter (argsOrder) {
  return (req, params) => argsOrder.reduce((result, argName) => {
    switch (argName) {
      case 'id':
        return [ ...result, req.params.__feathersId || null ];
      case 'data':
        return [ ...result, req.body ];
      case 'params':
        return [ ...result, params ];
    }
  }, []);
}
```
and `argsOrder` comes from `service.methods` or `Symbol`.

Using the Symbols implies that you need a require of `@feathersjs/feathers` to compare with the original Symbol, perhaps an https://docs.npmjs.com/files/package.json#optionaldependencies.
Symbols can only be used in `@feathersjs/feathersjs` if you pass all methods by `service.methods` (not just customs)

daffl:
I'll review this again over the weekend. The problem with Symbols is that they may not work in the browser (although I _think_ that the Polyfill Feathers requires should work for most).

bertho-zero:
Thank you. Symbols have polyfills and feathers no longer explicitly require babel-polyfill but we can put a note for those who want to use this feature.

Symbols are not compatible with all browsers, be sure to precompile your code or add a Polyfill for older browsers.
In other cases your services will have a `__feathersActivateHooks` property.

bertho-zero:
Perhaps you could release a major beta version of `@feathersjs/common` and a version of `@feathersjs/feathers` on a tag that is not "latest" ? It will allow to test that in production.

This change is a new feature without regression or change of API, it could even be published in a patch version without waiting for the Crow version.

daffl:
You are totally right. Sorry I'm so slow - that happens with me sometimes during summertime 😉 - I just released a version 2.0.0 of https://github.com/feathersjs/commons. Could you update the pull request pointing to that version and fix the conflicts? Then everything should pass right? How does #875 fit in here?

bertho-zero:
#875 is apart, I'll update it after.

daffl:
They should go out in the same version though right?

bertho-zero:
Not necessarily, It can go out in another version.
I'll update it next week.

PR #865, opened 2018-05-03
Allow methods to return a null result

bertho-zero:
I do not know if this modification will break the tests of other packages of the ecosystem (I do not think) but it seems normal to me to be able to return null value without it being an error.

In my case I need to return null when I get a `softDelete`d item instead of throw an error, the following code works with this PR:
```js
// For return `null` value after a get on a soft deleted event I need to avoid error
error( context ) {
  if ( context.error instanceof errors.NotFound && context.error.message === 'Item not found.' ) {
    context.result = null;
    return context;
  }

  log.error(
    `Error in '${context.path}' service method '${context.method}'\n${context.error.stack}\n`,
    inspect( _.omit( context.error, [ 'hook.app', 'hook.service' ] ), { colors: true } )
  );
}
```

This can also be useful in a before hook for pass the service call and return null.

### DIFF
--- a/lib/hooks.js
+++ b/lib/hooks.js
@@ -100,7 +100,7 @@ const hookMixin = exports.hookMixin = function hookMixin (service) {
           // A shallow copy of the hook object
           const errorHookObject = Object.assign({}, error.hook, hookObject, {
             type: 'error',
-            result: null,
+            result: undefined,
             original: error.hook,
             error
           });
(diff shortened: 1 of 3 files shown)


bertho-zero:
In my opinion result should be undefined, null can be a result.

daffl:
Totally agree. Sorry, my edit broke the build. We should definitely have a test for this, too.

daffl:
Released as 3.1.5, thank you!

bertho-zero:
Do not you think that a service call should be able to return null as a value? without going through the error hook.

This solution goes in the right direction but I am forced to create an error that is not one to change the result to null, which stops the chain of hooks in passing.

For me:
- undefined = not defined
- null = a possible result

daffl:
`null` return values for service should work just fine already.

bertho-zero:
You're right, after a more minimalist test it works.
(https://runkit.com/embed/aumy9wl9tzmz)

PR #874, opened 2018-05-27
Exports withHooks

bertho-zero:
This method also allows you to add hooks for a single call.:
```js
await withHooks({
  app,
  service: svc,
  method: 'create'
})({
  before: {/**/},
  after: {/**/},
  error: {/**/},
  finally: {/**/}
})(data)
```


bertho-zero:
The equivalent of this PR can be merged with https://github.com/feathersjs/feathers/pull/864 + https://github.com/bertho-zero/feathers/pull/2.

PR #880, opened 2018-06-02
Add 'services conserve Symbols' test

bertho-zero:
Here is a test that corresponds to the problem described in #863.


daffl:
If you rebase with master this should pass now. Thank you!

PR #924, opened 2018-08-06
withHooks + custom methods

bertho-zero:
https://github.com/feathersjs/feathers/pull/864 + https://github.com/feathersjs/feathers/pull/875

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,6 @@ node_modules
 
 # Users Environment Variables
 .lock-wscript
+
+# IDEs
+.idea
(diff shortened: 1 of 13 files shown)


beeplin:
@bertho-zero @daffl any docs on this new feature? Previously I was using my own solution for custom methods (between client and server), but apparently feathers 3.2.x broke my old solution. I have checked #864 and #875, though roughly. Any more documentation for me to dig in? Thanks~

daffl:
Maybe @bertho-zero can add an example here. We also talked about this a little more and in the future with https://github.com/feathersjs/feathers/issues/932 this will be much easier by being able to apply hooks to any asynchronous method.

Issue #462, opened 2016-11-16
RFC: Use Lerna with monorepo for core packages


justingreenberg:
as the feathers ecosystem continues to evolve, it is becoming increasingly difficult for both users and maintainers to track dependencies, coordinate tandem releases, keep docs in sync, etc.

it is also difficult to at a glance to differentiate between core feathers and external plugins in the feathers organization 

tl;dr: i recently used https://github.com/lerna/lerna (used by `babel`, `react`, `angular`, `meteor` etc) to migrate a large back office project with many custom npm modules into a "mono-repo", and it was *incredibly* simple. it automatically creates symlinks, i cannot express how much simpler maintenance is and how much time this has saved our developers....

```
feathers/       # https://github.comfeathersjsfeathers
    /scripts/    # manage releases, etc
    /packages/   # core feathers modules
        /feathers
        /feathers-client
        /feathers-commons
        /feathers-hooks
        /feathers-query-filters
        /feathers-rest
        /feathers-service-tests
        /feathers-socket-commons
        /feathers-authentication
        /feathers-authentication-client
        /feathers-authentication-jwt
        /feathers-authentication-local
        /feathers-authentication-oauth1
        /feathers-authentication-oauth2
        /feathers-authentication-permissions
        /feathers-authentication-popups
        /feathers-knex
        /feathers-localstorage
        /feathers-memory
        /feathers-mongodb
        /feathers-mongoose
        /feathers-nedb
        /feathers-sequelize
        /feathers-batch
        /feathers-bootstrap
        /feathers-configuration
        /feathers-hooks-common
        /feathers-socket-commons
        /feathers-swagger
```

or it could further be broken down by functionality...

```
feathers/       # https://github.comfeathersjsfeathers
    scripts/    # manage releases, etc
    packages/   # core feathers modules
        core/
            feathers
            feathers-client
            feathers-commons
            feathers-hooks
            feathers-query-filters
            feathers-rest
            feathers-service-tests
            feathers-socket-commons
        authentication/
            feathers-authentication
            feathers-authentication-client
            feathers-authentication-jwt
            feathers-authentication-local
            feathers-authentication-oauth1
            feathers-authentication-oauth2
            feathers-authentication-permissions
            feathers-authentication-popups
        storage-adapters/
            feathers-knex
            feathers-localstorage
            feathers-memory
            feathers-mongodb
            feathers-mongoose
            feathers-nedb
            feathers-sequelize
        plugins/
            feathers-batch
[...]

PR #938, opened 2018-08-21
Lerna: Add Feathers transport modules

daffl:
For #462. Adds the following Feathers transport modules to the repository:

- `@feathersjs/express`
- `@feathersjs/transport-commons`
- `@feathersjs/socketio`
- `@feathersjs/primus`


package.json
```diff
@@ -32,6 +32,7 @@
     ]
   },
   "devDependencies": {
+    "axios": "^0.18.0",
```
bertho-zero: Should not `axios` be in the `express` package.json instead?

mocha.opts
```diff
@@ -1,2 +1,2 @@
---recursive packages/**/test/**/*.test.js
+--recursive packages/*/test/**/*.test.js
```
bertho-zero: You can also add `--exclude **/node_modules/**` for this.

package.json
daffl: My approach was to move any `devDependency` that is (or will) be used in more than one repo to the top. Another small thing I'd like to do eventually was to unify the HTTP lib (currently there  `axios` and `request` seem to be used)

mocha.opts
daffl: Ah, yes, that was the problem I was having. Probably makes more sense to do this explicitly.

package.json
bertho-zero: I think that package devDependencies should remain in packages, and be added to the top if they are used outside these packages.

To avoid having duplicates, you can use lerna's '--hoist' option, and you'll go faster by using yarn instead of npm.

daffl: Makes sense, I started this because they recommended it in the Lerna Readme at https://github.com/lerna/lerna#common-devdependencies

bertho-zero: The reasons given are no longer valid since there is the `--hoist` option and the yarn workspaces.

PR #945, opened 2018-08-22
express: use service.methods

bertho-zero:
Apply feathersjs/express#29 `.diff`


bertho-zero:
@daffl I resolved the conflict with master.

bertho-zero:
@daffl I merged master and all the tests seem to pass.

daffl:
So the difficulty with this one is that we can't force the dependent Feathers version unless it is a major (breaking) change. If someone has e.g. `@feathersjs/feathers@3.1.x` things will break (already seeing some issues around that, e.g. in #1000).

bertho-zero:
`@feathersjs/express` probably requires a major version.

daffl:
I think so, too. Ok, I'll get this in with auth v3 (since authentication Express middleware will move here, too). It's going pretty well so it hopefully shouldn't be too long until a pre-release.

Mairu:
Please don't forget this for 4.0

PR #969, opened 2018-08-30
Use hoisted package-lock.json and remove files from sub-packages

daffl:
Decreases installation size and otherwise Lerna will detect a change in each package every time we run `npm install`.


package.json
```diff
@@ -19,7 +19,7 @@
     "node": ">= 6"
   },
   "scripts": {
-    "install": "lerna bootstrap",
+    "install": "lerna bootstrap --hoist --no-ci",
```
bertho-zero: Should not these options be in the package.json? (https://github.com/lerna/lerna#lernajson)

PR #971, opened 2018-08-30
prefer lerna.json

bertho-zero:
For https://github.com/feathersjs/feathers/pull/969


daffl:
The `package-lock.json` files should not be added to the sub-packages (see comment in #969). Otherwise after every `npm install` Lerna will detect everything as a changed module even though no code was actually changed.

daffl:
There also is a discussion around this in https://github.com/lerna/lerna/issues/913

bertho-zero:
How can I fix https://travis-ci.org/feathersjs/feathers/jobs/422724278#L496?

daffl:
That's why I added `--no-ci` to the command line which does not seem to be supported by `lerna.json`.

daffl:
It would be nice to get it working with `npm ci` eventually. I'll create an issue for the updates I think we would want to make to the test setup.

bertho-zero:
It seems that `ci` must be at the top level of lerna.json: https://github.com/lerna/lerna/blob/master/commands/bootstrap/__tests__/__fixtures__/basic/lerna.json

Issue #925, opened 2018-08-06
Database adapter security and usability improvements


daffl:
[Feathers database adapters](https://docs.feathersjs.com/api/databases/adapters.html) cover a wide range of standard CRUD functionality. Although [the standard behaviour is documented](https://docs.feathersjs.com/api/databases/adapters.html), their currently very open nature can make it easy to miss adding certain restrictions in order to secure your application properly. To improve this, I am planning to implement the following breaking changes:

## Single items and `params.query`

`get`, `remove`, `update` and `patch` will respect the `query` object additionally to the `id`. This will make it much easier to limit access for users by setting the `query` with limitations (e.g. an owner `userId`) much like it is already done for normal `find` calls.

```js
// Will throw 404 if `userId` is not the same for the entry with `accountid`
app.service('accounts').get('accountid', {
  query: { userId: 'authenticated user id' }
});
```

## Multiple updates

`create` with arrays and multiple updates (`patch`, `update` and `remove` with id `null`) will be disabled by default. They can be enabled explicitly by setting the `multi` options to `true` (or the list or methods to allow). It is then up to the developer to make sure that those calls are secured properly _and_ all affected hooks are also taking result arrays into consideration.

```js
const memory = require('feathers-memory');

app.use('/messages', memory({
  multi: true,
  paginate: {
    default: 2,
    max: 4
  }
}));

app.use('/messages', memory({
  // Only allow multiple `create` and `patch`
  multi: [ 'create', 'patch' ],
  paginate: {
    default: 2,
    max: 4
  }
}));
```

## Non Feathers query syntax

All adapters support [Feathers common query syntax](https://docs.feathersjs.com/api/databases/querying.html). However, some adapters like `feathers-nedb`, `feathers-mongodb` and `feathers-mongoose` allow passing other queries as well. The next version of all database adapters will no longer allow any non-Feathers queries that are directly passed to the database. Instead, additional database specific query properties that should be allowed have to be explicitly whitelisted and secured appropriately:

```js
const mongoose = require('feathers-mongoose');

app.use('/messages', mongoose({
  // Allowing $populate in Mongoose can expose protected properties!
  // They now have to be removed manually in this service
  whitelist: [ '$populate' ],
  paginate: {
    default: 2,
    max: 4
  }
}));
```

## Improved `feathers-service-tests`

[...]

PR #1010, opened 2018-09-19
Added generators for feathers-objection & feathers-cassandra

dekelev:


daffl:
This is great, thank you! I'll have a look at it on the weekend.

In order to keep them as official adapters in the future, do you plan on including the changes coming in https://github.com/feathersjs/feathers/issues/925?

dekelev:
Sure, these changes makes a lot of sense.

`feathers-objection` allows eager queries, like mongoose's $populate, but for example, `allowedEager: 'todos'` in the `users` service options acts as the whitelist and will only allow getting `user.todos`, using `query.$eager = 'todos'`.
`allowedEager: '[todos, account]'` will also allow getting the `account` relation.

`feathers-objection` also has `Upsert/Insert` feature, which will update relations too. this requires the `allowedUpsert`, `allowedInsert` and `createUseUpsertGraph` service options.

`feathers-cassandra` doesn't support relations, so I can't think of anything there that needs whitelisting.

Both packages should already respect the `query` object in `get`, `remove`, `update` and `patch`, but I'll test this to verify.

I'll add support & docs for the `multi` service option in both projects.

Let me know if you can think of anything else. enjoy your weekend!

daffl:
I'll keep you in the loop what the plan is with those updates and will try to get your PR in soon.

daffl:
Released in `@feathersjs/cli@3.8.5`

Issue #1001, opened 2018-09-14
Uncaught (in promise) TypeError: Cannot convert undefined or null to object


daffl:
_From @krthr on August 30, 2018 23:4_

### Steps to reproduce

- [ ] populateAccessToken functions isn't getting the accessToken token property. 

### Expected behavior
Login and create a new resource from the client.

### Actual behavior
Login and getting and error when i try to create a new resource:

```bash
Uncaught (in promise) TypeError: Cannot convert undefined or null to object
    at Function.assign (<anonymous>)
    at Object.eval (populate-access-token.js?39da:9)
    at promise.then.hookObject (hooks.js?8408:142)
(anonymous) @ populate-access-token.js?39da:9
promise.then.hookObject @ hooks.js?8408:142
...
```

**Module versions** (especially the part that's not working): 
- "@feathersjs/authentication-client": "1.0.3"
- "@feathersjs/feathers": "3.2.0"
- "@feathersjs/socketio-client": "1.1.1"

**NodeJS version**: 8.9.4
**Operating System**: Windows 10

_Copied from original issue: feathersjs/authentication-client#108_

daffl:
_From @krthr on August 30, 2018 23:5_

When i type `ApiClient.get('accessToken')` in the console I get the accessToken

daffl:
What logs do you get if you add

```
app.hooks({
  before(context) {
    console.log(context);
    return context;
  }
});
```

In your client side application? Also, does this still happen when explicitly installing `@feathersjs/feathers@3.1.7`?

daffl:
_From @Torge on August 31, 2018 19:2_

I got the same error. Its on the clientside for me. I think the problem isn't the access token but hook.params. It is undefined for me.
populateAccessToken tries to Object.assign it.
```
    Object.assign(hook.params, { accessToken: app.get('accessToken') });
```
context in populateAccessToken:
```
"{
  "type": "before",
  "method": "create",
  "path": "draft",
  "data": {}
}"
```
Setting the @feathersjs/feathers version from 3.2.0 to 3.1.7 solved the issue for me.

context in populateAccessToken @ 3.1.7:
```
{
  "type": "before",
  "method": "create",
  "path": "draft",
  "params": {},
  "data": {}
}
```

daffl:
I am not able to reproduce this with a new client build in the [feathers-chat](https://github.com/feathersjs/feathers-chat). How does your client side application setup look like?

daffl:
[...]

PR #1011, opened 2018-09-20
Consider that hook.params might be falsy

testless:
### Summary

Compare with issue https://github.com/feathersjs/feathers/issues/1001


daffl:
Thank you for the PR @testless. As I mentioned in https://github.com/feathersjs/feathers/issues/1001#issuecomment-423228192 this should never happen since Feathers normalizes those parameters internally. It would be good to have a breaking test for this first, otherwise we'd end up just treating a symptom but not the problem.

daffl:
Ok, found the root cause and fixed it in #1012. Thank you for digging into this!

PR #1041, opened 2018-10-05
Allow case-insensitive for local email (username) authentication

SwiftPolar:
### Summary

In most cases usernames and emails are case-insensitive. Current implementation prevents users from authenticating successfully when the emails do not match exactly (i.e. case sensitive).

### Other Information

This is just a minor change but it makes more sense to the end-user and developer as usernames in most cases do not require you to be case sensitive in the username field.


daffl:
Makes sense, thank you!

Issue #1049, opened 2018-10-09
Patching entity through service causes the socket entity to be wrongly updated


claustres:
### Steps to reproduce

> This behavior seems to have appeared after the migration to V3.

Authenticate a user which data model has two properties `email` and `name`.  Patch one of the property like `app.service('users').patch(id, { name: 'xxx' })`. After patch the user entity attached to the socket has lost the `email` property. It will cause any hook or operation depending on it to stop working on next service call.

### Expected behavior

The socket entity should be correctly patched.

### Actual behavior

The socket entity is updated instead of being patched.

[This code](https://github.com/feathersjs/feathers/blob/master/packages/authentication/lib/socket/handler.js#L170) could be the problem since the same function is called for both update or patch operations and this function [delete properties based on the input object](https://github.com/feathersjs/feathers/blob/master/packages/authentication/lib/socket/update-entity.js#L34).

### System configuration

**Module versions**:
```
"@feathersjs/authentication": "^2.1.7",
"@feathersjs/authentication-jwt": "^2.0.1",
"@feathersjs/authentication-local": "^1.2.1",
"@feathersjs/authentication-oauth2": "^1.1.0",
"@feathersjs/client": "^3.5.3",
"@feathersjs/configuration": "^1.0.2",
"@feathersjs/errors": "^3.3.0",
"@feathersjs/express": "^1.2.3",
"@feathersjs/feathers": "^3.1.7",
"@feathersjs/socketio": "^3.2.2"
```

**NodeJS version**: 8

**Operating System**: Windows

claustres:
If I am not wrong about this issue the proposed patch is to still call the `updateEntity` function but with an additional parameter indicating the operation. Based on this we could skip the code starting from [here](https://github.com/feathersjs/feathers/blob/master/packages/authentication/lib/socket/update-entity.js#L34) in case of a patch. Could work on a PR but I have to learn about lerna first I guess.

claustres:
I wonder about one thing with lerna, if I do a fork of feathers to patch eg the authentication module, then submit a PR, then how can I use my fork in my app waiting for the release ? Previously I could point my `package.json` toward the master branch of the `feathers-authentication` repo but now everything is in the same repo is it still possible ?

claustres:
I ran into issues with the new lerna repo.

First `npm run test` does not work saying `'..' is not a recognized command` when running `../../node_modules/.bin/mocha --opts ../../mocha.opts`.

[...]

Issue #845, opened 2018-03-28
Generate Jest tests instead of Mocha?


danawoodman:
Is there anything on the roadmap to support generating Jest tests instead of Mocha tests? Jest is getting quite popular and has become my defacto testing library. Right now Jest users will have to completely replace the generated tests. Perhaps a config option could be set and Feathers could generate Jest style tests instead? Just a thought!

daffl:
It is not on the roadmap but as with most generator improvements, I would be happy to accept a pull request that adds it to the generator.

Test files are just generated (as in there is no code transformations) so it should be fairly straightforward, the only thing would be how to figure out what testing library the current project is using. This could probably be done by checking `const testLib = pkg.devDependecies.jest ? 'jest' : 'mocha';`

rtyler:
In case anybody stumbles into this ticket some time before @danawoodman contributes a pull request :smiling_imp: , the only major change between using `jest` and `mocha` for feathersjs is changing `before(() => { });` and `after(() => { });` to `beforeEach` and `afterEach` respectively. (see [the jest documentation for more](https://facebook.github.io/jest/docs/en/setup-teardown.html)).

Otherwise the mocha-generated test  stubs are _pretty_ compatible out of the box.

daffl:
This has been done via https://github.com/feathersjs/feathers/pull/1057

eric-burel:
Hi,
Update from an existing app is not obvious so I'll post it here for googlers:
- add a script named `jest` in your package.json. Feathers cli will look for the existence of this script to select either jest or mocha (default); `"jest":"jest"`
- install Jest as a devDependency : `npm i --save-dev jest`

You're good to go, from now on the CLI will generate a jest test. Of course you'll need to update your old mocha tests for Jest.

Note that this does not sound either obvious or scalable to other test frameworks. I'd rather have an explicit "config" object (or even "feathers" if it's allowed?) in package.json to configure such choices, like it seems you did in a first iteration, instead of an auto-discovery.
Still, thanks a lot for this feature, using Jest is really pleasurable especially for full-stack since it's used a lot in the React ecosystem.

PR #1057, opened 2018-10-15
@feathers/cli: introduce option to choose jest for tests instead of mocha

morphatic:
### Summary

Addresses #845. (Also mentioned in [#1049](https://github.com/feathersjs/feathers/issues/1049#issuecomment-428652715))

This PR affects the `@feathers/cli` package (more specifically the generators in `generators-feathers`). In particular, the three generators affected are:
1. **`feathers generate app`**
Gives you the option to choose `jest` instead of `mocha` for testing.
2. **`feathers generate service`**
The auto-generated test uses whatever testing framework was selected when the app was generated.
3. **`feathers generate hook`**
Same as for services.

Usage should be obvious for app creation and transparent for service/hook creation. In addition to small changes in the syntax of the tests themselves (e.g. `before()` becomes `beforeAll()` and asynchronous tests require `expect.assertions([some number])`) it was necessary to modify two other files:
1. **`.eslintrc.json`** had to be renamed `eslintrc.json.js` and moved from the `generators/app/templates/static/` folder to the `generators/app/configs` folder. That's because `eslint` has to be told which testing framework will be used in the `env` property.
2. **`generators/app/configs/package.json.js`** has a number of changes:
    1. I added `NODE_ENV= ` to the `scripts.test` property. This was to squelch some unnecessary warning messages that were being generated when tests were run [described here](https://github.com/lorenwest/node-config/issues/386). Please note the space after the `=` sign. I admit this is a weird thing to put in your `test` script, but the worst that will happen if you take it out is that you either have to put up with unnecessary warnings in the console, or you have to add an unnecessary `test.json` file in the `./config` folder. Another option would be to [actually use the `test.json` file for something](https://generator.feathers-plus.com/get-started/#folders).
    2. ~~I added a `config.tester` property. As best as i can tell, it is totally [legitimate to put arbitrary properties under `config` in `package.json`](https://docs.npmjs.com/files/package.json#config) as long as you don't conflict with [pre-defined values](https://docs.npmjs.com/misc/config). [This StackOverflow thread recommends namespacing](https://stackoverflow.com/a/27232456/296725) any custom config properties. I can see an argument for not polluting `package.json` with a lot of extra config values, but this just seemed cleaner and easier than checking the keys of the `scripts` object to figure out which testing framework is in use.~~ After I read #845 more carefully, I saw that @daffl had a [better solution for figuring out what testing framework is being used](https://github.com/feathersjs/feathers/issues/845#issuecomment-377007196) so I switched to that.
    3. The name of the testing property that gets added to the `scripts` is now dependent upon user input. It defaults to `mocha` but allows the user to select `jest` instead.

I believe there are zero changes that would impact any other parts of either the CLI or the generated feathers apps. FYI, it won't hurt my feelings if this PR gets modified or rejected. This is my first time contributing to the project, and I tried as best to follow the style of the code upon which my edits were based. I think feathers is a pretty fantastic contribution to the world, and I'm happy to see all the organizational work that went into it this summer. Here's my small contribution to keeping it vital!

- [x] Are there any open issues that are related to this? #845
- [x] Is this PR dependent on PRs in other repos? **No.**

### Other Information

I added new tests based on the existing ones. All tests pass. The new ones are in `test/generators.using.jest.test.js`. I set them up to automatically `skip()` the tests for `cassandra` and `rethinkdb` since they both fail if the developer doesn't have those database servers installed and running on their systems. (BTW, I'd never heard of either server, so figuring out why my code wouldn't pass those tests got me hung up for quite a bit of time 😎).

### Note on change to `package.json`

In order to get the CI tests to pass, I had to update the `test` script in `package.json` to use the flag `--no-timeouts` instead of `--timeout 30000`. On my local system, it would also work with the timeout value set to 60000, but at that point, it seemed just easier to remove the timeouts altogether.

In any case, [**running the full lerna test suite on Node 6 OVER 40 MINUTES!!**](https://travis-ci.org/feathersjs/feathers/builds/441501555) I'd like to suggest that we need to figure out a way to expedite these tests. That is too long to have to wait to find out whether there's a bug in your code or not. When running locally, I could use `--grep [pattern]` to restrict my tests to just the ones I cared about, but that doesn't work with Travis.


morphatic:
Small gripe about codeclimate. The configuration flagged the following two bits of code as "too similar" and in need of refactoring:

```javascript
// from /packages/generator-feathers/generators/app/index.js
module.exports = class AppGenerator extends Generator {
  // ...
  prompting() {
    // ...
    const prompts = [{
      // ...
    }, {
      name: 'packager',
      type: 'list',
      message: 'Which package manager are you using (has to be installed globally)?',
      default: 'npm@>= 3.0.0',
      choices: [
        { name: 'npm',  value: 'npm@>= 3.0.0'   },
        { name: 'Yarn', value: 'yarn@>= 0.18.0' }
      ]
    }, {
      // ...
    }, { // THIS block is apparently "too similar" to the 'packager' block above
      name: 'tester',
      type: 'list',
      message: 'Which testing framework do you prefer?',
      default: 'mocha',
      choices: [
        { name: 'Mocha + assert', value: 'mocha' },
        { name: 'Jest',           value: 'jest'  }
      ]
    }];
    // ...
  }
}
```

It prevented my PR checks from passing until I had figured out a way to get codeclimate not to see them as similar. Really the similarity between these two blocks actually makes the code easier to read and maintain. In order to get codeclimate to leave me alone (notice that 4 out of the last 5 commits in this PR are tweaks to try to please codeclimate), I had to add an unnecessary property to the second block. It meant I had to dig into the Yeoman Generator code to discover that it uses Inquirer.js to generate its questions, which in turn has good documentation (but not excellent), so I had to dig into that code to discover that I could just add a `pageSize: 7` property to the question object, which essentially does nothing to the functionality but fools codeclimate into thinking the two blocks are not "too similar" anymore.

I totally get the purpose of codeclimate, but in this case, it seems to have been a hindrance to using best practices rather than a help. 😎

packages/generator-feathers/generators/service/index.js
```diff
@@ -20,51 +20,26 @@ module.exports = class ServiceGenerator extends Generator {
         message: 'What kind of service is it?',
```
daffl: Are these formatting changes in this file from your linter?

packages/generator-feathers/test/generators.using.jest.test.js
```diff
@@ -0,0 +1,125 @@
...
+        })
+        .then(() => child.kill())
+        .catch(e => new Promise((resolve, reject) => {
+          child.once('exit', () => reject(e));
+          child.kill('SIGKILL')
+        }));
+      });
+    };
+
+    it('generic',   () => testServiceGenerator('generic'));
```
daffl: We can probably remove the other database tests here and assume everything will work fine to not make the tests run a little quicker.

packages/generator-feathers/generators/service/index.js
morphatic: No. The CodeClimate checks originally failed because ["Function prompting has 91 lines of code (exceeds 80 allowed)."](https://codeclimate.com/github/feathersjs/feathers/pull/1057) I actually thought the formatting was better as it was. I was looking for ways to reduce the number of lines without changing behavior. I would be happy to change it back to the way it was.

packages/generator-feathers/test/generators.using.jest.test.js
morphatic: Cool. Skipping all but `generic` test reduced overall testing times on travis from over 40 minutes down to about 15 minutes. Not sure why the latest build failed, but the offending code was the `authentication` package, not the generators. Also, you may already know this, but apparently you can tell `lerna` to [only run the tests for the packages that were most recently updated](https://github.com/lerna/lerna/issues/1263). ([official docs](https://github.com/lerna/lerna/tree/master/core/filter-options)) I had never heard of `lerna` before I got into this project. It's pretty cool!

daffl:
Ah yes, sorry it's a flaky test that will be fixed soon. I think moving to the monorepo some of the Codeclimate settings got screwed up, I'll have to revisit those as well. Will merge and release this later this weekend. Thanks again for doing this, great work!

PR #1063, opened 2018-10-17
support a secretProvider

kevinranks:
Hello,

I hope you accept this simple PR that adds support for a `secretOrKeyProvider` as documented at [passport-jwt#usage](https://github.com/themikenicholson/passport-jwt#usage). 

The change simply checks if the passed secret is a function or not. It will then it will set the appropriate key (`secretOrKey` or `secretOrKeyProvider`) as defined in passport-jwt.

Thanks!


daffl:
This makes sense, thank you. Could you also add the option [in the documentation](https://docs.feathersjs.com/api/authentication/jwt.html#options). How would a test for this look like?

kevinranks:
@daffl Can you rebuild this? Not sure why the client tests failed...

Issue #987, opened 2018-09-09
config folder at root instead of custom folder


woodgates:
### Steps to reproduce
Hi.
I have this repo trying to make nuxt to work with feathers3.2 and its CLI.
https://github.com/woodgates/feathers-nuxt

The configuration is at "server/config" folder. As set in the documents I have my app.js with

```js
const path = require('path');
process.env['NODE_CONFIG_DIR'] = path.join(__dirname, 'config/');
```

When I create the first service with "feathers generate services" all works good except for the database configuration which is stored in the root/config folder instead of root/server/config.

So when I run npm run dev I got 
TypeError: Path must be a string. Received undefined

### Expected behavior
The database configuration should be stored on the root/server/config/default.json file.

### Actual behavior
The database configuration is stored on the root/config/default.json file.

### System configuration

**Module versions** (especially the part that's not working):
part of package.json:
  "directories": {
    "lib": "server",
    "test": "test/"
  },
  "dependencies": {
    "@feathersjs/configuration": "^2.0.2",
    "@feathersjs/express": "^1.2.5",
    "@feathersjs/feathers": "^3.2.1",
    "feathers-nedb": "^3.1.0",,
    "nuxt": "^1.4.2",

**NodeJS version**:
8.11.4
**Operating System**:
I working with Linux Mint 19.

**Browser Version**:
Chromium
**React Native Version**:

**Module Loader**:
npm

daffl:
It is important that this happens _before_ the `const configuration = require('@feathersjs/configuration')` call.

woodgates:
@daffl 
Its before. This is part of the app.js
```
'use strict'
const path = require('path');
process.env['NODE_CONFIG_DIR'] = path.join(__dirname, 'config/');

const compress = require('compression');
const helmet = require('helmet');
const cors = require('cors');
const logger = require('./logger');


const feathers = require('@feathersjs/feathers')
const configuration = require('@feathersjs/configuration');
```

woodgates:
@daffl 
I dont want my configuration in two places. I want the configuration to be at "server/config" folder but the feathers cli is ignoring this.

daffl:
I don't understand what the problem is then. If you set

```js
const path = require('path');
process.env['NODE_CONFIG_DIR'] = path.join(__dirname, 'config/');

const configuration = require('@feathersjs/configuration')l
```

[...]

PR #1075, opened 2018-10-27
Config path from package.json

jraut:
Hello,

This pull request is created because of #987 

It lets the CLI use a custom config folder.

Based on the discussions on the issue, the following changes were made to generating a new project
- When creating a project, the package.json is created with a value for a config directory
- The value is read from `NODE_CONFIG_DIR` env variable since it is used by default
- If the value is not set, it is set to `config/` (the current hardcoded value)

Generator uses this new customizable value. It falls back to `config/` on run time to maintain backwards compatibility.

Thanks for a great framework and happy Hactober!

### DIFF
--- a/packages/generator-feathers/generators/app/configs/package.json.js
+++ b/packages/generator-feathers/generators/app/configs/package.json.js
@@ -2,6 +2,8 @@ const semver = require('semver');
 
 module.exports = function(generator) {
   const major = semver.major(process.version);
+  const envConfigDir = process.env['NODE_CONFIG_DIR'];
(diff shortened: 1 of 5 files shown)


jraut:
Closing the pull request as the CLI and generators have been to their own repository.

daffl:
Hm, sorry about that, I've been meaning to backport this into https://github.com/feathersjs/generator-feathers next week.

jraut:
All right! 

Let me open this in generator-feathers if the proposed functionality could provide useful. There were issues with running tests so I was a bit unconfident about the pull request.

Issue #844, opened 2018-03-27
Deprecate Passport JS?


samburgers:
Hi core devs, 

Stumbled across this node authentication wrapper, and wondered if it's worth looking at moving away from Passport JS now that Passport appears to be largely unmaintained these days? Probably more applicable to Feathers Docs + the CLI, but thought I would raise attention here! Cheers

https://github.com/simov/grant

daffl:
This is actually good timing because I started looking into more framework independent ways for doing authentication. I'd be ok with only supporting oAuth if you are using Feathers with an Express, Koa or Hapi transport.

Unfortunately this might mean pretty significant breaking changes which is risky given that authentication seems to be the most challenging part of Feathers to people.

kokujin:
This lib could help:
[permit](https://github.com/ianstormtaylor/permit)

bisubus:
Passport's big selling point is its ecosystem (also true for Express in general because Passport is an important part of its ecosystem). Grant is most viable alternative to Passport, yet it's not nearly as developed as Passport and lacks features (bye, passport-local).
I suppose that investment into framework (agnostic) adapters for Passport is messy but only way to not break things. There were some initiatives but all of them are abandoned, like https://github.com/hapijs/travelogue

daffl:
The reason is probably that it is not worth the effort. The problem with PassportJS is that

1) By now Passport core (2 releases in the past three years) and most plugins seem to be largely unmaintained
2) Less activity can also mean it is more stable but judging by most issues staying open completely unacknowledged I'm assuming 1)
3) It does not work well with other frameworks
4) It does work even less well with Websockets
5) Very likely even less so with HTTP2 or completely non-HTTP transport protocols (which Feathers can support)

And ultimately especially for Feathers which tries to enforce a protocol independent architecture it doesn't really do much. Right now we are basically converting

`Feathers protocol independent format` -> `Passport friendly (HTTP like) format` -> `Passport strategy` -> `Custom code in passport strategy` -> `Strategy return value` -> `Feathers format`

Where what we really just want do do is

`Feathers protocol independent format` -> `Custom verification code` -> `Feathers format`

All something like `passport-local` really does is compare a username and password (which you have to do in your own handler anyway) so why not just stick it in a hook?

[...]

PR #1112, opened 2018-11-27
Added path and method in to express request for passport

fazilvp:
### Summary

- [ ] This patch is for helping custom authentication strategy based on path and method. Presently we are not able to do it because these attributes are missing in request object. As per express documentation, an express request is supposed to have these fields.

   Method: https://expressjs.com/en/api.html#req.method
   Path: https://expressjs.com/en/api.html#req.path

   But this fields are not included while creating an express like request object for passport in 
   @feathersjs/authentication hook.authenticate() function.

- [ ] Currently there are no open issues for this.
- [ ] This PR has no dependency to other PRs in other repos.

### DIFF
--- a/packages/authentication/lib/hooks/authenticate.js
+++ b/packages/authentication/lib/hooks/authenticate.js
@@ -47,6 +47,8 @@ module.exports = function authenticate (_strategies, options = {}) {
       query: hook.data,
       body: hook.data,
       params: hook.params,
+      path: hook.path,
+      method: hook.method,
       headers: hook.params.headers || {},
       cookies: hook.params.cookies || {},
       session: {}
--- a/packages/authentication/lib/hooks/authenticate.js
+++ b/packages/authentication/lib/hooks/authenticate.js
@@ -43,12 +43,21 @@ module.exports = function authenticate (_strategies, options = {}) {
 
     // NOTE (EK): Passport expects an express/connect
     // like request object. So we need to create one.
+    const methodMap = {
+      find: 'GET',
+      get: 'GET',
+      create: 'POST',
+      update: 'PUT',
+      patch: 'PATCH',
+      remove: 'DELETE'
+    };
+
     let request = {
       query: hook.data,
       body: hook.data,
       params: hook.params,
       path: hook.path,
-      method: hook.method,
+      method: methodMap[hook.method] || hook.method,
       headers: hook.params.headers || {},
       cookies: hook.params.cookies || {},
       session: {}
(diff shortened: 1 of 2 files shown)


shankarregmi:
+1

daffl:
Thank you for the PR and fixing the test. I believe the `method` will be the Feathers service method name though. Not sure if it needs to be normalized into something that Passport understands. The mapping would be:

```js
const methodMap = {
  find: 'GET',
  get: 'GET',
  create: 'POST',
  update: 'PUT',
  patch: 'PATCH',
  remove: 'DELETE'
};

const httpMethod = methodMap[hook.method] || hook.method;
```

fazilvp:
@daffl You are right. I have added the normalization. Will the same be applicable to path also? Because normally express path is starting with a front slash and feathers path is not. Please share your thoughts.

subodhpareek18:
I don't think baking method in is a good idea. At the least there should be the original feathers method somewhere.

`find` and `get` calls mean different things. And in my custom auth module I have different permissions for `service:find` and `service:get`.

And as far as path pased or method based strategies are concerned, can't they be done with hooks?

```js
export const authorize = () =>
  iff(
    every(ctx => !ctx.path.startsWith('auth'), isProvider('external')),
    async ctx => {
      let userPermissions = R.path(
        ['params', 'user', 'role', 'permissions'],
        ctx
      );

      if (!userPermissions) {
        throw new errors.Forbidden('This user has no permissions');
      }

      if (userPermissions) {
        const hasPermissions = R.any(
          p => p === `${ctx.path}:${ctx.method}`,
          userPermissions
        );
        if (!hasPermissions) {
          throw new errors.Forbidden('Not allowed to execute this action.');
        }
      }

      return ctx;
    }
  );
```

daffl:
@subodhpareek18 This change is for the `authenticate` hook which passes a _fake_ request to the Passport strategy you indicated (usually `jwt` or `local`). This fake request object should be as close as possible to a normal request that a Passport strategy expects otherwise some may not work. Although I'm open to deprecating Passport (see discussion in #844) until then it makes sense to fake it as good as possible.

@fazilvp I think either way works, it depends what use-case prompted this PR.

fazilvp:
@daffl Path without slash is perfectly OK for our use-case. Let me know if you think any other changes are needed in this PR.

subodhpareek18:
@daffl that makes sense. Though I would still suggest the original feathersMethod should be passed without any normalization. But I guess that can always be done through middlewares for now. I'm having to do a similar thing for origin based auth.


The code below is for an old app on feathers@2.x.x though
```js
export const rest = () => (req, res, next) => {
  req.feathers = {
    ...req.feathers,
    headers: req.headers,
    origin: req.get('origin')
  };
  next();
};

export const socketio = () => (socket, next) => {
  socket.feathers = {
    ...socket.feathers,
    headers: socket.handshake.headers,
    origin: socket.handshake.headers.origin
  };
  next();
};
```

PR #1124, opened 2018-12-06
Make custom query for authentication

saiichihashimoto:
The current OAuth implementations will update an existing user if it finds one. Great! However, it will only find an existing user if they've authenticated with that specific OAuth provider. But we can give the user the ability determine what an existing user means.

For example:
```js
function makeQuery(profile, options) {
  return {
    $or: [
      { [options.idField]: profile.id },
      { email: profile.email },
    ],
  };
}
```

Now, it would also find users that have the same email, meaning it would link together accounts! Very cool! 👍


saiichihashimoto:
@daffl, I'm not sure what the process is for things like this, so let me know what I need to do. 👍

daffl:
Thank you for the pull request! I think it makes sense to add this but we do need tests and updated documentation otherwise nobody will know about it and we don't know if it works 😄

saiichihashimoto:
@daffl sounds good, I'll get to that later today! Is the function signature of `makeQuery` OK?

saiichihashimoto:
@daffl, I wrote some basic tests to see if the query is being called. The documentation is a separate repo so that PR will be later.

daffl:
Great, thank you! Releases as `@feathersjs/authentication-oauth1@1.1.0` and `@feathersjs/authentication-oauth2@1.3.0`

PR #1131, opened 2018-12-16
fix: `filterQuery` throw error when query parameter is unknown

daffl:
In order to avoid issues with the database adapter updates, unknown query parameters starting with `$` should throw an error instead of just being removed. If they were just being removed, limiting queries would become more generic, causing security regressions.

### DIFF
--- a/packages/adapter-commons/lib/filter-query.js
+++ b/packages/adapter-commons/lib/filter-query.js
@@ -1,4 +1,5 @@
 const { _ } = require('@feathersjs/commons');
+const { BadRequest } = require('@feathersjs/errors');
 
 function parse (number) {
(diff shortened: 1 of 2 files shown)


kkkrist:
Is there a way to somehow whitelist query parameters? For example, I've been using [nedb's $not operator](https://github.com/louischatriot/nedb#logical-operators-or-and-not-where) which isn't possible anymore as of this PR.

daffl:
The service options now [include a whitelist parameter](https://github.com/feathersjs-ecosystem/feathers-nedb/#serviceoptions). Also see the [database adapter migration guide](https://crow.docs.feathersjs.com/migrating.html#database-adapters).

PR #1153, opened 2019-01-02
fix: Do not inherit app object from Object prototype

daffl:
It appears that some browsers (or plugins or something) override the Object prototype, specifically the `init` property, which we use to call to initialize the application (causing `.init is not a function` errors). The best way to avoid this is to inherit from `null` instead of the Object prototype.

### DIFF
--- a/packages/feathers/lib/index.js
+++ b/packages/feathers/lib/index.js
@@ -3,9 +3,12 @@ const Proto = require('uberproto');
 const Application = require('./application');
 const version = require('./version');
 const { ACTIVATE_HOOKS, activateHooks } = require('./hooks');
+// A base object Prototype that does not inherit from a
+// potentially polluted Object prototype
+const baseObject = Object.create(null);
 
 function createApplication () {
-  const app = Object.create(null);
+  const app = Object.create(baseObject);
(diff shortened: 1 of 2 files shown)


packages/feathers/lib/index.js
```diff
@@ -3,9 +3,12 @@ const Proto = require('uberproto');
 const Application = require('./application');
 const version = require('./version');
 const { ACTIVATE_HOOKS, activateHooks } = require('./hooks');
+// A base object Prototype that does not inherit from a
+// potentially polluted Object prototype
+const baseObject = Object.create(null);
 
 function createApplication () {
-  const app = {};
+  const app = Object.create(baseObject);
```
bertho-zero: That would not be enough to do that?

```js
const app = Object.create(null);
```

daffl: I thought so, too but [Uberproto](https://github.com/daffl/uberproto) tries to [get the Prototype](https://github.com/daffl/uberproto/blob/master/lib/proto.js#L74) when mixing into the `app` object, gets `null` and blows up so I needed to add another one to the chain.

PR #1163, opened 2019-01-09
[authentication] Consider data.payload in authentication.create

HenriBeck:
### Summary

The docs state that "Anything included in the data.payload object will be in the JWT's payload. If you include a payload object in params, it's properties will take precedence over data" which was not the case. It only considered payload inside params and not data

https://docs.feathersjs.com/api/authentication/server.html#servicecreatedata

No

No


daffl:
Thank you for the pull request but I think in this case the documentation is wrong. `data.payload` can not be used because that would make it possible to create valid JWTs for another user (if you know their id).

HenriBeck:
Yeah makes sense.

PR #1173, opened 2019-01-18
fix: added '$and' operator in filter-query

jaiyashree:
### Summary

Added '$and' operator in filter-query operators.

     - According to the manual, https://docs.mongodb.com/manual/reference/operator/query/and/   '$and' 
       is a permitted operator, but not listed in the `@feathersjs/adapter-commons/lib/filter-query.js`
       `const OPERATORS = ['$in', '$nin', '$lt', '$lte', '$gt', '$gte', '$ne', '$or'];`
- [ ] Are there any open issues that are related to this? - NO
- [ ] Is this PR dependent on PRs in other repos? - NO

### Other Information

Say, i wanted to check for a condition like columnA value must not-be '1.5' but the value must be greater than 0 and the columnA must be less than 5, then i would like to form a `and condition` which i dynamically build using `jQuery builder` like below:

`service.find( { $and: [ { columnA: { $ne: 1.5 } }, { columnA: { $gt: 0 } }, { columnA: { $lt: 5 } } ] } )`

This is currently not possible as '$and' throws 'Invalid query parameter '$and' '

### DIFF
--- a/packages/adapter-commons/lib/filter-query.js
+++ b/packages/adapter-commons/lib/filter-query.js
@@ -89,7 +89,7 @@ const FILTERS = {
   $select: (value) => value
 };
 
-const OPERATORS = ['$in', '$nin', '$lt', '$lte', '$gt', '$gte', '$ne', '$or'];
+const OPERATORS = ['$in', '$nin', '$lt', '$lte', '$gt', '$gte', '$ne', '$or', '$and'];
 
 // Converts Feathers special query parameters and pagination settings
 // and returns them separately a `filters` and the rest of the query


daffl:
Thank you for the pull request. As just discussed in https://github.com/feathersjs/docs/pull/1263, `$and` is not part of Feathers default query syntax at the moment but if all adapters support it and a proper tests has been added to `@feathersjs/adapter-tests` it might be worth to consider adding it.

cannc4:
We are also experiencing the same issue and would be easier if this PR was accepted. If not, do you mind clarifying the workaround for this?

cannc4:
I wasn't able to `whitelist` the parameters by setting the value in the service options
```
 const options = {
    name: 'tag-cloud',
    whitelist: ['$all', '$and', '$exists', '$text', '$search'],
    paginate: {
      default: 250,
      max: 100000
    }
  }
```
but it works as expected when added to the operator array in the script. 
Any thoughts?

daffl:
I'm going to close this since it has been open for a while and now has some conflicts. `$and` might be added to the Feathers query syntax in the future but needs a thorough review on the implications of it and how it would be implemented in non-MongoDB related databases. If the database supports this syntax already it can just be added to the [database adapter whitelist option](https://docs.feathersjs.com/api/databases/common.html#service-options)

Issue #1194, opened 2019-02-05
user hooks doesn't recieve any params.  


navneethkrish:
### Steps to reproduce
Same issue in authenticate-jwt like Ref: https://github.com/feathersjs/authentication-local/issues/14#issue-221814991

### Expected behavior
While authenticating using authenticate-jwt, the before hook of get method on the user service should be provided with feathers params specified (except root level params) in request.

### Actual behavior
The `get` hook on the user service does not receive any additional data other than user id because of verifier passes only user id.

Ref:https://github.com/feathersjs/feathers/blob/2a6dc7019926fb92840045e904c7b9ac61c4c9ea/packages/authentication-jwt/lib/verifier.js#L29.

whether it could be fixed by the same way or Is there any other alternative?

If need, we will fix and raise the PR?

changes: 
const params = omit(req.params || {}, 'query', 'provider', 'headers', 'session', 'cookies');
...
....
this.service.get(id, params).then(entity => {
      const newPayload = { [`${this.options.entity}Id`]: id };
      return done(null, entity, newPayload);
    })

### System configuration 
**Module versions** (especially the part that's not working):

* `feathers-authentication-jwt` - 2.0.9

Issue #1045, opened 2018-10-06
Authentication next


daffl:
I wrote about this a little already in the [Feathers Crow roadmap update](https://blog.feathersjs.com/in-the-crows-nest-feathersjs-roadmap-update-41a1eafc6f5). This issue is meant to collect all related issues that the next version of Feathers authentication will cover.

## Framework independent

With the current (Buzzard) release, Feathers core became fully framework independent, however, the authentication plugin is still registering some Express middleware. In the next version that middleware will move into @feathersjs/express and make authentication core also framework independent. This will allow new transports like Koa or MQTT.

- https://github.com/feathersjs/authentication/issues/384
- https://github.com/feathersjs/authentication/issues/648
- https://github.com/feathersjs/feathers/issues/878

## Improved socket authentication

Currently, websockets are authenticated and logged out through a custom event mechanism. This was causing some issues (e.g. infrequent “No auth token” errors), especially around reliable reconnection on both sides, the server and the client.

- https://github.com/feathersjs/authentication/issues/392
- https://github.com/feathersjs/authentication/issues/517
- https://github.com/feathersjs/authentication/issues/545
- https://github.com/feathersjs/authentication/issues/577
- https://github.com/feathersjs/authentication/issues/586
- https://github.com/feathersjs/authentication/issues/682
- https://github.com/feathersjs/authentication/issues/698
- https://github.com/feathersjs/authentication/issues/699
- https://github.com/feathersjs/feathers/issues/892
- https://github.com/feathersjs/feathers/issues/941
- https://github.com/feathersjs/feathers/issues/991

## Improved authentication client

This will remove dependence on the stateful access token and gracefully handle socket re-authentication.

- https://github.com/feathersjs/authentication-client/issues/64
- https://github.com/feathersjs/authentication-client/issues/67
- https://github.com/feathersjs/authentication-client/pull/99
- https://github.com/feathersjs/feathers/issues/854

## No more cookies, improved oAuth

Feathers is a library for creating APIs using JWT for authentication. The one instance where a normal Feathers setup uses cookies is to pass the JWT to the browser after an oAuth (Facebook, Google etc.) authentication flow. The new oAuth and authentication client will no longer use cookies and split the oAuth flow into two parts instead of trying to do everything at once. The oAuth access token will be set in a cross-domain friendly (only locally accessible) URL hash which can then be exchanged for a JWT using Feathers standard authentication mechanism.

- https://github.com/feathersjs/authentication/issues/120
[...]

PR #1195, opened 2019-02-05
fix(authentication): passes params to user hooks from verifier of aut…

navneethkrish:
PR for this issue : https://github.com/feathersjs/feathers/issues/1194#issue-406702850

### DIFF
--- a/packages/authentication-jwt/lib/verifier.js
+++ b/packages/authentication-jwt/lib/verifier.js
@@ -1,5 +1,6 @@
 const Debug = require('debug');
 const debug = Debug('@feathersjs/authentication-jwt:verify');
+const { omit } = require('lodash');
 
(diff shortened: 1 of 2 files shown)


daffl:
Thank you for the pull request @navneethkrish. A lot of this will be changing for Authentication v3 (https://github.com/feathersjs/feathers/issues/1045) which is [currently under development](https://github.com/feathersjs/feathers/tree/authentication/v3) and will be available as pre-releases later this month.

I'd recommend to make a copy of the verifier with the changes you are suggesting and [supply it as a custom verifier](https://docs.feathersjs.com/api/authentication/jwt.html#verifier) in you project.

daffl:
This is a good point though by the way. It will definitely do this going forward.

Issue #776, opened 2018-01-08
Conditionnally configure JWT expireIn?


D219:
I have a use case where some users will need a JWT with a longer timeout. I know that in a standard app, I can edit src/authentication.js to add a `expireIn` flag, but this result in all JWT having the extended validation time.

Is there a way, for example in a hook, to change the expireIn option for just that request? By manually searching inside it, the only thing I found is `hook.app.authentication.jwt.expireIn`, which doesn't look like what I want.

daffl:
You should be able to set JWT options dynamically in `params.jwt` in a `before` `create`hook:

```js
app.service('authentication').hooks({
  before: {
    create(context) {
      context.params.jwt.expiresIn = '2d';
    }
  }
});
```

D219:
It did work, thanks a lot :)

PR #1227, opened 2019-02-23
Allow for removing JWT fields via json configuration.

AndrewJDR:
jsonwebtoken actually supports not setting the following fields:
expiresIn
issuer
subject
audience
iat

And perhaps others.
Taking expiresIn as an example, if this is undefined, it means the JWT should never expire and will not have an exp field:
Furthermore, the option must be completely undefined to achieve this behavior, explicitly setting it to undefined, null, or 0 is not sufficient:
https://github.com/auth0/node-jsonwebtoken/issues/417
This means there needs to be a way of explicitly removing these keys from the jwt options object.

feathers/authentication is always providing default values for these fields,
and didn't seem to any working facility for overwriting them programatically at runtime.
A previously mentioned technique of overriding jwt parameters via context.params.jwt.* (https://github.com/feathersjs/feathers/issues/776)
no longer works because context.params.jwt is not available in the auth/before hook. I don't know why this is the case, and I would have to investigate the git history extensively to find out.

In any case, I thought it would be nice to have a declarative way (rather than programmatic) to remove various jwt options/fields by setting a configuration option, so this commit adds a "jwtRemoveOptions" to the authentication section of the feathers configuration. This should be set to an array of key names to remove from the jwt options.

In addition, a conditional was placed around the setTimeout: if expiresIn is undefined it doesn't set the timeout.

### DIFF
--- a/packages/authentication/lib/options.js
+++ b/packages/authentication/lib/options.js
@@ -13,6 +13,7 @@ const defaults = {
     httpOnly: false,
     secure: true
   },
+  jwtRemoveOptions: [],
   jwt: {
     header: { typ: 'access' }, // by default is an access token but can be any type
     audience: 'https://yourdomain.com', // The resource server where the token is processed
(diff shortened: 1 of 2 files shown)


daffl:
Thank you for the pull request @AndrewJDR. I included a change that makes this possible in the new authentication version which is currently being finalized.

PR #1252, opened 2019-03-19
fix: update readme

bastibense:
add copyright year 2019

### Summary

Updated copyright.

No

No


daffl:
Thanks! If we have two years already maybe we can include the full time range with 2014 - 2019?

PR #1266, opened 2019-04-01
ENV variables support updated

aliazlan4:
This code was previously checking if a value of env variable exists in the process.env but it should check if the key is present in the process.env and if the same key is also present in the env that its value should be used instead of that key's value present in the default.json or production.json files etc.

### Summary

here](https://github.com/feathersjs/feathers/blob/master/.github/contributing.md#pull-requests))

### Other Information

If there's anything else that's important and relevant to your pull
request, mention that information here. This could include
benchmarks, or other information.

Your PR will be reviewed by a core team member and they will work with you to get your changes merged in a timely manner. If merged your PR will automatically be added to the changelog in the next release.

If your changes involve documentation updates please mention that and link the appropriate PR in [feathers-docs](https://github.com/feathersjs/feathers-docs).

### DIFF
--- a/packages/configuration/lib/index.js
+++ b/packages/configuration/lib/index.js
@@ -23,8 +23,8 @@ function init () {
           if (value.indexOf('\\') === 0) {
             value = value.replace('\\', '');
           } else {
-            if (process.env[value]) {
-              value = process.env[value];
+            if (process.env[name]) {
+              value = process.env[name];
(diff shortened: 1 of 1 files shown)


daffl:
Thank you for the pull request @aliazlan4 but this is how it is currently intended. This is documented in the [Variable types](https://docs.feathersjs.com/api/configuration.html#variable-types) documentation:

> If the *value* is a valid environment variable (e.v. `NODE_ENV`), use its value instead

PR #1288, opened 2019-04-15
Added type definitions for hook-less service methods.

deskoh:
Related to https://github.com/feathersjs/feathers/pull/1269 to support [Calling the hook-less service methods](https://crow.docs.feathersjs.com/migrating.html#hook-less-service-methods) of the database adapters.

Hook-less service methods is also useful for in unit tests for creating / updating multiple items if `multi` flag is not set as the hook-less methods does not check for `multi` flag,


daffl:
Thank you! I think it makes sense to add them. Can we put them into a separate interface though since it's only implemented by the service adapters?

jordanbtucker:
I think this needs to go into adapter-commons instead. For example, the following code won't work because it just uses an object based service.

```js

const app = feathers()

const items = []

app.use('items', {
	async create (data) {
		data.id = items.length
		items.push(data)
		return data
	},
})

const itemsService = app.service('items')

itemsService.hooks({
	before: {
		create: [
			() => { console.log('Before hook') },
		],
	},
})

async function run () {
	await itemsService.create({text: 'Hello, World!'})
	await itemsService._create({text: 'No hooks'})
}

run()
```

If we put it into adapter-commons, then we might be able to use namespace merging, type intersection, module augmentation or some combination to ensure the type definition works properly.

deskoh:
@jordanbtucker Existing type definitions have similar problems for object-based service. For example, referencing the code above, the snippet below also exhibits similar problems:

```JavaScript
async function run () {
	await itemsService.create({text: 'Hello, World!'})
	await await itemsService.get(1);
	// throws TypeError: itemsService.get is not a function even though there is no type error using current type definitions.
}
```

The type resolution need to be able to distinguish object-based service and service using the Database [Common API](https://docs.feathersjs.com/api/databases/common.html). Not sure how to achieve this at this point.

PR #1307, opened 2019-04-24
fix: Handle error oAuth redirect in authentication client

daffl:


packages/authentication-client/test/index.test.ts
```diff
@@ -41,18 +41,18 @@ describe('@feathersjs/authentication-client', () => {
     assert.strictEqual(typeof app.logout, 'function');
   });
 
-  it('setJwt, getJwt, removeJwt', async () => {
+  it('setAccessToken, getAccessToken, removeJwt', async () => {
```
bertho-zero: `it('setAccessToken, getAccessToken, removeAccessToken', async () => {`

PR #1319, opened 2019-05-02
Fix authentication client logout function to guard against null

### LINKED ISSUE #1318: [Feathers 4.0.0-pre] Clientside app.logout() is not working
It throws:

```
Uncaught (in promise) TypeError: Cannot read property 'then' of null
    at AuthenticationClient.logout (core.js:100)
```

image

KidkArolis:
Fixes #1318


daffl:
I think this fix makes sense I'm just wondering why this was happening because `this.app.get('authentication')` should only be `undefined` if you are not logged in.

KidkArolis:
Ah, I see, thats right. Maybe it was related to some other auth issues i was experiencing. I will double check.

KidkArolis:
Yeah, when I call logout, the authentication is indeed defined. The reason it was undefined previously is possibly related to the other issues where I would succesfully authenticate, but then jwt strategy would not work for me logging me out. Hm.. or maybe I was just calling logout while being logged out expecting that not to throw.

In any case, can't hurt to guard, I guess. Good to know it's all working fine.

Issue #1324, opened 2019-05-03
[Feathers 4.0.0-pre] Authentication issues


KidkArolis:
I'm creating a little umbrella issue to keep track of the issues I've found with the auth in the 4.0.0. I will update this as things get addressed one way or another.

### 1) [FIXED] Backwards incompatible AUTHENTICATE symbol ✅

Previously, it was possible to pass `authenticated: true` param, for example in tests, to skip authentication hook and pass the user manually. in pre.0, this has been replaced with `AUTHENTICATE: false` sumbol. The change has been reverted and is now merged.

PR: https://github.com/feathersjs/feathers/pull/1317

### 2) [FIXED] Clientside app.logout() doesn't work ✅

Clientside `app.logout()` throws.

PR: https://github.com/feathersjs/feathers/pull/1319
Update: was only an issue if calling logout if already logged out.

### 3) [FIXED] Strategy errors get swallowed ✅

Each auth strategy is always being tried in sequence until a succesful match is found or they all fail. The issue with this is that the request might have been for strategy `jwt` but the first strategy in the list was `local`. In this case, we won't respond to the client with the real auth error. The issue is discussed in more depth here and a proposed fix has been suggested:

Discussion: https://github.com/feathersjs/feathers/issues/1323
PR: https://github.com/feathersjs/feathers/pull/1327

### 4) Local strategy fails to resolve user object in case hooks rely on user ❌

The new local strategy fetches user twice. Once as an internal call to retrieve the full user object and compare the password. And once more to get a "clean" user object to pass back to the clientside. When resolving the second time, the authenticated user object is not being passed to the find call, which means the call could fail if any hooks rely on the authenticated user object, e.g. to scope the call to an organisation with restrictToOwner.

PR: https://github.com/feathersjs/feathers/pull/1320

### 5) Local strategy makes new assumptions about user `find()` ❌

Previously, the /authenticate endpoint would only return the jwt access token. Now it also returns the resolved user, which is usually what you want. There might be one subtle issue with this approach in that this strategy now makes an assumption about whether the `find()` by username will work for an external call. Note: this is particularly dangerous if the user service strips out query params, e.g. if "email" or "username" gets stripped out, the local strategy selects all users (with limit: 1), essentially returning the wrong user (first match).

Suggestion: use `get(id)` to resolve the user object for external use.

### 6) JWT strategy fails because it forwards all params from authenticate hook ❌

[...]

PR #1333, opened 2019-05-07
fix: Improve authentication parameter handling

daffl:
For #1324 4), 5), 6) and 7) and should close #1320

### DIFF
--- a/packages/authentication/src/jwt.ts
+++ b/packages/authentication/src/jwt.ts
@@ -1,4 +1,5 @@
 import { NotAuthenticated } from '@feathersjs/errors';
+import { omit } from 'lodash';
 import { AuthenticationRequest } from './core';
 import { Params } from '@feathersjs/feathers';
 import { IncomingMessage } from 'http';
@@ -36,13 +37,20 @@ export class JWTStrategy extends AuthenticationBaseStrategy {
    * @param params Service call parameters
    */
   async getEntity (id: string, params: Params) {
+    const { entity } = this.configuration;
     const entityService = this.entityService;
 
     if (entityService === null) {
       throw new NotAuthenticated(`Could not find entity service`);
     }
 
-    return entityService.get(id, params);
+    const result = await entityService.get(id, omit(params, 'provider'));
+
+    if (!params.provider) {
+      return result;
+    }
+
+    return entityService.get(id, { ...params, [entity]: result });
   }
 
   async authenticate (authentication: AuthenticationRequest, params: Params) {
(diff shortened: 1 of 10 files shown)


packages/authentication/src/jwt.ts
```diff
@@ -42,7 +42,6 @@ export class JWTStrategy extends AuthenticationBaseStrategy {
       throw new NotAuthenticated(`Could not find entity service`);
     }
 
-    // @ts-ignore
     return entityService.get(id, params);
```
KidkArolis: Correct me if I'm wrong, but I think there's another edge case hiding here.

Say the client calls `app.authenticate({ strategy: 'jwt', accessToken })` - which, for example, my app does on initial render (I think that's mandatory to authenticate the socket?). That calls `strategy.authenticate()`, which calls `this.getEntity(entityId, params)` passing all of the params through. But entity can't be resolved from an external call if it's unauthenticated – `authenticate()` hook is skipped, but `restrictToOwner` or any custom logic looking at the user object is not skipped. This is why I had to resort to differentiating between internal and external calls in my "patched" jwt strategy:

```js
class CustomJWTStrategy extends JWTStrategy {
  async getEntity(id, params) {
    const { entity } = this.configuration
    const entityService = this.entityService

    if (entityService === null) {
      throw new errors.NotAuthenticated(`Could not find entity service`)
    }

    const result = await entityService.get(id, omit(params, ['provider', 'query']))

    if (params.provider) {
      return entityService.get(id, { ...omit(params, 'query'), [entity]: result })
    } else {
      return result
    }
  }
}
```

daffl: Yup, that makes sense, will add it to this PR.

daffl: Allright, it should be updated to work the same for both, the local and JWT strategies.

KidkArolis: Amazing. This all looks good to go to me.

emg110: Thanks daffi but not only authentication service suffers from this (NotFound Error) on service entities (while app.services is returning all of them intact and in place!!) still exists on feathers via socketio (express is fine) I have had to handle it via socket common hook  (like it's authorization!!) which is not the reason i rushed toward feathers in the first place , which was integration.

daffl: I'm not exactly sure what you mean. A separate issue with a small example to reproduce would be helpful.

emg110: OK, sure . just wanted to say thanks , great job and the rest I'll issue.

PR #1335, opened 2019-05-08
fix: Improve oAuth option handling

daffl:
Pull request that should close https://github.com/feathersjs/feathers/issues/1324 specifically

- `9)` by adding an `expressSession` option to provide a custom session store
- `10)` by using both `profile.sub` and `profile.id`
- `11)` by adding `getRedirect` to the strategy (so it is also customizable on a per-strategy basis)
- `12)` by supporting the `getEntityQuery` and `getEntityData` methods that should make customization easier

@KidkArolis let me know if you see anything that's missing

### DIFF
--- a/packages/authentication-oauth/src/strategy.ts
+++ b/packages/authentication-oauth/src/strategy.ts
@@ -38,7 +38,7 @@ export class OAuthStrategy extends AuthenticationBaseStrategy {
 
   async getEntityData (profile: OAuthProfile, _existingEntity: any, _params: Params) {
     return {
-      [`${this.name}Id`]: profile.id
+      [`${this.name}Id`]: profile.sub || profile.id
     };
   }
 
--- a/packages/authentication-oauth/src/strategy.ts
+++ b/packages/authentication-oauth/src/strategy.ts
@@ -1,8 +1,9 @@
 // @ts-ignore
 import getProfile from 'grant-profile/lib/client';
+import querystring from 'querystring';
 import Debug from 'debug';
 import {
-  AuthenticationRequest, AuthenticationBaseStrategy
+  AuthenticationRequest, AuthenticationBaseStrategy, AuthenticationResult
 } from '@feathersjs/authentication';
 import { Params } from '@feathersjs/feathers';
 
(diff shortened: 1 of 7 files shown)


packages/authentication-oauth/src/utils.ts
```diff
@@ -1,38 +1,24 @@
...
 
 export interface OauthSetupSettings {
-  path?: string;
-  authService?: string;
-  linkStrategy?: string;
-  getRedirect? (service: AuthenticationService, data: AuthenticationResult|Error): Promise<string>;
+  path: string;
+  authService: string;
+  linkStrategy: string;
+  expressSession: RequestHandler;
```
KidkArolis: That doesn't look like the right type?

daffl: It's `import { RequestHandler } from 'express';` right? It at least doesn't complain when initializing the default options in https://github.com/feathersjs/feathers/pull/1335/files/40ebbddeae3995e840c059146a6e3d3597087107#diff-b3c31ade33c8365501f61c7361f117b2R17

packages/authentication-oauth/src/strategy.ts
```diff
@@ -29,6 +30,18 @@ export class OAuthStrategy extends AuthenticationBaseStrategy {
...
 
+  async getEntityQuery (profile: OAuthProfile, _params: Params) {
+    return {
+      [`${this.name}Id`]: profile.sub || profile.id
+    };
+  }
+
+  async getEntityData (profile: OAuthProfile, _existingEntity: any, _params: Params) {
+    return {
+      [`${this.name}Id`]: profile.id
```
KidkArolis: Is profile.sub missing here?

packages/authentication-oauth/src/utils.ts
KidkArolis: Ah, I guess this makes sense if expressSession is like a middleware (RequestHandler).

packages/authentication-oauth/src/strategy.ts
daffl: Good catch. Fixed!

Issue #1360, opened 2019-05-19
[Feathers 4.0.0-pre.2] Authenticated: true is not added to context.params


KidkArolis:
Is this by design or by accident that the `authenticate()` hook bundled in `@feathersjs/authentication` no longer sets `context.params.authenticated = true`? This breaks users of `makeCallingParams` that are relying on this param to be passed through to nested service calls. The old hook was setting it: https://github.com/feathersjs/authentication/blob/master/lib/hooks/authenticate.js#L80.

KidkArolis:
I had to create such a proxy hook in my app, otherwise it's inconvenient to forward calls to other services from already authenticated services (since they rerun authentication() hook logic instead of skipping it):

```js
const auth = () =>
  combine(authenticate('jwt'), context => {
    if (context.params.provider) {
      context.params.authenticated = true
    }
  })
```

PR #1367, opened 2019-05-24
fix: Set authenticated: true after successful authentication

KidkArolis:
### Summary

Fixes https://github.com/feathersjs/feathers/issues/1360


daffl:
Good call. I removed it because it seemed redundant but with #1317 bringing back the param it makes sense to set it.

KidkArolis:
One alternative thought..

When i first wrote some feathers tests, it seemed a little superfluous that i had to pass both { user, authenticated }.

And so i think it was interesting that you were trying to remove the need for that param, by hiding it more behind an internal symbol.

But what if presence of user object would be enough to treat requests as authenticated. So instead of using/checking authenticated we just have { user }.

PR #1377, opened 2019-05-29
fix: Make oAuth paths more consistent and improve authentication client

daffl:
This pull request makes the oAuth callback paths consistent. It uses the redirects  discussed in https://github.com/simov/grant/pull/116 to make things more straightforward and backwards compatible. Also changes the base path to `/oauth` as previously discussed with @KidkArolis (it can be made backwards compatible by setting `authentication.oauth.defaults.path` to `/auth`).

The relevant oAuth paths now are:

- `http(s)://<host>/oauth/<provider>` to initiate the oAuth flow
- `http(s)://<host>/oauth/<provider>/callback` as the callback path that should be set in the oAuth application settings
- `http(s)://<host>/oauth/<provider>/authenticate` as the internal redirect

### DIFF
--- a/packages/authentication-oauth/test/express.test.ts
+++ b/packages/authentication-oauth/test/express.test.ts
@@ -15,15 +15,15 @@ describe('@feathersjs/authentication-oauth/express', () => {
   after(() => server.close());
 
   it('auth/test', async () => {
-    axios.get('http://localhost:9876/auth/test?feathers_token=testing');
+    axios.get('http://localhost:9876/oauth/connect/test?feathers_token=testing');
   });
 
   it('auth/test with query', async () => {
-    axios.get('http://localhost:9876/auth/test?other=test');
+    axios.get('http://localhost:9876/oauth/connect/test?other=test');
   });
 
   it('auth/test/authenticate', async () => {
-    const { data } = await axios.get('http://localhost:9876/auth/test/authenticate?id=expressTest');
+    const { data } = await axios.get('http://localhost:9876/oauth/connect/test/authenticate?id=expressTest');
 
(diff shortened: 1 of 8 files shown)


KidkArolis:
Looks good. Are you gonna be adding backwards compatible redirects?

daffl:
I tried but couldn't get it to work. oAuth always errors when I try to redirect from a backwards compatible callback path. At the moment it looks like this is only possible if my PR in Grant gets accepted. I might submit it again to discuss more.

daffl:
@KidkArolis Allright, got some help and managed to make the paths backwards compatible (updated in the description).

packages/authentication-oauth/test/express.test.ts
```diff
@@ -14,35 +14,35 @@ describe('@feathersjs/authentication-oauth/express', () => {
...
       assert.ok(/Cannot GET/.test(error.response.data));
       delete app.get('authentication').oauth.redirect;
     }
   });
 
-  it('auth/test/authenticate with error', async () => {
+  it('oauth/test/authenticate with error', async () => {
     try {
-      await axios.get('http://localhost:9876/oauth/connect/test/authenticate');
+      await axios.get('http://localhost:9876/oauth//test/authenticate');
```
KidkArolis: Why is there a double slash `//` here?

daffl: Definitely a find and replace issue 😄

KidkArolis: Ah, I thought this was an assertion, which was a bit more worrying, but it's just a get. In any case..

PR #1396, opened 2019-06-12
fix: Improve transport-commons types

vonagam:
Normalise types of `PublishMixin`, `ServiceAddons` and `Application` for `publish`/`registerPublish`:
- an event is `string` or `ALL_EVENTS` symbol.
- `callback` is now `publisher` in argument and variable naming.
- a publisher returns a channel, an array of channels, void or a promise* for those.
- return value is `this`.

\* - before a promise was not present in types and not mentioned in documentation, but based on [this line](https://github.com/feathersjs/feathers/blob/28888a1a6bd61eac778573234e2d668488f9989e/packages/transport-commons/src/channels/index.ts#L81) it is possible currently to return a promise. We can remove this feature from types and remove `Promise.resolve` from that line if this should not be supported.

Plus small performance improving in publishing: no point in creating array, using `_.get` and then `.find` with `typeof` when simple `or` condition will do (ideally publisher should be cached).

### DIFF
--- a/packages/transport-commons/src/channels/index.ts
+++ b/packages/transport-commons/src/channels/index.ts
@@ -1,20 +1,20 @@
 import Debug from 'debug';
-import { get, compact, flattenDeep, noop } from 'lodash';
+import { compact, flattenDeep, noop } from 'lodash';
 import { Channel } from './channel/base';
 import { CombinedChannel } from './channel/combined';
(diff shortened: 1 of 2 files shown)


daffl:
Nice, thank you! Will go out with the next pre-release.

PR #1397, opened 2019-06-12
chore: Remove src folder from .npmignore for ts packages

vonagam:
Any drawbacks to this? 

As i understand should not affect local development as `link` command ignores `.npmignore`.

### DIFF
--- a/packages/authentication-client/.npmignore
+++ b/packages/authentication-client/.npmignore
@@ -1,3 +1,2 @@
 test/
-src/
 tsconfig.json
--- a/packages/authentication-local/.npmignore
+++ b/packages/authentication-local/.npmignore
@@ -1,3 +1,2 @@
 test/
-src/
 tsconfig.json
--- a/packages/authentication-oauth/.npmignore
+++ b/packages/authentication-oauth/.npmignore
@@ -1,3 +1,2 @@
 test/
-src/
 tsconfig.json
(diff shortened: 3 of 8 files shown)


daffl:
Thank you for the pull request. Wouldn't we still want the correct line numbers and source references on errors though?

vonagam:
Hm... Well, while you will know a line for an error in ts source file, while debugging you will not be able to see any code (nor source, nor build) and a place of a said error. So debugging is definitely harder...

Reproduce yourself:
1) Create and install a package with `"@feathersjs/transport-commons": "4.0.0-pre.3"`.
2) Run `node --inspect-brk`.
3) Run such erroneous code:
```js
const Channel = require('@feathersjs/transport-commons/lib/channels/channel/base').Channel;
const channel = new Channel();
channel.filter(33);
```

You will see that the error will have [`base.ts:55`](https://github.com/feathersjs/feathers/blob/28888a1a6bd61eac778573234e2d668488f9989e/packages/transport-commons/src/channels/channel/base.ts#L55) as a location reported, but you will not be able to see such file and without seeing you will not be able to place breakpoints or understand fully what's going on.

Unfortunately there is no way to selectively disable source maps in dev tools for `@feathersjs/*` dependencies and keep them working for your own code.

Do you have any solution for this?

Other way to go is to not ignore `src` folder. That way while debugging you will see ts source file and will be able to place breakpoints.

daffl:
I think publishing the `src/` folder as well makes the most sense then. I guess there will be some slight overhead during startup but it would definitely be worth having the actual code references (otherwise most error messages might not be very helpful).

vonagam:
Done.

Issue #932, opened 2018-08-16
Koa style middleware: Hooks for any asynchronous method


daffl:
Hooks are a powerful way to control the flow of asynchronous methods. Koa popularized a new kind of middleware that allows an even more flexible approach to control this flow. I propose to adapt this pattern as a new `async` hook type.

## Problem

Currently, [Feathers hooks](https://docs.feathersjs.com/api/hooks.html) are split into three types, `before`, `after` and `error`:

![feathers-hooks-old](https://user-images.githubusercontent.com/338316/44229860-34ca6300-a14e-11e8-8d78-0b0923721737.png)

While this pattern works quite well, it is a little cumbersome to implement functionality that needs to have access to the `before`, `after` and/or `error` flow at the same time. For example functionality for a profiler that logs the method runtime or extracting and storing data (e.g. associations) that will be used after the method call returns requires a `before` and an `after` hook.

## Proposal

With [async/await](https://javascript.info/async-await) now being fully supported by all recent versions of Node - and the ability to fall back to Promises for Node 6 - we now have a powerful new way to create middleware that can control the whole flow within a single function.

This pattern has been implemented already in [KoaJS](https://koajs.com/) for processing HTTP requests. However, similar to [what I wrote before about Express style middleware](https://blog.feathersjs.com/api-service-composition-with-hooks-47af13aa6c01), this also doesn't have to be limited to the HTTP request/response cycle but can apply to any asynchronous function, including Feathers services. Basically this new kind of hook will wrap _around_ the following hooks and the service method call like this:

![feathers-hooks-2 1](https://user-images.githubusercontent.com/338316/44229864-3b58da80-a14e-11e8-934d-642e2f6cf880.png)

### How it works

Additionally to the [hook context](https://docs.feathersjs.com/api/hooks.html#hook-context) `async` hooks also get an asynchronous `next` function that controls passing the flow to the next hook in the chain.

The following example registers an application wide before, after and error logger and profiler for all service method calls:

```js
app.hooks({
  async: [
    async (context, next) => {
      try {
        const start = new Date().getTime();

        console.log(`Before calling ${context.method} on ${context.path}`);

        await next();

        console.log(`Successfully called ${context.method} on ${context.path}`, context.result);
        console.log('Calling ${context.method} on ${context.path} took ${new Date().getTime() - start}ms');
        return context;
      } catch(error) {
        console.error(`Error in ${context.method} on ${context.path}`, error.message);
        // Not re-throwing would swallow the error
[...]

PR #1407, opened 2019-06-18
fix: Clean up hooks code

vonagam:
Fixes two errors:

1) `finally` hooks can possibly run twice for same method call if one of them throws. Also a `finally` hook if throws activates `error` hooks, this is in conflict with this [comment](https://github.com/feathersjs/feathers/blob/f9d8536c400caba355ccc8a2cdf171dcd9cff555/packages/feathers/lib/events.js#L77) and also `Promise.finally` and `try/catch/finally` convention. Because currently it is like this ([1](https://github.com/feathersjs/feathers/blob/f9d8536c400caba355ccc8a2cdf171dcd9cff555/packages/feathers/lib/hooks/index.js#L73), [2](https://github.com/feathersjs/feathers/blob/f9d8536c400caba355ccc8a2cdf171dcd9cff555/packages/feathers/lib/hooks/index.js#L85)):
```js
try {
  run([...before])();
  method();
  run([...after, ...finally])();
} catch {
  run([...error, ...finally])();
}
```

2) An `error` hook can get a context which is not up to date, if any `before` hook returns new context object and a service method throws, because errors ([1](https://github.com/feathersjs/feathers/blob/f9d8536c400caba355ccc8a2cdf171dcd9cff555/packages/feathers/lib/hooks/index.js#L59), [2](https://github.com/feathersjs/feathers/blob/f9d8536c400caba355ccc8a2cdf171dcd9cff555/packages/feathers/lib/hooks/index.js#L62)) thrown in a method invocation do no have `hook` property set (which should contain a last valid hook context), that's why `error` hooks end up using an initial hook object.

Added fixes and tests for such cases. Also small clean up changes for readability and performance.

A question for future: What are use cases for allowing returning new context object from hook functions where modifying context in place is not as convenient?

### DIFF
--- a/packages/express/test/rest.test.js
+++ b/packages/express/test/rest.test.js
@@ -249,7 +249,8 @@ describe('@feathersjs/express/rest provider', () => {
                 arguments: ['dishes', paramsWithHeaders ],
                 type: 'error',
                 method: 'get',
-                path: 'hook-error'
+                path: 'hook-error',
+                original: data.hook.original
               },
               error: { message: 'I blew up' }
             });
--- a/packages/feathers/lib/hooks/base.js
+++ b/packages/feathers/lib/hooks/base.js
@@ -3,16 +3,14 @@ const { _ } = require('@feathersjs/commons');
 const assignArguments = context => {
   const { service, method } = context;
   const parameters = service.methods[method];
-  const argsObject = context.arguments.reduce((result, value, index) => {
-    result[parameters[index]] = value;
-    return result;
-  }, {});
 
-  if (!argsObject.params) {
-    argsObject.params = {};
+  context.arguments.forEach((value, index) => {
+    context[parameters[index]] = value;
+  });
+
+  if (!context.params) {
+    context.params = {};
   }
-
-  Object.assign(context, argsObject);
 
   return context;
 };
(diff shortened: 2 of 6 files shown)


vonagam:
What about adding `finally` to [the list of hook types](https://github.com/feathersjs/feathers/blob/f9d8536c400caba355ccc8a2cdf171dcd9cff555/packages/feathers/index.d.ts#L116)?

daffl:
This makes a lot of sense, thank you! That the `finally` hooks are not documented is somewhat intentional because they were really only intended to be used internally (to send events). For the next major version I'd prefer the Koa style hook system https://github.com/feathersjs/feathers/issues/932 which would allow a `finally` like this:

```js
app.service('users').hooks({
  create: [
    async (context, next) => {
      const finally = async () => {
        // Finally code here
      }

      try {
        await next();
        await finally();  
      } catch (error) {
        await finally();
        throw error;
      }
    }
  ]
}
```

A convenience wrapper may also make this shorter.

vonagam:
Cool about koa style, prefer it too.

By the way your example snippet has the same problem as this pull request solves for current system: if your `finally()` function throws it will run twice. Here is how it should be written: 

Instead of:
```js
try {
  await next();
  await finally();  
} catch (error) {
  await finally();
  throw error;
}
```

It should be:
```js
try {
  await next();
} finally {
  await finally();
}
```

daffl:
Ah, totally. That's totally it.

PR #1421, opened 2019-06-27
fix: @feathersjs/express: allow middleware arrays

vonagam:
Using as many middleware arrays or functions as you want is valid for express, since it uses `array-flatten` on passed arguments ([1](https://github.com/expressjs/express/blob/e1b45ebd050b6f06aa38cda5aaf0c21708b0c71e/lib/router/index.js#L448), [2](https://github.com/expressjs/express/blob/e1b45ebd050b6f06aa38cda5aaf0c21708b0c71e/lib/router/route.js#L194)).

Now does not throw on valid usage for express and allows to use arrays of middleware before and after feathers service.

### DIFF
--- a/packages/express/lib/index.js
+++ b/packages/express/lib/index.js
@@ -28,7 +28,7 @@ function feathersExpress (feathersApp) {
       let middleware = Array.from(arguments)
         .slice(1)
         .reduce(function (middleware, arg) {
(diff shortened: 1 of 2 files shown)


daffl:
🎉 Will also go out with the next pre-release next week.

PR #1422, opened 2019-06-27
feat: helper script to fail tests on unhandled rejections

vonagam:
Usage with mocha: `--require @feathersjs/adapter-tests/lib/rejections`.

### DIFF
--- a/packages/adapter-tests/lib/rejections.js
+++ b/packages/adapter-tests/lib/rejections.js
@@ -1,0 +1,15 @@
+/* tslint:disable:no-console */
+
+let count = 0;
+
+process.on('unhandledRejection', (reason) => {
+  console.error('Unhandled rejection:', reason);
+  count++;
+});
+
+process.on('exit', (code) => {
+  if (count !== 0) {
+    console.error(`Total count of unhandled rejections: ${count}`);
+    process.exitCode = code || 1;
+  }
+});


daffl:
I locked the version in KnexJS via https://github.com/feathersjs-ecosystem/feathers-knex/commit/054360d0eef04c31ec2c95d989093edea7ffe51c so it is on the latest which doesn't have the unhandled promise rejection issue. Since it really was a bug in the upstream library do you think we'll still need this? I don't think I've seen this problem anywhere else before since actual test and adapter errors should still fail normally.

vonagam:
The problem was on both sides. 

Knex introduced code which produced unhandled rejections. That their fault. 

But on feathers side current test system did not catch those rejections, [all was green](https://github.com/feathersjs-ecosystem/feathers-knex/pull/208) and a pull request from greenkeeper was merged. That what should be fixed as unhandled rejections is a serious deal. And it will take only one line in `mocha.opts` to ensure this will not happen latter in knex adapter or any other.

Later on when node will change behaviour and start to crash on unhandled rejections by itself this file and the mocha option can be simply removed.

vonagam:
Changed the script. Now tests will continue to run after an unhandled rejection, but before process exit it will ensure that exit status is non-zero if there were any rejections. I think this is more suitable for tests than abrupt exit as you will be able to see tests results.

vonagam:
Alternative solution is to add [`loud-rejection`](https://github.com/sindresorhus/loud-rejection) to dependencies, then in adapter you use it like `--require loud-rejection/register`. This solution requires no additional file but an additional dependency.

vonagam:
Closing in favour of using `loud-rejection` directly in specific packages.

[Here](https://github.com/feathersjs-ecosystem/feathers-knex/pull/210) is PR for `feathers-knex`.

daffl:
That makes sense, thank you! Since it's only a development dependency I think this is fine.

PR #1426, opened 2019-07-02
fix: @feathersjs/adapter-commons: remove data from `remove` arguments

vonagam:


daffl:
Oops. I guess it worked anyway but definitely not correct. Thank you!

PR #1429, opened 2019-07-02
fix: @feathersjs/express: replace `reduce` with `map`

vonagam:
Easier to read and faster 4 times: 500,000 vs 2,000,000 ops/sec.

Also small difference (fix): before if you `activateHooks()` some method with a nonstandard argument resulted express args getter will throw an error, now it returns valid array with `undefined` for unknown arguments. (Vanilla feathers supports nonstandard arguments, but there is no test for such case.)

Also [`makeArguments`](https://github.com/feathersjs/feathers/blob/9585ac0ee8b051c869e1b351b15da202604af244/packages/commons/src/hooks.ts#L52) and [`defaultMakeArguments`](https://github.com/feathersjs/feathers/blob/9585ac0ee8b051c869e1b351b15da202604af244/packages/commons/src/hooks.ts#L34) can be safely removed before major release, since they have been replaced with [new system](https://github.com/feathersjs/feathers/blob/9585ac0ee8b051c869e1b351b15da202604af244/packages/feathers/lib/hooks/index.js#L56) and not used anywhere. If for some backward compatibility reason you want to keep them, we can replace their internals with same one liner:

```js
return hook.service.methods[hook.method].map((value) => hook[value]);
```

### DIFF
--- a/packages/express/lib/rest/getHandler.js
+++ b/packages/express/lib/rest/getHandler.js
@@ -32,16 +32,16 @@ function getAllowedMethods (service, routes) {
 }
 
 function makeArgsGetter (argsOrder) {
-  return (req, params) => argsOrder.reduce((result, argName) => {
+  return (req, params) => argsOrder.map((argName) => {
     switch (argName) {
       case 'id':
-        return [ ...result, req.params.__feathersId || null ];
+        return req.params.__feathersId || null;
       case 'data':
-        return [ ...result, req.body ];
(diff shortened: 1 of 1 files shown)


daffl:
👍 Looks like it is indeed not used but I think it makes sense to keep it in for backwards compatibility of other libraries that might still use it. Can definitely be removed with the new hook system in v5.

PR #1437, opened 2019-07-04
Add missing types property to package.json files

ihadeed:
Packages with typings need to have `types`/`typings` property populated so the TypeScript compiler knows where to find it.


daffl:
Thank you for the pull request. I haven't noticed that this is necessary if the `.d.ts` files are right alongside their transpiled files though.

ihadeed:
I think it might have been my IDE only (WebStorm); can't remember if the compiler had any warnings, it compiled fine.

I added `node_modules/@feathersjs/` to my `typeRoots` which fixed the IDE issues and caused the compiler to throw warnings. Adding `types` to those `package.json` files seems to work well with IDE + compiler without modifications to the `tsconfig.json`.

Here is my config just in case:
```json
{
  "compilerOptions": {
    "module": "commonjs",
    "target": "es2018",
    "outDir": "dist",
    "moduleResolution": "node",
    "baseUrl": ".",
    "lib": [
      "es2018"
    ],
    "declaration": false,
    "allowUnreachableCode": false,
    "typeRoots": [
      "node_modules/@types"
    ],
    "sourceMap": true,
    "esModuleInterop": true
  },
  "include": [
    "src/**/*"
  ],
  "exclude": [
    "node_modules",
    "**/*.spec.ts"
  ]
}
```

daffl:
I haven't noticed it with VSCode or the TS compiler. Does it work for you with `"types": "lib"`? If it does, I think that would make more sense since it can be more than one file.

ihadeed:
Just gave it another round of testing.

1. uninstalled the packages _(auth, auth-local)_ from my local `link` and re-installed from npm:
  - app compiles fine
  - no IDE errors
  - if I ask the IDE for suggestions from the `@feathersjs/authentication-local`, it doesn't give me anything
  - if I type `LocalStrategy` and let the IDE find it, it imports it from `@feathersjs/authentication-local/lib` since the typing are in that directory and they don't define/declare the main module

2. tried providing paths to definitions manually in my tsconfig and that did not help, same results as # 1 

3. re-linked my local modified libraries
  - auto complete & suggestions worked as expected: 
image
  - using the "go to declaration" feature takes me to the source directory though, not the `.d.ts`
  - deleted the `src` directory and tried "go to declaration" again, it took me to `.d.ts`


---

According to this: https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#including-declarations-in-your-npm-package

The `types`/`typings` property is required unless if your module has `index.js` and `index.d.ts` on the root level -- which is not the case here.

I hope that helps.

PR #1468, opened 2019-07-18
fix: @feathersjs/adapter-commons: `update` id is non-nullable

vonagam:
Reflect in id type that `update` method never supports multi-record changes.


daffl:
Thank you!

PR #1506, opened 2019-08-15
Conditional removeAccessToken for 401/404 case

KidkArolis:
This would allow to override the removeAccessToken() method to conditionally not remove the token. In particular my use case is something like:

```js
if (err.code === 401 || err.code === 404) {
  // remove the token
} else {
  // temporary issue, show an error screen
}
```

Not sure this implementation is the best, maybe an option of something like `removeTokenOnError: err => bool` in `auth` would be nicer?

### DIFF
--- a/packages/authentication-client/src/core.ts
+++ b/packages/authentication-client/src/core.ts
@@ -107,7 +107,7 @@ export class AuthenticationClient {
       });
   }
 
-  removeAccessToken () {
+  removeAccessToken (error?: Error) {
     return this.storage.removeItem(this.options.storageKey);
   }
 
@@ -134,7 +134,7 @@ export class AuthenticationClient {
           accessToken
         });
       }).catch((error: Error) =>
-        this.removeAccessToken().then(() => Promise.reject(error))
+        this.removeAccessToken(error).then(() => Promise.reject(error))
       );
     }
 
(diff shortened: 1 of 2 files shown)


KidkArolis:
Introducing new options should probably be avoided.. so also wondering about alternative solutions.

daffl:
Could you maybe add a test case for how the flow would look like? It's also always possible to extend the client class and provide your own functionality.

KidkArolis:
I can add the test case, if you think this feature is something worth adding.

My thinking is that if we only cleared token on 401/404 errors, that might be a better default. Right now, if you drop a connection because the server is restarting (you only have one, or kubernetes is spinning your pod up and down), Feathers will try to reconnect (which is great), but if that errors with some temporary 502 Bad Gateway, the token gets wiped.

I'm _ok_ with having to extend the service, but what pushed me to explore a PR is that there is no "good" way to extend. Pretty much the only option is to completely reimplement `reAuthenticate` method, since `removeToken` doesn't receive the error.

What I ended up doing instead:

```js
  // feathers removes access token too agressively
  // in case of temporary server errors, e.g. 500 or timeouts
  // we don't want to remove the token
  feathers.authentication._removeAccessToken =
    feathers.authentication.removeAccessToken
  feathers.authentication.removeAccessToken = function done() {
    return Promise.resolve(true)
  }
```

And then calling `feathers.authentication._removeAccessToken` manually in my logout function. Although, this means that my reconnect might be broken now. Unless it keeps retrying.. Hm...

PR #1520, opened 2019-08-22
__hooks become writable and configurable

bertho-zero:
We should be able to modify or delete this value.

In a particular case it will avoid me this kind of error:
```
Uncaught TypeError: Cannot redefine property: __hooks
```
```
Uncaught TypeError: Cannot delete property '__hooks' of #<Object>
```

### DIFF
--- a/packages/commons/src/hooks.ts
+++ b/packages/commons/src/hooks.ts
@@ -154,7 +154,9 @@ export function enableHooks (obj: any, methods: string[], types: string[]) {
 
   // Add non-enumerable `__hooks` property to the object
   Object.defineProperty(obj, '__hooks', {
-    value: hookData
+    configurable: true,
+    value: hookData,
+    writable: true
   });
 
   return Object.assign(obj, {


daffl:
What would be the use case for this? It's usually not ideal having to modify undocumented framework internals in an application. For skipping hooks we can now e.g. use the [adapter hook-less service methods](https://crow.docs.feathersjs.com/guides/migrating.html#hook-less-service-methods).

bertho-zero:
I wish I could edit or delete hooks, in my case I use uberproto to mixin a service in a const. This is not a normal use of Feathers but why protect these values?

```js
const Users = require('../services/users');

const service = Object.create(Users.prototype);

module.exports = service; // like a factory ¯\_(ツ)_/¯

module.exports.init = () => {
  const instance = new Users({ /* options */ });

  const subApp = feathers()
    .use( '/', instance )
    .setup();

  const svc = subApp
    .service('/')
    .hooks(hooks);

  Proto.mixin(svc, service);
};
```

This is not a choice but an imposed technical constraint.

However, I do not have these errors anymore.

bertho-zero:
This property just needs not to be enumerable.

daffl:
Similar to my comment in #1521 but I guess this one makes sense for testing purposes. Although this will change in the future anyway via #932.

bertho-zero:
It's actually in a test suite that it stuck, because of the init called several times.

These are internal values but if we want to modify or delete them in cases of advanced or atypical use we should be able to do it. https://github.com/feathersjs/feathers/pull/1520#issuecomment-524260203

PR #1521, opened 2019-08-22
Configurable and writable service properties

bertho-zero:
(see https://github.com/feathersjs/feathers/pull/1520)


daffl:
What would be the use cases for this? These were intentionally made not configurable because they are internal and shouldn't have to be modified in a normal application. If they do it either means some external features is missing or the app is trying to do things that are not intended.

PR #1525, opened 2019-08-24
fix: Only remove token on NotAuthenticated error in authentication client and handle error better

### LINKED ISSUE #1522: Authentication-client removes acces token on failed reauthenticate.
https://github.com/feathersjs/feathers/blob/6d723e81509c8e3c3020b723c696e33150ff2023/packages/authentication-client/src/core.ts#L136-L138

Authentication client removes authentication token on every failed reauthentication. It's not referenced in docs and is problematic when authentication server goes offline.
In my use case, I want to check if user can be authenticated with current token. No matter what the error is, on fail, it's removed - I would like to keep the token and try again later.

daffl:
This PR adds a `handleError` method that only removes the access token on `NotAuthenticated` errors. It can be [customized](https://crow.docs.feathersjs.com/api/authentication/client.html#customization) for custom error handling logic.

- Closes #1522
- Closes #1506

### DIFF
--- a/packages/authentication-client/src/core.ts
+++ b/packages/authentication-client/src/core.ts
@@ -119,12 +119,12 @@ export class AuthenticationClient {
   }
 
   handleError (error: FeathersError, type: 'authenticate'|'logout') {
-    if (error.code === 401) {
+    if (error.code === 401 || error.code === 403) {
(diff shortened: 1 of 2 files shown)


KidkArolis:
This is great! I would add 404 for cases where the token is old and the user has been deleted and authenticate() returns 404

PR #1551, opened 2019-09-09
fix: Fix regression in transport commons

### LINKED ISSUE #1547: Regression in @featherjs/transport-commons
`@feathersjs/transport-commons` was already at v4 so needs to be backwards compatible. There was a regression introduced in the `getDispatcher` method causing the following error:

Screen Shot 2019-09-06 at 10 00 37 AM

This was introduced in https://github.com/feathersjs/feathers/commit/64807e36bbf7df1def6c0444da5b4714b2931dba#diff-0e4e91edd89fb128213ae8cbca1a0c14L38 and should be possible to make backwards compatible again by checking for `instanceof WeakMap`.

daffl:
Closes #1547


daffl:
Fixed in v4.3.1

PR #1552, opened 2019-09-09
fix: Use long-timeout for JWT expiration timers

### LINKED ISSUE #1546: JWT Token expireIn > 24d causes a overflow error with Sockets
### Information about the bug
When you set the JWT Token `expireIn` to greater than 24d (~24,855) it causes a overflow error when you authenticate with sockets, REST is not affected by this.

This means that a connection in the `app.channel('authenticated');` gets removed, which in turn means the events from feathers does not get published to clients. 
But it still works as a valid authentication token.

I figured the timeout timer is to remove a connection from `app.channel('authenticated')` when the token expires.
I found the part which removes it in the `authentication` library. [Link](https://github.com/feathersjs/feathers/blob/master/packages/authentication/src/jwt.ts#L36-L40)

Snippet from the file [jwt.ts](https://github.com/feathersjs/feathers/blob/master/packages/authentication/src/jwt.ts)
```javascript
      // The time (in ms) until the token expires
      const duration = (exp * 1000) - new Date().getTime();
      // This may have to be a `logout` event but right now we don't want
      // the whole context object lingering around until the timer is gone
      const timer = setTimeout(() => this.app.emit('disconnect', connection), duration);
```
The problem is that `setTimeout()` does not fit more that 32bit, see [here](https://stackoverflow.com/questions/3468607/why-does-settimeout-break-for-large-millisecond-delay-values), and in the documentation for [NodeJS](https://nodejs.org/api/timers.html#timers_settimeout_callback_delay_args) it says 

> When delay is larger than 2147483647 or less than 1, the delay will be set to 1. Non-integer delays are truncated to an integer.

### Steps to reproduce
Create a new project, set the `expireIn` to >24d, then authenticate using a socket.

### Expected behavior
When a socket authenticates it should still be in the connection pool of `app.channel('authenticated');`

### Actual behavior
It authenticates then it is removed from the `app.channel('authenticated');` immediately.

### System configuration

```json
{
    "@feathersjs/authentication": "^4.3.0",
}
```

**NodeJS version**: v12.9.1

daffl:
Closes #1546

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3,20 +3,131 @@
 	"requires": true,
 	"lockfileVersion": 1,
 	"dependencies": {
+		"@angular/common": {
+			"version": "8.2.5",
+			"resolved": "https://registry.npmjs.org/@angular/common/-/common-8.2.5.tgz",
+			"integrity": "sha512-7iSDLVhS+jbVRkECpbTzU9+6IQPS3Wl0CF73EA0sdzPbTC2GKvGfM9WLnIZZIxewkii6Wn1Yb0x0qRdWMT2STA==",
+			"requires": {
+				"tslib": "^1.9.0"
+			}
+		},
+		"@angular/core": {
+			"version": "8.2.5",
+			"resolved": "https://registry.npmjs.org/@angular/core/-/core-8.2.5.tgz",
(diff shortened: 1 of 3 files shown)


daffl:
Fixed in v4.3.1

PR #1558, opened 2019-09-13
 Consistency in path resolution for configuration values provided from environment variables.

deskoh:
### Summary
With the following `config/default.json`:

```jsonc
{
  "frontend": "../public",
  //...
}
```

`app,get('frontend')` would return something like `path/to/app/public`.

However, if we specify the same value using environment variable, i.e. `FRONTEND=../public` with the following:

```jsonc
{
  "frontend": "FRONTEND",
  //...
}
```

One would expect the same result, however we would get `../public` as environment variable replacement and absolute path resolution is mutually exclusive.

This PR standardize aims to standardize this behavior.

The [documentation](https://docs.feathersjs.com/api/configuration.html#variable-types) does not seem to contradict either results.


daffl:
Totally makes sense, thank you!

PR #1563, opened 2019-09-17
Removed opinionated protocol setting

troykelly:
**Problem**
OAuth requests do not possess the ability to set the protocol and port from the configuration.
In addition, when in a development environment, the opinionated nature of the protocol setting forced a `redirect_url` with only a `http://` protocol.

### DIFF
--- a/packages/authentication-oauth/src/index.ts
+++ b/packages/authentication-oauth/src/index.ts
@@ -24,21 +24,21 @@ export const setup = (options: OauthSetupSettings) => (app: Application) => {
   }
 
   const { strategyNames } = service;
-  
+
   // Set up all the defaults
   const { path = '/oauth' } = oauth.defaults || {};
   const port = app.get('port');
+  const ignorePort = (app.get('ignorePort') === true) ? true : false;
+
   let host = app.get('host');
-  let protocol = 'https';
-  
+  let protocol = app.get('protocol');
   // Development environments commonly run on HTTP with an extended port
-  if (app.get('env') === 'development') {
-    protocol = 'http';
-    if (String(port) !== '80') {
-      host += ':' + port;
-    }
+  if (!protocol) {
+    protocol = (app.get('env') === 'development') ? 'http' : 'https';
   }
-  
+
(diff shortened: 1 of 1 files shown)


KidkArolis:
This is the part of the code that was just refined 2 weeks ago - https://github.com/feathersjs/feathers/pull/1543 and we were pretty happy with the result.

Could you describe what you're trying to do that's not working as expected (e.g. give example URLs)? You can configure the port/protocol business for each environment in your configuration files if the defaults don't work:

```
authentication:
  oauth:
    defaults:
      host: custom-host.with-any-port.com:123
      protocol: https
```

Defaults are designed to fit typical use cases:
- in dev - http with port
- in prd - https with no port

troykelly:
I'm trying to not have http:// appended to the URI.
I also need a mechanism to "ignore" the port setting - for example, when running in App Engine, Google Run, or using a service like Ngrok. The local service may be port 3030, but publically it is https://example.com/

KidkArolis:
OK. I see your issue, but I think this is all configurable. In your config file (I'm using yml, but it's the same in json), there are 2 significant parts of config for your use case:

```yml
# this is how the app is running locally
# this is also used to generate defaults for oauth
host: localhost
port: 3030

# but you can override oauth defaults to whatever you need
authentication:
  oauth:
    defaults:
      host: custom-host-with-or-without-port.com
      protocol: custom-protocol-irrespective-of-environment
```

If you don't put anything in `authentication.oauth.defaults` - that's when feathers generates that based on "good defaults" heuristic - http with port in dev, https with no port in production.

Since you want to use no port in development, you can just override that directly with:

```yml
authentication:
  oauth:
    defaults:
      host: myapp.ngrok.io
      protocol: https
```

Let me know if that worked or if I'm still missing something.

The relevant part of the code btw is this bit where feathers deeply merges the generated defaults with your own defaults (and the rest of oauth configuration): https://github.com/compwright/feathers/blob/master/packages/authentication-oauth/src/index.ts#L28-L35

daffl:
Thank you for the pull request! As per the discussion, I think we should leave it the way it is for now since it is always possible to customize those settings.

PR #1571, opened 2019-09-21
fix: check for undefined access token

### LINKED ISSUE #1568: feathers auth assumes an accessToken exists for socket (regression)
Prior to https://github.com/feathersjs/feathers/commit/2707c3331721be6a4a0a943b0fbc118379a1ad2a#diff-847cf0caf6782a09a6a9cf0969d02f50R35 in 4.3.0 you could (in an after hook) remove the `accessToken` from `hook.result` in an after hook.  We are doing this to enforce a 2fa flow in our application, to request further info in a future auth request.  We however still wanted to return a 20x response code with the request.

With the above the code assumes there is an access token in the result, and throws an ugly "jwt must be provided" error in this situation.

Is it reasonable to assume that a successful call to authenticate always includes an accessToken, and we need to rethink our design?  Or should this socket code check for accessToken existence prior to running this block of code?

We could return an error in this case if that's our only option.

keywordnew:
### Summary

This PR adds a check for if the access token is undefined. 

Fixes #1568.

Nope, AFAIK.

### Other Information

This fix does not yet have a test.

### DIFF
--- a/packages/authentication/src/jwt.ts
+++ b/packages/authentication/src/jwt.ts
@@ -32,8 +32,10 @@ export class JWTStrategy extends AuthenticationBaseStrategy {
     const isValidLogout = event === 'logout' && connection.authentication && authResult &&
       connection.authentication.accessToken === authResult.accessToken;
 
-    if (authResult && event === 'login') {
-      const { accessToken } = authResult;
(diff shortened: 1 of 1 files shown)


daffl:
Awesome! 🎉

PR #1572, opened 2019-09-22
Nudge grant version up

8BallBomBom:
Getting this error when updating packages via npm.

Warning "@feathersjs/authentication-oauth > grant-profile@0.0.5" has incorrect peer dependency "grant@>=4.6.3".

Seems grant-profile is set to depend on grant 4.6.3 or higher where as we are using 4.6.2
Pretty straight forward update.
Might be some other changes to take note of [here](https://github.com/simov/grant/blob/master/CHANGELOG.md#v463-20190907).


daffl:
Thank you! Will go out shortly.

PR #1585, opened 2019-09-25
enable async hooks

bertho-zero:
https://github.com/feathersjs/feathers/issues/932


bertho-zero:
Works well with https://github.com/koajs/compose/pull/102

daffl:
Sweet. This is great. Eventually I'd like to look into replacing all the old hook processing code with the async hooks but this is a great start and good to know that it was fairly easy to integrate.

PR #1590, opened 2019-09-28
allow to unshift hooks

bertho-zero:
Allow registering of hooks with `$unshift` and `$push` like this:

```js
app.hooks({
  before: {
    find: {
      $unshift: someHook()
    }
  }
})
```


bertho-zero:
@daffl I think similar functionality is required in `@feathersjs/hooks`, we may want to define basic hooks at the service level and possibly add a more global validator when integrating the service into the app.

daffl:
I opened up a follow-up discussion for this in https://github.com/feathersjs/hooks/issues/11

daffl:
I'm going to close this for #1798 (which has passing tests now)

PR #1594, opened 2019-10-01
Add qs options

DaddyWarbucks:
### Summary

- [ x] Tell us about the problem your pull request is solving.
This change allows the client to pass options to the underlying `qs` library that parses query 
objects into query strings. I initially discovered the need for this as I want to parse `null` differently than an empty string `''`. QS  by default treats these as the same, but you can pass an option `strictNullHandling ` to handle that differently.

- [ x] Are there any open issues that are related to this?
Not to my knowledge.

- [ x] Is this PR dependent on PRs in other repos?
Not to my knowledge

### Other Information

I wasn't sure if this is something that should be tested?


DaddyWarbucks:
In hindsight, it may be better to create a new option called `stringifyQuery` (open to suggestions on names) that takes a function. 
Then it is left to the developer to use _any_ parser they want, but fallback to basic `qs` if none is added.
```
{ stringifyQuery: query => qs.stringify({ strictNullHandling: true  }) }

...
const queryString =  this.options.stringifyQuery
  ? this.options.stringifyQuery(params)
  :  query.stringify(params);
```

daffl:
Yes, I think a `getQuery` method that can be overriden makes the most sense so something like this:

```js
makeUrl (params, id) {
  params = params || {};
  let url = this.base;

  if (typeof id !== 'undefined' && id !== null) {
    url += `/${encodeURIComponent(id)}`;
  }

  return url + this.getQuery(params);
}

getQuery (params) {
  if (Object.keys(params).length !== 0) {
    const queryString = query.stringify(params);

    return `?${queryString}`;
  } 

  return '';
}
```

bertho-zero:
The `addQueryPrefix` option seems to meet the need:
```js
getQuery(params) {
  return qs.stringify(params, { addQueryPrefix: true });
}
```

```js
qs.stringify({ a: 'b', c: 'd' }, { addQueryPrefix: true }) // "?a=b&c=d"
qs.stringify({}, { addQueryPrefix: true }) // ""
```

PR #1599, opened 2019-10-03
add Service core class

bertho-zero:
This PR makes it possible to use hooks with services outside of an app.

For this the class of the service must extend `Service`.

```js
class Users extends Service {
 // ...
}
```

To enable hooks you can use the `methods` option of the constructor or the `enableHooks` method.
After that the `hooks` method is available.

```js
class Users extends Service {
  constructor(options) {
    super({
      methods: {
        sayHello: ['data', 'params']
      }
    })
  }

  // ...
}

const users = new Users();

users.setup(); // add hooks to method calls

/**********/
/*** or ***/
/**********/

class Users extends Service {
  // ...
}

const users = new Users();

users.enableHooks({
  sayHello: ['data', 'params']
});
users.setup();
```

`AdapterService` extends `Service`, so it is possible to use DBs outside of an app.
(tested with feathers-knex)


bertho-zero:
This example works:

```js
const { Service: KnexService } = require('feathers-knex');

const Users = new KnexService({
  Model: db,
  name: 'people'
}).setup();

// Hooks service directly !
users.hooks({
  before: transaction.start({ getKnex: () => db }),
  after: transaction.end(),
  error: transaction.rollback()
});

const app = feathers();

app.use('/users', users); // Add `app` and `location` for hooks (and events system)
```

daffl:
So I've been meaning to revisit the new hooks system hopefully soon, my question here would be if you think this would still make sense if going forward we can do things like

```js
const hooks = require('@feathersjs/hooks')

class MyService {
  async sayHello (name) {
    return `Hello ${name}`;
  }
}

// Service level hooks
hooks(MyService, {
  sayHello: [ async (context, next) => {
    await next();
  }]
});

const serviceInstance = new MyService();
// Instance level hooks
hooks(serviceInstance, {
  sayHello: [ async (context, next) => {
    context.result.replace('!!!', '');
    await next();
  }]
});
```

bertho-zero:
That would be great !
That would be possible with the `hooks` branch, this PR is for `master`.

PR #1605, opened 2019-10-07
fix: Small improvements

vonagam:
Small types and consistency style fixes. But also one real fix:

`handleConnection` method of `AuthenticationBase` and `JWTStrategy` was never called in response to `disconnect` event, so when a jwt expiration timer emitted `disconnect` `connection.authentication` was not removed and the timer was not cleared. Added listener for `disconnect` event to fix this. Also added `expirationTimers.delete` after a timer is cleared and `lt.clearTimeout` before a timer is set (to ensure that only one timer per connection exist at a time, otherwise multiple calls to `reAuthenticate` with force can create more than one timer).

### DIFF
--- a/packages/authentication-client/src/core.ts
+++ b/packages/authentication-client/src/core.ts
@@ -149,7 +149,7 @@ export class AuthenticationClient {
     return authPromise;
   }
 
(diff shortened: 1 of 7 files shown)


daffl:
That makes sense, thanks! I thought this was being tested but I guess the test only covered that the authentication information had been removed.

PR #1607, opened 2019-10-07
fix: Check query  for NaN

### LINKED ISSUE #1606: [Security] Querying $limit=null
### Steps to reproduce
Do HTTP request to any service with parameter `$limit=null`. This allow to users request data without limit. This happen with a value equal to `null` or any `string`.

### Expected behavior
Add validation for not allow to externals request to access to unlimited data with `$limit` value different to a number.

### Actual behavior
It allow to users access without limit, bypassing `paginate.max` value in `default.json` and causes an overload in server.

daffl:
Closes #1606

### DIFF
--- a/packages/adapter-commons/src/filter-query.ts
+++ b/packages/adapter-commons/src/filter-query.ts
@@ -13,7 +13,7 @@ function parse (number: any) {
 // default and max pagination settings
 function getLimit (limit: any, paginate: any) {
   if (paginate && paginate.default) {
-    const lower = typeof limit === 'number' ? limit : paginate.default;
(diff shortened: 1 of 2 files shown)


daffl:
All dependencies should be updated and published in

- [ ] feathers-memory
- [ ] feathers-localstorage
- [ ] feathers-nedb
- [ ] feathers-mongodb
- [ ] feathers-mongoose
- [ ] feathers-sequelize
- [ ] feathers-knex
- [ ] feathers-elasticsearch
- [ ] feathers-objection

Will also add deprecation messages for earlier versions as suggested by @bitflower.

Issue #1591, opened 2019-09-30
Set oAuth redirect URL dynamically


fadiquader:
### Steps to reproduce
I'm working on a project where I need to set the oAuth redirect URL dynamically (I need to set the domain or subdomain dynamically). I've read that this is possible in the latest version of Feathers (https://github.com/feathersjs/feathers/issues/990). However, I've struggled to make this work in practice. Looking at the code, it seems that it is not possible to dynamically set the redirect URL -- what I see in the code here just uses the redirect URL from the environment variables: https://github.com/feathersjs/feathers/blob/master/packages/authentication-oauth/src/strategy.ts#L75

### Expected behavior
I'd like to be able to dynamically set the redirect URL, changing the domain or subdomain. This seems to be possible with Grant directly, but I can't see how to do this with Feathers.

### Actual behavior
I can't see any way to do this! Would really appreciate your advice on how to do this (if it's possible)

### System configuration

Tell us about the applicable parts of your setup.

**Module versions** (especially the part that's not working):
Feathers v4 (latest)

daffl:
All methods can be easily customized by overriding them in your own class which is [documented here](https://docs.feathersjs.com/api/authentication/oauth.html#customization) and also shown with a GitHub login [in the guide](https://docs.feathersjs.com/guides/basics/authentication.html#github-login-oauth). In your case by overriding the `getRedirect` method you linked to:

```js
class MyDynamicRedirectStrategy extends OAuthStrategy {
  async getRedirect(authResult) {
    const { user } = authResult;
    // Get the redirect url e.g. from the users organization
    const { redirectUrl } = await app.service('organizations').get(user.organizationId);
    // This is necessary if it should work with the standard Authentication
    // client (which could be customized as well)
    const query = authResult.accessToken ? {
      access_token: authResult.accessToken
    } : {
      error: data.message || 'OAuth Authentication not successful'
    };

    return `https://${redirectUrl}#${querystring.stringify(query)}`;
  }
}

// ...
authentication.register('github', new MyDynamicRedirectStrategy());
```

fadiquader:
@daffl 
Thank you for getting back, what I mean is by accessing req.query for example so I’m unable to access the req object, because i want to set it dynamically by querying backend.

daffl:
I'm assuming this is what [Grants dynamic overrides](https://github.com/simov/grant#dynamic-override) are for.

fadiquader:
@daffl 
[...]

PR #1608, opened 2019-10-08
feat(authentication-oauth): Set oAuth redirect URL dynamically

fadiquader:
### Summary
Referred to this bug https://github.com/feathersjs/feathers/issues/1591 
Fixed set oauth redirect URL dynamically.

### DIFF
--- a/packages/authentication-oauth/src/express.ts
+++ b/packages/authentication-oauth/src/express.ts
@@ -50,12 +50,20 @@ export default (options: OauthSetupSettings) => {
 
     authApp.get('/:name/authenticate', async (req, res, next) => {
       const { name } = req.params as any;
-      const { accessToken, grant, query } = req.session;
+      const { accessToken, grant, query = {} } = req.session;
       const service = app.defaultAuthentication(authService);
       const [ strategy ] = service.getStrategies(name) as OAuthStrategy[];
+      const params = {
+        authStrategies: [ name ],
+        authentication: accessToken ? {
+          strategy: linkStrategy,
+          accessToken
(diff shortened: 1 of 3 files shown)


daffl:
Thank you! I updated the PR with a test and changed it to pass all of `params` to the `getRedirect` method.

The default oAuth strategy now also includes the original `redirect` setting. So a query redirect like `?redirect=/dashboard` with a `oauth.redirect` option of `https://app.mydomain.com` will redirect to `https://app.mydomain.com/dashboard`. The redirect can't be a generic URL, otherwise an attacker could steal a token by sending a link that redirects to their URL.

elv1n:
@daffl Redirect only works when `featers_token` provided to link accounts. It will be extremely useful to have a query in session to manipulate with it in `getRedirect` fn.
Are there any downsides to have the feature?

daffl:
Are you sure? This should work whenever the `?redirect=` query parameter is set during oAuth login.

elv1n:
Sorry, if I missed something
Here is a skip part
```
 if (feathers_token) {
        debug(`Got feathers_token query parameter to link accounts`, feathers_token);
        req.session.accessToken = feathers_token;
        req.session.query = query;
      }
```
and then it takes from session on authenticate route
```
 authApp.get('/:name/authenticate', async (req, res, next) => {
      const { name } = req.params as any;
      const { accessToken, grant, query = {} } = req.session;
```

daffl:
Dang, you're totally right, that should definitely live outside of that statement. PR would be welcome.

PR #1609, opened 2019-10-08
fix: Typings for express request and response properties

vonagam:


packages/express/index.d.ts
```diff
@@ -36,3 +36,14 @@ interface FeathersExpress extends Express {
 declare namespace feathersExpress {
     type Application<T = any> = express.Express & FeathersApplication<T>;
 }
+
+declare module 'express-serve-static-core' {
+    interface Request {
+        feathers?: Partial<FeathersParams>;
```
daffl: Should this also include `app?: Application<T = any>`?

vonagam: `app` is already defined on `Request` [here](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/f58986ca20210f8e6612942db077b9228740a315/types/express-serve-static-core/index.d.ts#L474) by express itself.

But we can add this:
```ts
interface Application extends FeathersApplication {
}
```

Unfortunately tslint complains about empty interface... Will have to figure out how to silence it.

daffl: Wouldn't we be able to just override the property type anyway?

vonagam: Nope

vonagam: Ok. Because in types definition for express app `set`/`enable`/`disable` return `Application` instead of `this` there is a type conflict which i cannot resolve...

daffl: Fair. We'll leave it at that for now then. Thanks again!

PR #1611, opened 2019-10-09
feat(authentication): Add setup method for auth strategies

vonagam:
Add `setup` method to `AuthenticationStrategy` by analogy with service setup. More convenient alternative to `setAuthentication`/`setApplication`/`setName` and removes the need for getters, since it ensures that all services are already registered by the time of a call.

For example, right now every time you get [`entityService`](https://github.com/feathersjs/feathers/blob/6a56056df623ecf4c0d8b43688b7787976a3f853/packages/authentication/src/strategy.ts#L25) in a strategy you call:
1) [`AuthenticationBaseStrategy.configuration`](https://github.com/feathersjs/feathers/blob/6a56056df623ecf4c0d8b43688b7787976a3f853/packages/authentication/src/strategy.ts#L26)
2) [`AuthenticationBase.configuration`](https://github.com/feathersjs/feathers/blob/6a56056df623ecf4c0d8b43688b7787976a3f853/packages/authentication/src/strategy.ts#L22)
3) [`Object.assign` and `app.get`](https://github.com/feathersjs/feathers/blob/6a56056df623ecf4c0d8b43688b7787976a3f853/packages/authentication/src/core.ts#L107)
4) [`app.service`](https://github.com/feathersjs/feathers/blob/6a56056df623ecf4c0d8b43688b7787976a3f853/packages/authentication/src/strategy.ts#L32)

All of those calls can be done once in `setup` method instead doing them every time.

Also fixed small inconsistencies in `app.setup`:

1) Moved [this line](https://github.com/feathersjs/feathers/blob/6a56056df623ecf4c0d8b43688b7787976a3f853/packages/feathers/lib/application.js#L143) up since otherwise a service registered in other service setup call will not get setup call.

2) Moved [those lines](https://github.com/feathersjs/feathers/blob/master/packages/feathers/lib/application.js#L121-L124) down so that setup call for a dynamically registered service happens after the service is registered just like for a usual service.


jnardone:
couldn't you do all of this in `setApplication`?

vonagam:
> it ensures that all services are already registered by the time of a call

`setApplication` (with `setName`, `setAuthentication` and `verifyConfiguration`) is [called](https://github.com/feathersjs/feathers/blob/dfba6ec2f21adf3aa739218cf870eaaaa5df6e9c/packages/authentication/src/core.ts#L129) in `register` method, before setup, and not everything is set up at that point (not all services registered).

Why have those separate methods when one will do? In Service you do not have `setApplication`, `setName` or `verifyConfiguration`, only `setup`.

jnardone:
Good points.  This would be very useful for a project I'm currently working on.

daffl:
Hm, sorry I'm not sure why this got closed. Going to reopen but will review again to add in v5 (which will also have asynchronous application setup).

vonagam:
Rebased.

daffl:
Will go out in the next prerelease. Better late than never I guess, sorry about that :smile:

PR #1620, opened 2019-10-14
fix: Remove adapter commons type alternatives

daffl:
Changes back some things from #1567 since it broke type compatibility in existing applications (like the feathers-chat and generated application).

### DIFF
--- a/packages/adapter-commons/src/service.ts
+++ b/packages/adapter-commons/src/service.ts
@@ -26,7 +26,7 @@ export interface ServiceOptions {
 }
 
 export interface InternalServiceMethods<T = any> {
-    _find (params?: Params): Promise<T[] | Paginated<T>>;
+    _find (params?: Params): Promise<T | T[] | Paginated<T>>;
     _get (id: Id, params?: Params): Promise<T>;
     _create (data: Partial<T> | Array<Partial<T>>, params?: Params): Promise<T | T[]>;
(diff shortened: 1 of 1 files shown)


vonagam:
That's why i wrote in point 3:

> Small breaking change, super easily fixed by removing `T |` from types of adapters in `_find` methods.

Were there other problems than this?

daffl:
Ah, sorry I missed that one, I thought only the other PR was a breaking change. But yes, basically aside from all TS app builds, all database adapter types broke because they were incompatible with this change. I'm planning on adding dtslint so that it should catch things like this hopefully earlier.

vonagam:
Ok, just to clarify and understand. 

I understand why [this line](https://github.com/feathersjs/feathers/blob/3931ee6928b6bddad511fe2efa708f3582a9c71d/packages/adapter-commons/src/service.ts#L29) was reverted to include `T` as possible return value. Because it is (small) breaking change and to fix it you need to change [this line](https://github.com/feathersjs-ecosystem/feathers-objection/blob/b69be2fcf167a7d8194ae0a5022a4540e85fbff0/types/index.d.ts#L26) in every adapter:
```ts
// from
_find(params?: Params): Promise<T | T[] | Paginated<T>>;
// to
_find(params?: Params): Promise<T[] | Paginated<T>>;
```

But was there a problem with overloads in `AdapterService` or were they reverted in bulk with `_find`?

PR #1677, opened 2019-11-11
Add scope parameter to feathers configuration

soulofmischief:
### Summary

I dug myself into edge-case city trying to use node-config with a monorepo architecture. While I would prefer a solution which allows each package to have its own config which extends the root, that sounds like an upstream problem, and so this solution provided a quick and useful escape hatch. Now I can scope each package under its own namespace in the root node config directory.

i.e. my configuration might look like

```
{
  client: {
    value: 42,
  }
  server: {
    port: 8000,
  }
}
```

and I would get the correct configuration with

```
process.env['NODE_CONFIG_DIR'] = path.resolve( paths.config )
const configuration = require( '@feathersjs/configuration' )

app.configure( configuration( 'server' ))
const port = app.get( 'port' )
```

preventing configuration data meant for other packages from being exposed.

If this is a welcome feature, I'll write the appropriate tests and documentation. I could also generalize init() to take a Symbol as well.

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,6 +1,6 @@
 # These are supported funding model platforms
 
-github: # Replace with up to 4 GitHub Sponsors-enabled usernames e.g., [user1, user2]
+github: [ daffl ]
 patreon: # Replace with a single Patreon username
 open_collective: feathers
 ko_fi: # Replace with a single Ko-fi username
--- a/changelog.md
+++ b/changelog.md
@@ -3,6 +3,18 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## [4.3.11](https://github.com/feathersjs/feathers/compare/v4.3.10...v4.3.11) (2019-11-11)
+
+
+### Bug Fixes
+
+* **authentication:** Retain object references in authenticate hook ([#1675](https://github.com/feathersjs/feathers/issues/1675)) ([e1939be](https://github.com/feathersjs/feathers/commit/e1939be19d4e79d3f5e2fe69ba894a11c627ae99))
+* **authentication-oauth:** Allow hash based redirects ([#1676](https://github.com/feathersjs/feathers/issues/1676)) ([ffe7cf3](https://github.com/feathersjs/feathers/commit/ffe7cf3fbb4e62d7689065cf7b61f25f602ce8cf))
+
+
+
+
+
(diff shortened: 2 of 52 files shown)


daffl:
This seems like a good addition to me. Would it be possible to

- Include a test (adding a nested config in https://github.com/feathersjs/feathers/blob/master/packages/configuration/test/config/default.json and creating a new application with that key in https://github.com/feathersjs/feathers/blob/master/packages/configuration/test/index.test.ts
- Add a PR via https://docs.feathersjs.com/api/configuration.html for to add this to the documentation

daffl:
Otherwise I'd have to close this PR which would be unfortunate.

soulofmischief:
@daffl Sorry, I've been on vacation and then caught the flu. Can we open the PR back up?

daffl:
Sure. Let me know if you need any help with the tests or docs.

soulofmischief:
Will do. I should have time tomorrow to wrap this PR up.

soulofmischief:
Update: I had a complication from the flu which led to another infection. I've been bedridden all of January. Antibiotics are working their magic and I'll write up that test and doc change in a few days after I've recovered.

soulofmischief:
Sorry for the delay. Turns out I'd already written tests... Let me know if the documentation isn't up to snuff.

daffl:
Thank you for updating that. The browser test does get flaky some times, I'm still not entirely sure why (Puppetteer was supposed to be more reliable than what we had before). Will go out with the next feature release.

soulofmischief:
@daffl I gave it some thought, and while I typically prefer brevity and for things to fail hard and fast, perhaps it would be better to make this code a bit more robust, returning an empty object instead of `undefined`, and possibly an error message to console. Thoughts?

daffl:
Hey @soulofmischief sorry it all took so long and thank you again for the pull request. After more deliberation on how to improve configuration handling, I created https://github.com/feathersjs/feathers/issues/1895 which discusses steps taken in the upcoming version to remove any custom functionality added by `@feathersjs/configuration` and this unfortunately also applies to this PR.

`@feathersjs/configuration` was always intended to be a tiny wrapper over https://github.com/lorenwest/node-config and it turned out that any customization of `node-config` behaviour  (like the automatic environment variable substitution) caused more confusion and problems than it was useful. Because it is so widely used, `node-config` should already provide battle tested features for almost all configuration use cases and we should rely on those or work together on making `node-config` itself better.

soulofmischief:
@daffl that's fine with me! If node-config is exposed more directly then implementing this on an ad-hoc basis is a cinch.

PR #1688, opened 2019-11-17
Allow specific service typings for `Hook` and `HookContext`

deskoh:
The `service` property of `HookContext` is of type `Service<T>`.

```ts
export default (options = {}): Hook => {
  return async (context: HookContext<User>) => {
    // No typings for `_create` method (catch-all definition will type unknown fields to `any)
    context.service._create(...)
    return context;
  };
}
```

This PR is backwards compatible and will allow the following:

```ts
import { Service } from 'feathers-mongoose';

export default (options = {}): Hook<User, Service<User>> => {
  return async (context) => {
    // Correct service typings.
    context.service._create(...)
    return context;
  };
}
```


daffl:
I've been meaning to do this, thank you!

PR #1696, opened 2019-11-20
fix(authentication-client): Reset authentication promise on socket disconnect

daffl:
This PR updates the authentication client to defer authentication as soon as a socket disconnects instead of only on socket reconnect because calls made while disconnected will still get a `NotAuthenticated` errors.


marshallswain:
Not sure what Code Climate's deal is, but everything else look like :shipit:

KidkArolis:
Tested in my app and it works great, much simpler than what I was doing.

Issue #1712, opened 2019-11-29
Query filtering does not filter objects in arrays


dysc:
### Steps to reproduce

```
app.service('users').find({
  query: {
    $and: [{
      $anything: 'function () { }'
    }]
  },
})
```

### Expected behavior

Keys prefixed with `$` that are not in `options.filters` and `options.whitelist` should throw an error, even when being used in an object within an array of the query.

### Actual behavior

```
MongoError: unknown top level operator: $anything MongoError: unknown top level operator: $anything
```

### System configuration

Tell us about the applicable parts of your setup.

**Module versions** (especially the part that's not working): All

**NodeJS version**: All

**Operating System**: All

**Browser Version**: All

**React Native Version**: All

**Module Loader**: All

PR #1713, opened 2019-11-29
fix: Clean arrays in query

dysc:
### Summary

Clean objects within arrays in feathers query (#1712)

### Other Information

This bug may expose attack surface


daffl:
Thank you for the pull request @dysc. This is a good fix but the tests should still be passing and ideally there would also be new tests to verify the new functionality.

PR #1716, opened 2019-12-01
Allow service typing to specify pagination option.

deskoh:
This PR allows narrowing of the return type of `find` by specifying pagination options.

```ts
type MessagesServiceOptions = Omit<Partial<ServiceOptions>, 'paginate'>;

export class Messages extends Service<any, MessagesServiceOptions > {
  constructor(options: MessagesServiceOptions , app: Application) {
    super(options);
  }
}

// When service is declared, the `find` method will exclude `Paginated<any>` from the return type
declare module '../../declarations' {
  interface ServiceTypes {
    'messages': Messages & ServiceAddons<any>;
  }
}

// messages is of type `Promise<any[]>`
const messages = service('messages').find(...);
```
For service with pagination, the declaration could be:

```ts
type MessagesServiceOptions = Partial<ServiceOptions> & { paginate: any };

// ...

// messages is of type `Promise<Paginated<any>>`
const messages = service('messages').find(...);
```

Aside, shouldn't find always return either `T[]` or `Paginated<T>`? Under what circumstances will `find` return type `T`?


daffl:
See discussion in https://github.com/feathersjs/feathers/pull/1837#issuecomment-612691538

PR #1732, opened 2019-12-10
fix: Add `params.authentication` type, remove `hook.connection` type

vonagam:
`connection` is in `Params` and defined in `transport-commons`.


daffl:
👍

PR #1733, opened 2019-12-10
feat: Add `exposed` property to hook context

vonagam:
The idea is that unless a context object is returned from the call hooks that modify context returning properties other than `result` (like `statusCode` and `dispatch` or other custom ones) may be skipped as those properties will not be used outside of hooks of the call itself.

Possible alternative names: `dispatching` or `public`.

### DIFF
--- a/packages/express/test/rest.test.ts
+++ b/packages/express/test/rest.test.ts
@@ -159,6 +159,7 @@ describe('@feathersjs/express/rest provider', () => {
           method: 'get',
           path: 'hook',
           event: null,
+          exposed: true,
           result: { description: 'You have to do dishes' },
           addedProperty: true
         });
@@ -245,6 +246,7 @@ describe('@feathersjs/express/rest provider', () => {
               event: null,
               method: 'get',
               path: 'hook-error',
+              exposed: true,
               original: data.hook.original
             },
             error: { message: 'I blew up' }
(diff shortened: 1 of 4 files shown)


vonagam:
Last one of those closings.
I will rebase it when asked.
Assuming that the proposition is accepted.

daffl:
Sounds good. Will definitely revisit this for the next version as well.

vonagam:
Rebased.

daffl:
I'm going to close this one since it has now diverged quite a bit and I think this can also be handled by checking for `params.provider`.

vonagam:
I'm ok with closing, since the hooks functionality was moved to separate repository and it should be added there, not here.

But no, just checking `params.provider` is not an option here: As you can see in ["Customizing and returning the context"](https://github.com/feathersjs/hooks#customizing-and-returning-the-context) section, you can call function just to get a `context.result` (in this case context will be internal and will not be exposed to the caller) or you can call it with a context as a last argument and in this case you will receive a full context with all info exposed. Params that are set on a context are irrelevant.

So if you want to provide some info by setting/changing props on context, you may want to skip this (especially if computations are expensivish) in case that context is not exposed since nobody will see the results and they are irrelevant.

daffl:
I think then we should add it in https://github.com/feathersjs/hooks as a Symbol property you can read via `context[RETURNED]` (or `EXPOSED` or `PUBLIC`). From a Feathers perspective currently, I believe what is set in `params.provider` indicates if e.g. things `context.http` should be set.

vonagam:
No real objection to it being a symbol property, thought it will be the only one and it is kinda strange, but ok.

There is nothing stopping people from passing/setting provider param in internal calls. There are use cases for that. So as a library developer I cannot be sure that presence of provider means that `dispatch`/`http` or other properties are worth computing.

Issue #1102, opened 2018-11-21
Multiple Rest Clients


tim-field:
Hi I'm trying to have my service A use remote services B and C should this be possible?  It works perfectly if I just have service B, but fails with `Only one default client provider can be configured` once I add service C.

I've tried various invocations of `use` and `configure`  without any luck 

```
const B = rest("http://localhost:3030")
const C = rest("http://localhost:3080")

const requestClient = request.defaults({})

const app = express(feathers())
  .use(express.json())
  .use(express.urlencoded({ extended: true }))
  .use(cors())
  .configure(B.request(requestClient))
  .configure(C.request(requestClient))
  .configure(express.rest())
  .configure(A)
  .use("/", (req, res) => {
    res.json({ service: "period-report", status: "running" })
  })
```

Thanks

tim-field:
Is there no way to do this natively ?

daffl:
The REST clients are just normal services themselves and can be registered directly. They require a `connection` (reference to the transport library), `name` and `base`:

```
const superagent = require('superagent');
const SuperAgentClient = require('@feathersjs/rest-client/lib/superagent');

app.use('/users', new SuperAgentClient({
  connection: superagent,
  name: 'users',
  base: 'http://myotherFeathersServer.com'
});
```

tim-field:
Thanks @daffl that worked.  This is quite different from the example here https://docs.feathersjs.com/api/client/rest.html#restbaseurl

So it seems one should only use `@feathersjs/rest-client` when a single rest service will be used?

daffl:
Going to reopen because I did want to make it a little easier to instantiate separate clients. Right now, only the main single client usage is documented. Not sure how the API would look like yet though.

```js
const server1 = rest('http://feathers-api1.com')
const server2 = rest('http://feathers-api2.com')

app.use('/users', server1.superagent('/users', superagent));
app.use('/messages', server2.superagent('/messages', superagent));
```

tim-field:
@daffl is the above an example of something that should work now, or a design goal?

But what I would like to know is if I can call a sub service of one of these registered services. 
```JS
.use(
    "/foo",
    new requestClient({
      connection: request,
      name: "foo",
      base: "foo-ap1.com"
    })
  )
```

Lets say my `foo` api ( which is now a service used in this api ) internally has a number of services  which I'd call like so `foo-ap1.com/users` if accessing directly.   How do I use the these services here ? 

I was hoping that in my app I could  `

```JS
[...]

PR #1776, opened 2020-01-03
allow qs options to be passed through

Mattchewone:
Allows for adding additional options for the qs.stringify method. I came to this as it's currently not possible at least from my trial and error, to be able to override the rest client so that `getQuery` can be extended so I can provide my own `qs` with options.

```js
const feathersClient = feathers()
  .configure(restClient.fetch(fetch, {
    qsStringifyOptions: {
      strictNullHandling: true
    }
  }));
```

### DIFF
--- a/packages/rest-client/lib/base.js
+++ b/packages/rest-client/lib/base.js
@@ -18,6 +18,7 @@ class Base {
     this.options = settings.options;
     this.connection = settings.connection;
     this.base = `${settings.base}/${this.name}`;
+    this.qsStringifyOptions = this.options.qsStringifyOptions || {};
   }
 
   makeUrl (query, id) {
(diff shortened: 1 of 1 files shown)


marshallswain:
Looks good to me.

daffl:
Thank you for the pull request @Mattchewone. I think what we'd really want to do here instead of just adding another option is being able to add a custom client service class with a customized `getQuery`. The API I'm thinking would look like this:

```js
const feathers = require('@feathersjs/feathers');
const rest = require('@feathersjs/rest-client');
​
const app = feathers();
​
class MyFetchClient extends rest.Fetch {
  getQuery () {
    return {};
  }
}
​
// Connect to a different URL
const restClient = rest('http://feathers-api.com');
​
// Configure an AJAX library (see below) with that client 
app.configure(restClient.fetch(window.fetch, MyFetchClient));
​
// Connect to the `http://feathers-api.com/messages` service
const messages = app.service('messages');
```

Also related to https://github.com/feathersjs/feathers/issues/1102

Mattchewone:
Closed due to newer version - https://github.com/feathersjs/feathers/pull/1780

PR #1815, opened 2020-02-18
Make TypeScript HookMap and HooksObject interfaces generic.

JorgenVatle:
Made the `HookMap` and `HooksObject` interfaces generic. Beyond adding options for one to specify types for the associated `HookContext`, this helps a lot with the reusability of these interfaces.


daffl:
Makes sense, thank you!

PR #1837, opened 2020-02-29
Allow more specific types to be specified for `create` parameter

deskoh:
This PR allow more specific types to be specified for `create` service method.

```ts
interface MessageResponse {
  text: string;
  createdAt: string;
  updatedAt: string;
}

type MessageRequest = Omit<MessageResponse, 'createdAt' | 'updatedAt'>

declare var MyService: Service<MessageResponse, MessageRequest>;
const x = MyService.create({
  // Error: Property 'text' is missing in type '{}'
});
```

### DIFF
--- a/packages/adapter-commons/src/service.ts
+++ b/packages/adapter-commons/src/service.ts
@@ -25,16 +25,16 @@ export interface ServiceOptions {
   filters: string[];
 }
 
-export interface InternalServiceMethods<T = any> {
+export interface InternalServiceMethods<T = any, TCreateReq = Partial<T>> {
     _find (params?: Params): Promise<T | T[] | Paginated<T>>;
     _get (id: Id, params?: Params): Promise<T>;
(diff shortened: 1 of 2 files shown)


daffl:
Thank you for those PRs. They look good but I think we'll need TSLint set up to be able to merge them and be sure it doesn't conflict with the actual adapter types since that has happened before.

daffl:
Side note: The tests are all passing now (needed to be merged with latest master)

deskoh:
@daffl Any guidance on how TSLint could be setup?

daffl:
I wouldn't expect anything special outside of the normal setup. Looks like TSLint has been deprecated in favour of ESLint so I'm not sure what the up to date process is.

daffl:
I moved the adapter modules into the https://github.com/feathersjs/databases repository as well as the TS version of `feathers-memory` (other core adapters to follow shortly) so any change will be tested on the actual implementation. If you could reopen these changes in https://github.com/feathersjs/databases, if they pass we can merge them.

PR #1838, opened 2020-03-01
fix: Improve Service typings for DB Common API

deskoh:
Similar to #1567, improve typings for `create`, `patch`, `remove` overloads for DB Common API.


daffl:
Makes sense. Getting this in to move the `adapter` repositories over into https://github.com/feathersjs/databases.

PR #1869, opened 2020-03-15
feat(authentication-oauth): callback for proxying raw request params into authentication service

mikekolganov:
**Problem**: I can't get user's IP address on registration and login via OAuth.

This PR was created after this issue: #1833

Also I have related PR to feathersjs/docs#1447

**Case**: I have app with Facebook OAuth login flow (JWT). I need to save users IP addresses when they're registered or logged in and expected to use hooks for authentication service. Middleware for throwing raw request data to Feathers (IP address in this case) is added to `middleware/index.js`, assigning new keys to `req.feathers` object. I can see this data in context.params of all hooks except authentication service hooks (`after: create()`). I noticed that params are limited here and not carrying any of custom params from req.feathers object in [authentication-oauth/src/express.ts](https://github.com/feathersjs/feathers/blob/master/packages/authentication-oauth/src/express.ts#L63)

**Example of use**

middleware:
```js
  app.use((req, res, next) => {
    if (! req.feathers) {
      req.feathers = {};
    }

    req.feathers.ip = req.connection.remoteAddress;
    req.feathers.userAgent = req.headers ['user-agent'];
    next();
  });
```

authentication config:
```js
  app.configure(expressOauth({
    serviceParamsCallback: req => {
      const { ip, userAgent } = req.feathers;

      return { ip, userAgent };
    },
  }));
```

### DIFF
--- a/packages/authentication-oauth/src/express.ts
+++ b/packages/authentication-oauth/src/express.ts
@@ -17,7 +17,7 @@ const debug = Debug('@feathersjs/authentication-oauth/express');
 
 export default (options: OauthSetupSettings) => {
   return (feathersApp: Application) => {
-    const { authService, linkStrategy } = options;
+    const { authService, linkStrategy, serviceParamsCallback } = options;
     const app = feathersApp as ExpressApplication;
     const config = app.get('grant');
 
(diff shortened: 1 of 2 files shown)


daffl:
Thank you for the pull request. I think supporting this makes sense but couldn't we still just use `req.feathers` from the middleware by changing https://github.com/feathersjs/feathers/blob/40b4358bb6abeadff8d4678df881428dcbbd36fb/packages/authentication-oauth/src/express.ts#L63 to

```js
const params = {
  ...req.feathers,
  authStrategies: [ name ],
  authentication: accessToken ? {
    strategy: linkStrategy,
    accessToken
  } : null,
  query,
  redirect
};
```

daffl:
I added a PR that implements my suggestion in #1886. I think it solves the same issue without adding an additional option.

PR #1886, opened 2020-03-23
fix(authentication-oauth): Allow req.feathers to be used in oAuth authentication requests

daffl:
This pull request allows to use `req.feathers` in oAuth strategies as mentioned in https://github.com/feathersjs/feathers/pull/1869#issuecomment-601221026.

- Closes https://github.com/feathersjs/docs/pull/1447
- Closes https://github.com/feathersjs/feathers/pull/1869

### DIFF
--- a/packages/authentication-oauth/src/express.ts
+++ b/packages/authentication-oauth/src/express.ts
@@ -61,6 +61,7 @@ export default (options: OauthSetupSettings) => {
       const service = app.defaultAuthentication(authService);
       const [ strategy ] = service.getStrategies(name) as OAuthStrategy[];
       const params = {
+        ...req.feathers,
         authStrategies: [ name ],
         authentication: accessToken ? {
           strategy: linkStrategy,
--- a/packages/authentication-oauth/test/express.test.ts
+++ b/packages/authentication-oauth/test/express.test.ts
@@ -27,6 +27,7 @@ describe('@feathersjs/authentication-oauth/express', () => {
 
     assert.ok(data.accessToken);
     assert.equal(data.user.testId, 'expressTest');
+    assert.equal(data.fromMiddleware, 'testing');
   });
 
   it('oauth/test/authenticate with redirect', async () => {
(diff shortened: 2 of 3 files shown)


mikekolganov:
Thank you! I'll close my PRs now.

daffl:
Thanks! Wanted to make sure that you agree it indeed addresses the same issue. Also, thank you for the report and pull request! Will go out with the next release shortly.

KidkArolis:
Re posting from Slack:

This means params.provider = ‘rest’ is set now, and before it wasn’t, this can affect logic such as looking up entities via findEntity. In my case, findEntity (to find user) stopped working, because the request appeared to be external (due to provider being set), and you’re not allowed to get users if you’re not authenticated.

I need to update my logic to exclude provider in the oauth flow. But wondering if this could affect other peeps.

Perhaps Feathers oauth strategy should be updated to exclude provider somewhere in the flow of authenticate -> findEntity? (since this isn’t an external request, this is oauth internally trying to resolve an entity).

daffl:
Ah darn, I didn't think of that but since there is a way to access the oAuth strategy directly through an external request (authenticate with an existing oAuth access token) it's probably a good thing to get the entities with protections in place.

KidkArolis:
Hm.. thanks for pointing that out, I might have to restrict that edge case further.

Thing is, the "protections" often depend on the user accessing the data, but in this case - there is no user.. unless we find the entity and then re-find it again with the user passed in, the way it's done in the jwt strategy, basically.

daffl:
Yeah, I'm starting to wonder if the non-JWT strategies should _only_ ever create an access token and not return the user. Maybe a longer lived refresh token that can only create new access JWTs which also returns the user information.

sebastien-prudhomme:
Got also a side effect with ```feathers-permission``` because of these lines:

```
    if (!entity) {
      debug(`hook.params.${entityName} does not exist. If you were expecting it to be defined check your hook order and your idField options in your auth config.`);

      if (params.provider) {
        throw new Forbidden('You do not have the correct permissions (invalid permission entity).');
      }

      return context;
    }
```

PR #1888, opened 2020-03-24
Add type keys to service pagination options.

JorgenVatle:
Just a simple tweak that defines the missing types for the [`pagination`](packages/adapter-commons/src/service.ts#L23) options of [`ServiceOptions`](packages/adapter-commons/src/service.ts#L19). Here I'm assuming that the only supported options are `max` and `default`. 

I was unable to find anything that would indicate that there are additional options. But I assume the available options could be tied to the database adapter in use? If the pagination object should accept any list of customizable options, perhaps we could consider refactoring to something like the following snippet;
```ts
export interface ServiceOptions {
  events: string[];
  multi: boolean|string[];
  id: string;
  paginate: {
    [key: string]: any;
    default?: number;
    max?: number;
  }
  whitelist: string[];
  filters: string[];
}
```

This would keep type-hinting for the two [documented pagination options](https://docs.feathersjs.com/api/databases/common.html#pagination) while allowing for use of additional options.

### DIFF
--- a/packages/adapter-commons/src/service.ts
+++ b/packages/adapter-commons/src/service.ts
@@ -20,7 +20,10 @@ export interface ServiceOptions {
   events: string[];
   multi: boolean|string[];
   id: string;
-  paginate: any;
+  paginate: {
+    default?: number;
+    max?: number;
+  }
   whitelist: string[];
   filters: string[];
 }


deskoh:
@JorgenVatle Looking at the following [line](https://github.com/feathersjs/databases/blob/master/packages/adapter-commons/src/filter-query.ts#L15) and the [docs](https://docs.feathersjs.com/api/databases/common.html#pagination), `paginate` can take the value of `false` as well to disable pagination?

JorgenVatle:
@deskoh This PR only addresses service registration options. `app.use('/example', new ServiceAdapter({ ... }))` The actual query API referenced in the following example provided by the docs you mentioned should remain unaffected by this PR. 👍
image

JorgenVatle:
@deskoh To follow up on the previous note, I do believe the query API already has the proper type definitions for the `paginate` param. See [@feathersjs/feathers/index.d.ts](https://github.com/feathersjs/feathers/blob/master/packages/feathers/index.d.ts#L36). 👍

The query API however, appears to lack type hints for the `$limit` and `$skip` options. Unless that's been changed with a recent PR. 😅

PR #1890, opened 2020-03-25
Replace long-timeout with long-settimeout

Kamahl19:
### Summary

The currently used package for long timeouts (https://www.npmjs.com/package/long-timeout) does not have types and has no tests. I have created a new package with the same functionality written purely in Typescript with full test coverage https://www.npmjs.com/package/long-settimeout

With this change we alro get rid of the `// @ts-ignore` which is generally a bad practice.

PR is failing because I probably didnt updated package.json properly. I dont use npm for few years nor with Lerna. If you guys are interested in merging this, could you give me some instructions how to properly replace the package? Thanks <3

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3,72 +3,143 @@
 	"requires": true,
 	"lockfileVersion": 1,
 	"dependencies": {
+		"@angular/common": {
+			"version": "9.0.7",
+			"resolved": "https://registry.npmjs.org/@angular/common/-/common-9.0.7.tgz",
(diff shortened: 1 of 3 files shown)


daffl:
Thank you for the pull request. The failing Feathers tests look like it is not working as expected at the moment. I think it's a good suggestion but if we change it, I would prefer it as code within Feathers modules (maybe `@feathersjs/commons`) for the following reasons:

- Remove an external dependency
- Easier to audit
- Easier Deno build (eventually)

daffl:
Closing since I haven't heard back in a while. A new PR is always be welcome.

PR #1901, opened 2020-04-07
docs: Fix simple typo, serivce -> service

timgates42:
There is a small typo in packages/express/lib/index.js.

Should read `service` rather than `serivce`.


daffl:
Thank you!

PR #1920, opened 2020-04-16
chore(typescript): Migrate @feathersjs/errors to modern classes and TypeScript

daffl:
This pull request uses modern ES classes for error types and removes undocumented initialization modes and workarounds for browser engines that are no longer supported.


packages/errors/package.json
```diff
@@ -3,8 +3,8 @@
 	"description": "Common error types for Feathers apps",
```
bertho-zero, line 3: Why does this package use tabs and spaces?

bertho-zero:
Cool, I think we should integrate the concept of `verror` in this package, it's really practical.

The problem with `verror` is that it uses heavy dependencies and is not compatible for browsers, but it is easily replaceable.

https://github.com/joyent/node-verror

packages/errors/package.json
daffl: I was wondering why it looked odd. Fixed in https://github.com/feathersjs/feathers/commit/cbd31c10c2c574de63d6ca5e55dbfb73a5fdd758

daffl:
Interesting, I hadn't seen that before. Hoping that this refactoring as a first step combined with the new hooks will also improve error messages (and stack traces).

PR #1929, opened 2020-04-24
[Dove] use @feathers/hooks and add async type

bertho-zero:
From https://github.com/feathersjs/feathers/pull/1798


daffl:
Great, thank you! I'll have a look to see how to update to the latest version and start core migration to TypeScript (which I think would make sense to do at the same time).

PR #1938, opened 2020-04-30
chore(docs): Add inline docs to service methods

JorgenVatle:
Yet another TypeScript type definition tweak here 😊

This pull request adds the hook-less service methods described in [the Feathers migration documentation](https://docs.feathersjs.com/guides/migrating.html#hook-less-service-methods) to the `Service` type, allowing for usage of these methods with proper type hinting.

Additionally; I've gone ahead and added inline documentation to all service methods in the form of JSDocs, allowing for nice quick-reference documentation for those running modern code editors and IDEs.

[image](https://globglo.gabgalab.io/x45h9Y3DWUci)

### DIFF
--- a/packages/feathers/index.d.ts
+++ b/packages/feathers/index.d.ts
@@ -299,6 +299,7 @@ declare namespace createApplication {
          * Retrieve all resources from this service, skipping any service-level hooks.
          *
          * @param params - Service call parameters {@link Params}
+         * @see {@link HookLessServiceMethods}
          * @see {@link https://docs.feathersjs.com/api/services.html#find-params|Feathers API Documentation: .find(params)}
          */
         _find (params?: Params): Promise<T | T[] | Paginated<T>>;
@@ -308,6 +309,7 @@ declare namespace createApplication {
          *
          * @param id - ID of the resource to locate
          * @param params - Service call parameters {@link Params}
+         * @see {@link HookLessServiceMethods}
          * @see {@link https://docs.feathersjs.com/api/services.html#get-id-params|Feathers API Documentation: .get(id, params)}
          */
         _get (id: Id, params?: Params): Promise<T>;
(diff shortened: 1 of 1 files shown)


daffl:
Thank you for the pull request but these are already defined in the `@feathersjs/adapter-commons` repository at https://github.com/feathersjs/databases/blob/master/packages/adapter-commons/src/service.ts and used by all database adapters. If you set the type properly to the adapter service class they will show up.

JorgenVatle:
Oh, I did spot these prior to opening the PR here. 😅 This PR adds the methods to the Service interface, making the type definition accessible to users who access their service through e.g. `app.service('...')`. From what I can tell, these methods are not normally accessible. The reason they don't throw a red underline/compile error is due to the `[key: string]: any` definition in the `ServiceMethods` interface.

daffl:
I am not sure why this is not showing up for you properly. If you generate an application with the `@featherjs/cli` it will set up the [correct service types automatically](https://github.com/feathersjs/feathers-chat-ts/blob/master/src/services/users/users.service.ts#L10).

For example things are showing up in the [feathers-chat-ts](https://github.com/feathersjs/feathers-chat-ts) for me just as expected:

Screen Shot 2020-04-30 at 8 18 42 AM

JorgenVatle:
Oh, I see! I was on a bit of an ancient version of Feathers when adjusting these types and had just assumed the `InternalServiceMethods` interface was unrelated to the application service function. Sorry for taking up your time with all that. 😅

I imagine the JSDocs could still be useful though - I'll move them over in case you feel it's something the types could benefit from. 👍 Otherwise, feel free to close the PR 👌

JorgenVatle:
@daffl Latest commit of this pull request should now just have the JSDoc additions to the ServiceOverloads and ServiceMethods interfaces. 😊 (b3225a9) The proposed `HooklessServiceMethods` interface have been omitted. 👍

PR #1941, opened 2020-05-01
Updated typings for projects with strictNullChecks

deskoh:
This updates the generated typings for `getRedirect` from:

```ts
getRedirect(data: AuthenticationResult | Error, params?: Params): Promise<string>;
```

to

```ts
getRedirect(data: AuthenticationResult | Error, params?: Params): Promise<string | null>;
```

TypeScript projects using `strictNullChecks` that overrides `getRedirect`  will not allow `null` to be returned using the original typings.


deskoh:
Simplified the PR by specifying the return type explicitly. Should be backwards compatible for projects without [`strictNullChecks`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#--strictnullchecks).

daffl:
Makes sense, thank you!

PR #1950, opened 2020-05-07
omit provider so it's a server call

ismail-codinglab:
### Summary

It should not act as a user and do a call to get the user as a server


daffl:
Thank you for the pull request but this response will be returned to the client so the provider can't be omitted. You can [customize the `getEntity` method](https://docs.feathersjs.com/api/authentication/local.html#customization) if you would like to return something else.

Issue #1970, opened 2020-05-27
Typing Proposal - Overload `find` Service Method


robbyphillips:
### Expected behavior
It would be nice if the return type was automatically inferred from adapter services based on `params.paginate`

```typescript
let users: User[] = await app.service('users').find({ paginate: false })
let paginated: Paginated<User> = await app.service('users').find()
```
### Actual behavior

Instead, the type system does not know which return type to use, so we need to cast it.

```typescript
let paginated = await app.service('users').find() as Paginated<User>
```

### Proposal

I think we can solve this with an overload.  Something like:

```typescript
interface ParamsNoPaginate extends Params {  // not a great name, I know..
  paginate: false
}

// service.d.ts
export declare class AdapterService<T = any> implements ServiceMethods<T> {
  // ...
  find(params: ParamsNoPaginate): Promise<T[]>
  find(params?: Params): Promise<Paginate<T>>
  // ...
}
```

I'm less comfortable with conditional types, but I suppose that would work as well and it would avoid having to extend `Params`.

GVanderLugt:
I'm new to TypeScript, so I'm not comfortable commenting on the implementation, but this would definitely be nice to have.

PR #1972, opened 2020-05-28
feat: add overload types for `find` service methods

robbyphillips:
### Summary

I thought this might be a better way to have a conversation about #1970.

The goal is to let typescript automatically infer the return type for `find` methods based on `params.paginate`.

I do not think it is dependent on any other PRs.

### Other Information

Adding these type definitions did have one unexpected side effect where TS no longer approved of line 106 in `packages/authentication-oauth/src/strategy.ts`, but it was such a little change that I included it in this PR.

I can think of two things that I would really like some feeback on:

1.  The main `ServiceMethods<T>` definition for `find` has the signature `find (params?: Params): Promise<T | T[] | Paginated<T>>;`, and the `Promise<T>` result is not included in the overloads

2. I am not sure if changes similar to the one in the `strategy.ts` file would be needed in some ecosystem libs

### DIFF
--- a/packages/authentication-oauth/src/strategy.ts
+++ b/packages/authentication-oauth/src/strategy.ts
@@ -103,7 +103,8 @@ export class OAuthStrategy extends AuthenticationBaseStrategy {
       ...params,
       query
     });
-    const [ entity = null ] = result.data ? result.data : result;
+
+    const [ entity = null ] = Array.isArray(result) ? result : result.data
 
     debug('findEntity returning', entity);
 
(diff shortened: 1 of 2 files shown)


robbyphillips:
Hey @daffl, just wondering if I should've done something different here.  Totally get that something like this might be very low priority -- it just seems like a pretty straightforward "quality of life" improvement to me.

daffl:
Sorry about that and thank you for the PR @robbyphillips. Been meaning to address this but I am also in the process of moving the adapters specific typings into https://github.com/feathersjs/databases and https://github.com/feathersjs/databases/tree/master/packages/adapter-commons

robbyphillips:
No worries at all, just happy to contribute if it's helpful.

And thanks for pointing out where new work is happening.  Is this part of a reorg with v5?

Issue #1961, opened 2020-05-18
Update Grant and grant-profile


daffl:
There have been breaking changes that need to be taken into consideration. Migration guide at https://github.com/simov/grant/blob/master/MIGRATION.md

daffl:
This has been done in v5.0.0-pre.1 and is now in beta testing. Latest prereleases can be installed via `@feathersjs/<packagename>@pre`, updating the version numbers of all `@feathersjs` packages to `^5.0.0-pre.1` or by running

```
npx ncu -u --newest
```

Upgrade instructions are continuously updated and can be found in the [v5 migration guide](https://dove.docs.feathersjs.com/guides/migrating.html).

PR #1989, opened 2020-06-09
chore(dependencies)!: update grant from 4.x to 5.x

wolfgangwalther:
This updates `@feathersjs/authentication-oauth` to use the new version of grant (5.x). (#1961)

Among other stuff, the new grant version offers the possibility to pass options to the underlying http client, which is used for the request for `profile_url`. In my project I need to add proxy settings to this request.

This should go together with a corresponding PR for the docs, as some things have changed with grant. This still needs to be done.

Unfortunately I was not able to test those changes in a real project. I tried installing the `@feathersjs/authentication-oauth` package from my own git repo instead of npm, but this didn't work out well, because of the monorepo structure.
Because the unit tests don't implement a full oauth flow, I guess this should still be tested somehow. Especially the "getting the profile" part, because that was not tested but mocked before already and has completely changed now.
Any ideas on that, how to proceed here?

### DIFF
--- a/packages/authentication-oauth/src/express.ts
+++ b/packages/authentication-oauth/src/express.ts
@@ -4,7 +4,6 @@ import Debug from 'debug';
 import session from 'express-session';
 import { Application } from '@feathersjs/feathers';
 import { AuthenticationResult } from '@feathersjs/authentication';
-import qs from 'querystring';
 import {
   Application as ExpressApplication,
   original as express
@@ -26,7 +25,7 @@ export default (options: OauthSetupSettings) => {
       return;
     }
 
-    const { path } = config.defaults;
+    const { prefix } = config.defaults;
     const expressSession = options.expressSession || session({
       secret: Math.random().toString(36).substring(7),
       saveUninitialized: true,
@@ -37,9 +36,8 @@ export default (options: OauthSetupSettings) => {
 
     authApp.use(expressSession);
 
-    authApp.get('/:name', (req, res) => {
+    authApp.get('/:name', (req, _res, next) => {
       const { feathers_token, redirect, ...query } = req.query;
-      const { name } = req.params as any;
 
       if (feathers_token) {
         debug(`Got feathers_token query parameter to link accounts`, feathers_token);
@@ -48,11 +46,7 @@ export default (options: OauthSetupSettings) => {
       req.session.redirect = redirect;
       req.session.query = query;
 
-      res.redirect(`${path}/connect/${name}?${qs.stringify(query as any)}`);
-    });
-
-    authApp.get('/:name/callback', (req: any, res: any) => {
-      res.redirect(`${path}/connect/${req.params.name}/callback?${qs.stringify(req.query)}`);
+      next()
     });
 
     authApp.get('/:name/authenticate', async (req, res, next) => {
@@ -111,6 +105,6 @@ export default (options: OauthSetupSettings) => {
     authApp.use(grantApp);
 
     app.set('grant', grantApp.config);
-    app.use(path, authApp);
+    app.use(prefix, authApp);
   };
 };
--- a/packages/authentication-oauth/src/index.ts
+++ b/packages/authentication-oauth/src/index.ts
@@ -28,7 +28,7 @@ export const setup = (options: OauthSetupSettings) => (app: Application) => {
   const { strategyNames } = service;
 
   // Set up all the defaults
-  const { path = '/oauth' } = oauth.defaults || {};
+  const { prefix = '/oauth' } = oauth.defaults || {};
   const port = app.get('port');
   let host = app.get('host');
   let protocol = 'https';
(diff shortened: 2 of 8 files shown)


packages/authentication-oauth/test/express.test.ts
```diff
@@ -40,12 +48,4 @@ describe('@feathersjs/authentication-oauth/express', () => {
       delete app.get('authentication').oauth.redirect;
     }
   });
-
-  it('oauth/test/authenticate with error', async () => {
```
wolfgangwalther, line 44: Not sure what to do with that test. The thrown error was only implemented in the mocked `getProfile` function. Not sure what that was supposed to test in the first place.

Since `getProfile` doesn't exist anymore, I just removed the test for now.

packages/authentication-oauth/src/express.ts
```diff
@@ -48,11 +46,7 @@ export default (options: OauthSetupSettings) => {
       req.session.redirect = redirect;
       req.session.query = query;
 
-      res.redirect(`${path}/connect/${name}?${qs.stringify(query as any)}`);
-    });
-
-    authApp.get('/:name/callback', (req: any, res: any) => {
-      res.redirect(`${path}/connect/${req.params.name}/callback?${qs.stringify(req.query)}`);
```
wolfgangwalther, line 55: The new grant version mounts at `/prefix` instead of `/prefix/connect` - so redirecting is not needed anymore, but the request can just be passed on to the next handler.

packages/authentication-oauth/src/index.ts
```diff
@@ -43,16 +43,16 @@ export const setup = (options: OauthSetupSettings) => (app: Application) => {
...
   const grant = merge({
     defaults: {
-      path,
-      host,
-      protocol,
-      transport: 'session'
+      prefix,
+      origin: `${protocol}://${host}`,
+      transport: 'session',
+      response: ['tokens', 'raw', 'profile']
```
wolfgangwalther, line 49: `['tokens', 'raw']` should be the default - not sure whether `tokens` would be enough?

`profile` is added, to make grant make the request to `profile_url`.

packages/authentication-oauth/src/strategy.ts
```diff
@@ -1,5 +1,4 @@
 // @ts-ignore
-import getProfile from 'grant-profile/lib/client';
```
wolfgangwalther, line 2: `grant-profile` is now merged into `grant` itself.

daffl:
Thank you for the pull request, this is great! Since these are definitely breaking changes, could you do me a favour and resubmit it against the v5 branch at https://github.com/feathersjs/feathers/tree/dove?

To test the changes in an application, in the Feathers repository you can do

```
cd packages/authentication-oauth
npm run compile
npm link
```

Then in your project

```
npm link @feathersjs/authentication-oauth
```

The same steps work for any other modules that changed.

As for getting the profile, I think the `getProfile` method should stay as it is but just return the `data.profile` that Grant includes now. The other problem is that I don't think this change allows to authenticate with an existing access token which is very important for usage in e.g. mobile applications.

One way to improve the tests would be to create a test oAuth application, get an access token via an HTTP request and run the test with that. It wouldn't test the browser flow though of course.

wolfgangwalther:
So I rebased on dove, force-pushed and changed the base branch for this PR - I guess that should do it in that regard?

Keeping `getProfile`: I guess I expected that and actually I agree ;)

> The other problem is that I don't think this change allows to authenticate with an existing access token which is very important for usage in e.g. mobile applications.

Can you elaborate? Whats "this change" here - the whole update of grant or just the "removing getProfile" part?

Thanks for the hint about `npm link`. That sounds useful, but won't be as straightforward in my case, I guess: My application runs in docker containers - for development as well. So "global node_modules" is different for my application and the feathers repo. But I'll figure something out with that.

nathanbrizzee:
Hi @daffl , any idea when this pull request might be available?  I have a desperate need behind our corporate ntlm firewall.  All web access is proxied so I can't get any of the OAuth grant to work unless I am outside our proxy server. The current 4.x branch works fine as long as I am not behind a proxy.  Otherwise, It simply times out.  I've tried the monkey patch suggestion by the author [https://github.com/simov/grant/issues/93](https://github.com/simov/grant/issues/93)  , but I can't get it to work either.  Any help would be greatly appreciated.

daffl:
Because of the breaking changes this pull request requires a major release of which upgrading Grant is only a minor part. There is no release date for v5 other than before the end of the year and there is a lot that still needs to be done. You are welcome to [book an office hour](https://aquil.io/office-hours) to see if there are ways to get things working for you before.

wolfgangwalther:
>As for getting the profile, I think the getProfile method should stay as it is but just return the data.profile that Grant includes now. The other problem is that I don't think this change allows to authenticate with an existing access token which is very important for usage in e.g. mobile applications.

So I brought back `getProfile`. Does that solve the "access token" problem you are referencing, @daffl, or do you have something else in mind here?

> One way to improve the tests would be to create a test oAuth application, get an access token via an HTTP request and run the test with that. It wouldn't test the browser flow though of course.

Will look into this.

wolfgangwalther:
The build is failing - but I have no idea how that error might be related to my changes. I don't know TypeScript well, though.

daffl:
Thank you again @wolfgangwalther, sorry it took so long! I got the build fixed and it is now merged with the `dove` branch. I will do some more testing and figure out how to bring back authenticating with existing oAuth access tokens and get it out with another v5.x pre-release.

PR #2000, opened 2020-06-26
Fix #1999

### LINKED ISSUE #1999: Hook context param entity has unexpected redacted credentials
https://github.com/feathersjs/feathers/blob/a12a1cf8c286b623817078e8b611223c173b04ee/packages/authentication/src/jwt.ts#L133

Should be:

```
const value = await this.getEntity(entityId, omit(params, 'provider'));
```

Shouldn't this omit the provider? I am redacting some private keys from the user record when provider is set and because of this, context.params.[entity].credentials is redacted even for internal use.

nsainaney:
This fixes #1999 - missed the prize for the 2000th issue by 1! 😀

Proposal:
The entity loaded into hook.params.[entity] should bypass any redaction hooks on [entity].find as this is an internal call. I recommend calling getEntity here without the provider as this is an internal call.

I would appreciate alternative solutions if this causes other issues.

### DIFF
--- a/packages/authentication/src/jwt.ts
+++ b/packages/authentication/src/jwt.ts
@@ -130,7 +130,7 @@ export class JWTStrategy extends AuthenticationBaseStrategy {
     }
 
     const entityId = await this.getEntityId(result, params);
-    const value = await this.getEntity(entityId, params);
+    const value = await this.getEntity(entityId, omit(params, 'provider'));
 
     return {
       ...result,


nsainaney:
I see there is a unit test to ensure precisely the opposite of what I am proposing. I guess the alternative is to create an inherited JWTStrategy and override `authenticate`. If so, you can go ahead and close this issue.

Thanks

daffl:
Yep, that's what I'd recommend as well. Thank you for the PR though! Always appreciated even when they are not merged.

PR #2009, opened 2020-07-10
fix(authentication): Include query params when authenticating via authenticate hook

### LINKED ISSUE #2007: Issue authenticating user via JWT when users service requires query params
### Copied from Slack Convo
the authenticate hook cannot function as is when the db requires query params to correctly lookup the user entity during authentication, which is a common case for hosted mongodb databases which require a shard-key (specifically when using Azure CosmosDB)

### Expected
Using `authenticate("JWT")` hook successfully authenticates user.

### Actual
Authenticate hook fails with db error, `No record found for id ***`.

flyboarder:
Supersedes #2008 fixes #2007

### DIFF
--- a/packages/authentication/src/hooks/authenticate.ts
+++ b/packages/authentication/src/hooks/authenticate.ts
@@ -46,7 +46,7 @@ export default (originalSettings: string | AuthenticateHookSettings, ...original
     }
 
     if (authentication) {
-      const authParams = omit(params, 'provider', 'authentication', 'query');
+      const authParams = omit(params, 'provider', 'authentication');
 
       debug('Authenticating with', authentication, strategies);
 


daffl:
Looks like nothing is breaking and I reviewed the dependent calls and there really appears no reason anymore for not including the query.

DevBayer:
It seems to me that something is failing in his logic, including the query parameter is breaking my code, I use the authenticate('jwt') hook as a header everywhere and is performing the query to the service by adding in the query a column that obviously does not exist in the users model because the main query is directed on another service that uses another object model.

jchamb:
@daffl @DevBayer Same thing for me. letting the query param through basically  broke every request in my app, as it was trying to attach extra fields into the user query

daffl:
Not sure how i missed that, that's exactly what I was worried about. Fixed in v4.5.6.

flyboarder:
@daffl Would you like to review #2008 as a possible alternative solution?

jchamb:
Do most of the adapters support pulling the model attributes like sequelize? It would be pretty easy to apply a whitelist of allowed fields by just pulling the models attributes.

flyboarder:
@jchamb that would assume the user service is backed by a database adapter, and not for example a proxy to another service.

jchamb:
@flyboarder very true. Was just thinking as an easy default whitelist for what is prob the more common/standard installs. I would say that, that would also be an easy config, but if your going to have to config something might as well just make be explicit like your #2008

PR #2013, opened 2020-07-13
Add JWT getEntityQuery

flyboarder:
Brings both local-strategy and jwt-strategy into alignment.

Added getEntityQuery to JWT strategy, allows extending JWT class to include additional query parameters in entity lookup.

Closes https://github.com/feathersjs/feathers/issues/2007

### DIFF
--- a/packages/authentication-local/src/strategy.ts
+++ b/packages/authentication-local/src/strategy.ts
@@ -45,7 +45,7 @@ export class LocalStrategy extends AuthenticationBaseStrategy {
   }
 
   async findEntity (username: string, params: Params) {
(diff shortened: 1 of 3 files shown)


daffl:
Alright, I updated it so that `getEntityQuery` goes back to the default behaviour since it looked like it was a regression for some people. Basically what you should be able to do now is [customize the strategy](https://docs.feathersjs.com/api/authentication/jwt.html#customization) and change [getEntityQuery](https://docs.feathersjs.com/api/authentication/jwt.html#getentityquery-params) to return the query object you need based on the original `params`.

Issue #1337, opened 2019-05-08
Add support for refresh tokens


marshallswain:
We currently allow getting a new token by posting a valid auth token to `<loginEndpoint>/refresh`.  Refresh tokens have a slightly different workflow as explained here:
https://auth0.com/learn/refresh-tokens

philoez98:
Is there any mindful way of using refresh tokens in feathers right now? Any plans of adding support for them?

bujji1:
Hello All ,

Looks like it is still not available ( or Am I missing some thing ) . Could you please let us know when it will be available ?  I see there is passport refresh token repository available . Any one tried this ?
https://github.com/fiznool/passport-oauth2-refresh

khanshakeeb:
Hello all,
Can you anyone help me to understand how token can be refresh for the google strategy? Because I will not have password in google login scenario?

Thanks

th0r:
> Another approach is to store refresh token inside accessToken's payload

So anyone who have **even one** of your **even expired** access tokens will be able to easily generate an endless count of a new access tokens or am I miss something?

Refresh tokens should be **securely** stored on the client because of exact same reason! 

> @deiucanta the good news is that we kept this feature in mind while we designed auth@1.0. I don't think it will be long before we get it in place and documented.

This was posted in 2016. Guys, do you still have plans to support this must-have feature?

daffl:
No. You can't do anything with an expired token. Also, refresh tokens are much more easily possible in [the v4 prerelease](https://blog.feathersjs.com/feathers-spring-2019-news-and-v4-beta-test-4c023c283e87). A cookbook entry for how to do it will be part of the final release.

th0r:
> Also, refresh tokens are much more easily possible in [the v4 prerelease](https://blog.feathersjs.com/feathers-spring-2019-news-and-v4-beta-test-4c023c283e87).

Is there an approximate release date? 

> A cookbook entry for how to do it will be part of the final release.

Until this guide is released, could you explain in a few words how it can be done in v4 prerelease?

th0r:
@daffl ping

ellisadigvom:
Is there an official way to do this yet? v4 is here and I don't see anything in the docs

MichaelErmer:
@daffl could you elaborate how this is achievable with 4.0 and without hacks to the authentication service?

sarkistlt:
[...]

PR #2026, opened 2020-07-24
Enable refresh-token server side support

jackywxd:
### Summary

Enable refresh-token server side support by extending the existing Authentication Service. As refresh-token is a stateful resource, it also depends on refresh-token entity service, similar as local strategy depends on user entity service.

Related issue: #1337 

Includes unit test cases


daffl:
Thank you very much for this pull request @jackywxd! Sorry it took so long to get back to it. I think this is a great start. My questions would be

- If it can be its own module, it is always easier to publish it separately instead of pulling it into the core
- Aren't there any changes that need to be made in the core modules (like the `authentication-client`)?

jackywxd:
@daffl David, thanks for your time! You are right, it can be on its own module and it is a much cleaner and easier solution. 

The current code change is only applied for server-side, not related to `authentication-client`. I have't got a chance to look at the client-side code yet. Ideally, the client-side integration should be included. 

Anyway, I will close this request for now, will see if I can make changes to both server and client and publish as separated modules.

Thanks again!

daffl:
Great, thank you. Please let me know if you need anything. Once it is ready we should definitely also do a writeup.

itoonx:
I'm looking for it 👍

bujji1:
@jackywxd!  I found only the TypeScript version of the "refresh token "  fork , do we have the Java script version also? Thanks for your help on this.

PR #2028, opened 2020-07-26
Some of the published packages include coverage

jasonk:
I'm guessing you probably didn't intend to publish the coverage data as
part of these packages.  This PR just adds `coverage/` to all of your
`.npmignore` files, but in general I would recommend using the files
field in package.json to specify what should be published, rather than
depending on your `.npmignore` to include everything that shouldn't be
published..

[files]: https://docs.npmjs.com/files/package.json#files

```
$ du -ksh node_modules/@feathers*/coverage
356K	authentication-local/coverage
612K	authentication/coverage
348K	express/coverage
560K	transport-commons/coverage
```

### DIFF
--- a/packages/client/.npmignore
+++ b/packages/client/.npmignore
@@ -1,4 +1,4 @@
 src/
 test/
 browser/
-!dist/
+!dist/coverage/
--- a/packages/feathers/.npmignore
+++ b/packages/feathers/.npmignore
@@ -1,1 +1,1 @@
-test/
+test/coverage/
--- a/packages/authentication-client/.npmignore
+++ b/packages/authentication-client/.npmignore
@@ -1,2 +1,3 @@
 test/
 tsconfig.json
+coverage/
--- a/packages/authentication-local/.npmignore
+++ b/packages/authentication-local/.npmignore
@@ -1,2 +1,3 @@
 test/
 tsconfig.json
+coverage/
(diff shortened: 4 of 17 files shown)


packages/feathers/.npmignore
```diff
@@ -1 +1 @@
-test/
\ No newline at end of file
+test/coverage/
```
bertho-zero, line 1: This folder does not exist and this modification will include the tests.

packages/client/.npmignore
```diff
@@ -1,4 +1,4 @@
 src/
 test/
 browser/
-!dist/
\ No newline at end of file
+!dist/coverage/
```
bertho-zero, line 4: I think it should be:

```suggestion
!dist/
.cache/
.mocha-puppeteer/
.nyc_output/
```

daffl:
Thank you! I keep forgetting that it always needs to go into `.npmignore` and `.gitignore`.

bertho-zero:
@daffl Have you seen my suggestions? some folders do not exist (https://unpkg.com/browse/@feathersjs/client@4.5.7/)

daffl:
I don't think I saw it but I made a follow-up in https://github.com/feathersjs/feathers/pull/2036 where I specify the files explicitly instead. This should also take care of the other junk that's showing in the folder you linked.

Issue #1895, opened 2020-04-02
Remove environment variable substitution in configuration


daffl:
An often recurring issue is changing the local authentication field to `username` in Windows systems in the configuration which needs to be escaped as per the [current conventions for @feathersjs/configuration](https://crow.docs.feathersjs.com/api/configuration.html#usage). This has also been causing difficult to debug problems with other environment variables.

I was briefly debating a different syntax to denote environment variables but after some more research, I am instead proposing a breaking change that removes this functionality and instead use the [environment variable functionality of node-config](https://github.com/lorenwest/node-config/wiki/Environment-Variables#custom-environment-variables).

KidkArolis:
👍 I'm already using the node-config approach and it works great. Could probably create a migration script/command for this?

daffl:
Possibly. Since we don't know the actual environment variables it'd probably mean copying `default.json` into `config/custom-environment-variables.json`. Maybe with the additional assumption that only `UPPERCASE` names are considered environment variables.

mkovel:
@daffl  Hey, 
have you considered moving out from node-config and use another package for feathers-config functionality?
https://github.com/HenrikJoreteg/getconfig
https://github.com/tusharmath/node-config-ts
It looks like they have similar functionality, but lack some of the node-config limitations

Could you please share your thoughts on why splitting a single config file into two files and further supporting two config files (node-config approach) better / more convenient than using a single config file with special syntax for env vars?

daffl:
I think node-config works quite well but as with anything in Feathers, `@feathersjs/configuration` is just a small plugin that you don't have to use if you don't want to and can easily replace with any of the alternatives (especially in the new version, `@featherjs/configuration` is basically 20 lines of code that call `app.set(name, value)` from whatever got loaded into `node-config`).

Personally I'm not a fan of environment variables anymore because you loose type information and they are difficult to validate and figure out what the application actually got passed. I moved all my apps to just pull in the environment specific JSON configuration as one single environment variable.

mkovel:
But where and how you manage secrets? 
Do you store them in the environment-specific JSON configuration?

daffl:
[...]

PR #2065, opened 2020-09-12
fix(configuration): Fix handling of config values that start with . or .. but are not relative paths

koehlerb:
### Summary

This PR fixes the problem where a config value that is not a relative path gets normalized
to an absolute path when it shouldn't be. For example, values like ".foo" and "..bar" are
not relative paths.

- [ x] Are there any open issues that are related to this?

Not that I am aware of.

- [x ] Is this PR dependent on PRs in other repos?

Not that I am aware of.

### DIFF
--- a/packages/configuration/src/index.ts
+++ b/packages/configuration/src/index.ts
@@ -25,7 +25,7 @@ export default function init () {
             if (process.env[value]) {
               value = process.env[value];
             }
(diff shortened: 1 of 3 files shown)


daffl:
Thank you @koehlerb! This makes sense and will go out with the next patch release.

As a side note: The next breaking version will change the configuration behaviour to be more in line with node-config (see #1895 for more information)

PR #2090, opened 2020-10-12
(TypeDef) Add user property to the Params.

shihabmridha:
### Summary
I am using TypeScript. When I tried to access `context.params.user` property from the service hook the IntelliSense did not work. The reason is that the type of `context.params` is `Params`. But, `Params` does not contain the `user` field in it.

### Solution
Add `user` property in the `Params` interface.

### Checklist

### DIFF
--- a/packages/feathers/index.d.ts
+++ b/packages/feathers/index.d.ts
@@ -42,7 +42,7 @@ declare namespace createApplication {
         provider?: string;
         route?: {[key: string]: string};
         headers?: {[key: string]: any};
-        user?: {[key: string]: string};
+        user?: {[key: string]: any};
 
         [key: string]: any; // (JL) not sure if we want this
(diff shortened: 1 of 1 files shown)


J3m5:
The user object could have a key with any value like Number or Boolean, shouldn't the type of key values be "any" ?

daffl:
Thank you!

philipdalen:
@shihabmridha, thanks for adding the user the the params type but after I updated to the latest version of feathers this change introduced some breaking changes. We usually would add the user object returned from the "getEntity" method from the authentication strategy directly to user, that means that the user object always actually have been of type object even through it wasn't enforced before this release. With this change we cannot add the user object to the user prop anymore.

Any reason why user shouldn't be of type object?

Thanks!

PR #2100, opened 2020-10-22
fix oauth session.destroy is undefined when use cookie-session package

fadiquader:
Hi @daffl 
Oauth in our project suddenly broke, and it turned out that we use `cookie-session` for OAuth like so 
```
const cookieSession = require('cookie-session');

  app.configure(expressOauth({
    expressSession: cookieSession({
      name: 'session',
      keys: ['grant'],
      maxAge: 60 * 1000
    })
  }));
```
that fixes OAuth when scaling the app Kubernetes. So now when trying to authenticate, it throws an error because `cookie-session` doesn't have `destroy`.
The bug is in `authentication-oauth` repo. 

Thank you.

### DIFF
--- a/packages/authentication-oauth/src/express.ts
+++ b/packages/authentication-oauth/src/express.ts
@@ -96,7 +96,7 @@ export default (options: OauthSetupSettings) => {
         };
 
         await new Promise((resolve, reject) =>
-          req.session.destroy(err => err ? reject(err) : resolve())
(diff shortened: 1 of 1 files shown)


daffl:
Thanks. I didn't know that it is not always available. Wouldn't the promise just never resolve with this change though?

fadiquader:
@daffl You're right! Updated the PR.
Thanks

daffl:
Cool, thanks again, will go out shortly.

PR #2106, opened 2020-10-23
docs(typescript): fix misleading parameter name

Gabrola:
For IDEs with autocompletion, this argument name is slightly misleading. `context` makes more sense and is in line with the rest of the hooks documentation.


daffl:
Makes sense, thank you!

PR #2110, opened 2020-10-28
fix: clean script in client package causing test to fail

bwgjoseph:
Current `clean` script in `client/package.json` is causing test to fail


daffl:
Great, thank you! Looks like that was the one issue for Windows users when cloning the repository and running the tests.

Issue #1742, opened 2019-12-16
How to make `context.app.service('service-name')` Typed?


ismail-codinglab:
I am running a feathers v4 app converted from v3 and am trying to make everything typed.

This is perfectly typed
![copyshot](https://user-images.githubusercontent.com/48755937/70871429-7caa0600-1f9f-11ea-9f31-223bd03b15a2.png)

But when I do it this way, it's not typed at all.

![copyshot](https://user-images.githubusercontent.com/48755937/70871443-98ada780-1f9f-11ea-84fc-747e3b84c9fc.png)

I did my research, could not find it in the docs, nor in known examples

Two questions:
- How can I make it typed?
- Is there an example that is fully typed?

daffl:
This is currently not possible other than casting the app into the `Application` from the declarations but will change in a future version.

Ismail-codinglab:
`but will change in a future version.`

Does this mean you have a possible solution in mind I can work with?
Or should I try myself to get something like that working?

Ismail-codinglab:
As a temporary workaround can you make the context.app.service accept a generic?
e.g.

```
context.app.service<AsdModel>('asd');
```

robbyphillips:
As a workaround, I have found that extending `HookContext` in the declarations file is sufficient.

```typescript
// declarations.d.ts

import { Application as ExpressFeathers } from '@feathersjs/express'
import { HookContext as FeathersHookContext } from '@feathersjs/feathers'

// A mapping of service names to types. Will be extended in service files.
export interface ServiceTypes {}
// The application instance type that will be used everywhere else
export type Application = ExpressFeathers<ServiceTypes>

export interface HookContext<T = any> extends FeathersHookContext<T> {
  app: Application
}
```

Then I import the `HookContext` from declarations instead of feathers.

ismail-codinglab:
Yes that's exactly what I did. And I have put in eslint/tslint the hookContext from feathers in the banned imports so developers can't by accident import the wrong hookccontext

bwgjoseph:
Hi @ismail-codinglab, are you able to share how you did the `eslint` rule to disable import of certain class/module/file?

latipun7:
> This is currently not possible other than casting the app into the `Application` from the declarations but will change in a future version.

It would be great if `app.get(name)` is typed too. #1952

chrisbag:
> As a workaround, I have found that extending `HookContext` in the declarations file is sufficient.
> 
> ```ts
[...]

PR #2121, opened 2020-11-11
Update index.d.ts

nvitaterna:
Allow application generic on hooks.

### Summary

This allows you to set up a `HooksObject` with a custom Application interface so you don't need to manually set the typings in each hook.

Before:
```ts
import { Application } from '../../declarations';

const hooks: HooksObject<Data> = {
  before: {
    all: [
      (context) => {
        // need to set the typing to get intellisense for your current application
        (context as { app: Application }).app.service('service').create...
      }
    ]
  }
}
```

After:
```ts
const hooks: HooksObject<Data, Application> = {
  before: {
    all: [
      (context) => {
        // application has services intellisense on the app.service() method
        context.app.service('service').create...
      }
    ]
  }
}
```

### Other Information

This only affects TypeScript users, just a minor change that doesn't have any breaking changes.

### DIFF
--- a/packages/feathers/index.d.ts
+++ b/packages/feathers/index.d.ts
@@ -130,7 +130,7 @@ declare namespace createApplication {
         event?: null;
     }
 
-    interface HookMap<T = any, S = any, A = any> {
+    interface HookMap<T = any, S = Service<T>, A = Application> {
         all: Hook<T, S, A> | Hook<T, S, A>[];
         find: Hook<T, S, A> | Hook<T, S, A>[];
         get: Hook<T, S, A> | Hook<T, S, A>[];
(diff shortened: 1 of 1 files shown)


bwgjoseph:
I guess https://github.com/feathersjs/generator-feathers/pull/542 or https://github.com/feathersjs/feathers/issues/1742 covers your issue

nvitaterna:
> I guess [feathersjs/generator-feathers#542](https://github.com/feathersjs/generator-feathers/pull/542) or #1742 covers your issue

[feathersjs/generator-feathers#542](https://github.com/feathersjs/generator-feathers/pull/542) still requires you to manually set the `context` type on every single hook function. This PR would let you set it on the `HooksObject` so the following functions are already typed.

nvitaterna:
Upon further usage, this looks like it will need some more changes to be fully compatible.

daffl:
Thank you for the pull request anyway. It is indeed difficult to introduce changes like this in a non-breaking way, which is why this is also being worked in in the upcoming major version (https://github.com/feathersjs/feathers/tree/dove) in https://github.com/feathersjs/feathers/tree/dove-typescript-refactor

PR #2125, opened 2020-11-13
RFC: add support for Service Request and Response type

bwgjoseph:
Hi,

This PR would likely be a breaking change, so I am actually hoping to make it into `dove` instead of `crow`. I have most of the description written in [this-repo](https://github.com/bwgjoseph/feathers-request-response) that I created to showcase, but I will put some of key info here.

---

For the longest time, I have been wanting to have some ways to allow me to define a different set of request/response interface since it can be different (and it is for most cases that I have encounter). There have been some chats about this in `slack` as well, and I thought I can start making some changes and see if this can be accepted, and I'm sure can be further improved as well.

> https://feathersjs.slack.com/archives/C52QSFK0A/p1594138177347400
> https://feathersjs.slack.com/archives/C3M4SBWT0/p1600014049010600

## Objective

Allow to declare a different set of interface for `Request` and `Response` which can also greatly improve dev-experience.

## Example

```js
// post.interface.ts
interface Base {
    readonly _id: string;
    createdAt: Date;
    updatedAt: Date;
}

/ Client facing interface
interface PostRequest {
    title: string;
    body: string;
    comments?: string[];
}

// Base fields generated at server side
interface ServerPostRequest extends PostRequest, Omit<Base, '_id'> {}

interface PostResponse extends Omit<PostRequest, 'comments'>, Base {}

// post.service.ts
declare module '../../declarations' {
  interface ServiceTypes {
    'post': Service<PostRequest, PostResponse> & Post;
  }
}

// if you choose not to have different request/response interface, you can still do this
// the response type will still be PostInterface
declare module '../../declarations' {
  interface ServiceTypes {
    'post': Service<PostInterface> & Post;
  }
}
```

## Result

infer-create-request-1
infer-create-request-2
infer-create-response-1
infer-create-response-2

---

For more details, and example, please refer to [feathers-request-response-demo](https://github.com/bwgjoseph/feathers-request-response)

If there is a better way to do this, I would like to have your feedback. This should not affect any actual functionality of the application, it all just providing better typing's for typescript developers.

Thanks!

### DIFF
--- a/packages/feathers/index.d.ts
+++ b/packages/feathers/index.d.ts
@@ -55,9 +55,9 @@ declare namespace createApplication {
     }
 
     // tslint:disable-next-line void-return
-    type Hook<T = any, S = Service<T>> = (context: HookContext<T, S>) => (Promise<HookContext<T, S> | void> | HookContext<T, S> | void);
+    type Hook<T = any, R = T, S = Service<T, R>> = (context: HookContext<T, R, S>) => (Promise<HookContext<T, R, S> | void> | HookContext<T, R, S> | void);
 
-    interface HookContext<T = any, S = Service<T>> {
+    interface HookContext<T = any, R = T, S = Service<T, R>> {
         /**
          * A read only property that contains the Feathers application object. This can be used to
          * retrieve other services (via context.app.service('name')) or configuration values.
@@ -103,7 +103,7 @@ declare namespace createApplication {
          *  - A before hook to skip the actual service method (database) call
          *  - An error hook to swallow the error and return a result instead
          */
-        result?: T;
+        result?: R;
         /**
          * A read only property and contains the service this hook currently runs on.
          */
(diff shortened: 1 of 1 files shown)


bwgjoseph:
@daffl would this be considered for `v5` seeing that the first pre-release is out. Thanks

daffl:
This is a good suggestion. Since there already was a major refactoring in the works (PR via #2255) I included it there so we'll close this PR. Thank you for putting this up!

bwgjoseph:
Thanks @daffl! 

Will try it out after you have release a alpha/beta for it. The only thing that I noticed (now, with a quick look) is that you had the request at the 2nd arg, and response at the first. `ServiceInterface<Response, Request = Response>`. I guess it make sense if majority of the usage is without a different request, then there is no need to define the 2nd arg. 👍

Issue #1976, opened 2020-06-02
Custom methods the Feathers way


daffl:
Custom methods are one of the most commonly requested features for Feathers. As the [FAQ points out](https://crow.docs.feathersjs.com/help/faq.html#how-do-i-create-custom-methods), it is possible to handle 90% of all use cases with the normal CRUD service interface and hooks for workflows. It does however still leave some edge cases that would be good if a service could expose additional methods externally. One reason for this limitation was that it is difficult to secure arbitrary methods properly. The new https://github.com/feathersjs/hooks general purpose hooks make this much easier.

This is why starting at v5, in addition to the existing service methods, it will be possible to register your own custom methods. A custom method has a name and __fixed__ parameters of `(data, params)` where `data` is the payload and `params` is the usual service method parameters (including things like provider, authenticated user, query etc.). Method information can be passed as `options` to `app.use` (`app.use(path, service, options)`). This will also allow to more easily disable existing service methods (or their events) externally.

## Example

```js
class MyMessageService {
  async create (data, params) {}

  async findOne (data, params) {}

  async resendNotification (data, params) {}
}

app.use('/messages', new MyMessageService(), {
  methods: [ 'create', 'findOne', 'resendNotification' ],
  events: [ 'my-custom-event' ]
})
```

The available options are:

- `methods`: A list of all publicly exposed methods. Will default to the standard service methods. When passed those will have to be added explicitly.
- `events`: The custom service events (previous in `service.events`)

## Calling via transports

- Via REST a custom method can be called using the POST method and the `X-Service-Method` header. `data` will be the body payload.
- Via websockets the method will be available as `socket.emit(methodName, 'messages', { username: 'dave' }, query)`
- Feathers does not have an introspection mechanism, so when using the Feathers client, custom methods will have to be defined again explicitly on the client
- In GraphQL the method will be exposed as a mutation under a unique name

J3m5:
I think that this is a great addition that will allow more flexibility for building APIs and help keeping the hooks lean and simple.

webafra:
hello .

thanks to public this option.

where route findOne in rest api ?
/message/:id ?

bertho-zero:
I think that a custom method must also be able to have as arguments `id, data, params` or `id, params`.
To determine if there is `id` or not, we could apply this mechanism also on `/messages/:id`.

---

[...]

PR #2134, opened 2020-11-21
chore: upgrade @feathersjs/hooks

bertho-zero:
Works with https://github.com/feathersjs/hooks/pull/67


daffl:
Awesome. Hopefully getting out a new pre-release this weekend!

daffl:
This has been published in v5.0.0-pre.1 for beta testing. I'm wondering if we should have another pass to use the `@feathersjs/hooks` functionality more directly (instead of e.g. keeping `service.__hooks` around). This might also play into some other changes like

- Removing `Uberproto` and instead using ES6 classes and `Object.create` to extend service (started looking into that yesterday but had to look into some other things to wrap up)
- Another TypeScript refactoring pass turning the Application from a mixin into a class and improving service, configuration (`app.get`) and hook context type-ability
- Allowing custom methods on all transports (discussion in https://github.com/feathersjs/feathers/issues/1976)

bertho-zero:
The mixin on services is only used to return the context if we have `true` as the last argument, we can remove it to rely only on `@feathersjs/hooks`.

However, I don't see how to remove Uberproto for the Express app merge in the Feathers app.

PR #2140, opened 2020-11-25
feat: Allow reAuthentication using specific strategy

deskoh:
Allow reAuthentication using different strategy.

This use case arises if one were to implement an alternate JWT strategy with existing built-in Feathers JWT strategy


daffl:
That makes sense. Will go out with the next patch release. We'll have to make sure to also update [the docs](https://docs.feathersjs.com/api/authentication/client.html#reauthenticate-force) accordingly when adding new parameters or flags.

PR #2163, opened 2020-12-17
refactor: remove useless methods

bertho-zero:
Express now uses `service.methods`.


bertho-zero:
https://runkit.com/embed/av00zbvpjq6e

I think it is better to do without `service.methods` and use `@feathersjs/hooks` directly to find the name of the desired parameters:

```js
getManager(wrappedSayHello).getParams()
```

daffl:
Thanks! I made a comment about that in https://github.com/feathersjs/feathers/pull/2134#issuecomment-747712751

Issue #2128, opened 2020-11-16
socket.io-client v 3.0.1 incompatible


camurai:
Ran into an issue with @feathersjs/socketio-client not working with the latest version of socket.io-client (3.0.1) It fails to establish the socket connection returning (bad handshake)

### Steps to reproduce
create a new feathersjs application that uses socketio

install these packages in your frontend:
@feathersjs/feathers@4.5.10
@feathersjs/socketio-client@4.5.10
socket.io-client@3.0.1

in your frontend set up the socket connection following the code from the @featherjs/socketio-client documentation.
```
const feathers = require('@feathersjs/feathers');
const socketio = require('@feathersjs/socketio-client');
const io = require('socket.io-client');

const socket = io('http://api.feathersjs.com');
const app = feathers();

// Set up Socket.io client with the socket
app.configure(socketio(socket));
```

### Expected behavior
Should successfully establish the connection.

### Actual behavior
gets a (bad handshake) return, and then proceeds to try again repeatedly with no success.

If I downgrade socket.io-client to version 2.3.0 it works without an issue.
I am not sure if this is a bug with @feathersjs/socketio-client, socket.io-client v3.0.1 or if there is just some additional configuration necessary to get socket.io-client v3.0.1 working, in which case the documentation should probably be updated.

### System configuration

**Module versions** (especially the part that's not working):
@feathersjs/feathers 4.5.10
@feathersjs/socketio-client 4.5.10
socket.io-client 3.0.1

**NodeJS version**:
12.18.0
**Operating System**:
Windows 10

**Browser Version**:
Chrome v86.0.4240.198

**React Native Version**:

**Module Loader**:

J3m5:
The socket.io library, client and server parts, has introduced breaking change in it's last major 3.0 update.
For now, we need to pin the socket.io dependencies to 2.x in order to avoid incompatibilities.

daffl:
Correct. You can **not** upgrade your `socket.io-client` dependencies to version 3. Because of some pretty significant breaking changes this upgrade can not happen until Feathers v5.

joacub:
what is the release date for Feathers v5 ?

daffl:
Due to the Pandemic, having lost most sponsors and with contributors needing to focus on their own lives and businesses first, there currently is no set release date. [Sponsorships are always welcome](https://github.com/sponsors/daffl) and with two or three more companies or individuals at the gold or platinum tier I'd be willing to commit to a more fixed date.

[...]

PR #2185, opened 2021-01-07
fix: Add fallback for breaking Socket.io change

daffl:
Related to https://github.com/feathersjs/feathers/issues/2128#issuecomment-755590395 this pull request adds the not-really recommended fallback because otherwise even reinstalling your packages will break clients from being able to reconnect.


J3m5:
Maybe we should make it possible to pass the origins params as an option, what do you think ?

J3m5:
Breaking change has just been reverted in [2.4.1 ](https://github.com/socketio/socket.io/releases/tag/2.4.1) one minute before my last comment ^^

daffl:
Nice, thanks for the update! I figured it's not too great if thousands of app with millions of users just break without warning or indicator what happened just by reinstalling dependencies. We can close this one then.

PR #1717, opened 2019-12-01
Return type narrowing for service multi CRUD.

deskoh:
This PR narrow the return type of service `multi` `create`, `patch` and `remove` methods.

Consider the following database service:

```ts
type Message = { text: string }

// Return type: Promise<Message>
service.create({ text: 'msg' });
// Return type: Promise<Message[]>
service.create([{ text: 'msg' }]);

// Return type: Promise<Message>
service.update(1, { text: 'msg' });
// Typing error remains (Database adapter does not support multi update)
service.update(null, { text: 'msg' });

// Return type: Promise<Message>
service.patch(1, { text: 'msg' });
// Return type: Promise<Message[]>
service.patch(null, { text: 'msg' });
```

PR #2203, opened 2021-01-27
fix(adapter-commons): Return missing overloads

vonagam:
Same as before, were lost in transition back somehow.

Related PRs: #1567, #1717, #1838, [#4](https://github.com/feathersjs-ecosystem/databases/pull/4)


daffl:
Great, thank you!

PR #2204, opened 2021-01-27
feat(adapter-commons): Added mongoDB like search in embedded objects

### LINKED ISSUE #2102: Enhancement proposal: Full support for MongoDB-like sorting criteria in sorter
I propose that `sorter` should be non-destructively and fully backwards compatibly enhanced to support the MongoDB sorting criteria of sub objects e.g. `sorter({name: 1, age: -1, "family.children.age": 1})`. This will make the `sorter` very useful to use as part of many applications and can eliminate many ad hoc sorting code blocks for in-memory sorts.

It can be implemented like this (without any optimization):
```js
exports function sorter ($sort: any) {
  function getVal (a: any, sortKeys: any) {
    let keys = sortKeys.map(key => key);
    let val = a;
    do {
      let key = keys.shift();
      val = val[key];
    } while (keys.length);

    return val;
  };

  const criteria = Object.keys($sort).map(key => {
    const direction = $sort[key];
    const keys = key.split('.');

    return { keys, direction };
  });

  return function (a: any, b: any) {
    let compare;

    for (const criterion of criteria) {
      compare = criterion.direction * exports.compare(getVal(a, criterion.keys), getVal(b, criterion.keys));

      if (compare !== 0) {
        return compare;
      }
    }

    return 0;
  };
}
```
Implemented like above, it will be backwards compatible (but a little bit slower - I think is is easy to optimize it but the prize to pay will be readability).

Of cause, if the proposal is implemented, it will be necessary to document it.

mhillerstrom:
Fixes #2102 and adds ability to use sorting criteria for embedded objects like:

```
import { sorter } from '@feathersjs/adapter-commons';
...
let sortEmbedded = sorter({"item.category": 1, "item.type": 1, amount: 1});
let result = data.sort(sortEmbedded);
```

### DIFF
--- a/packages/adapter-commons/src/sort.ts
+++ b/packages/adapter-commons/src/sort.ts
@@ -69,17 +69,36 @@ export function compare (a: any, b: any, compareStrings: any = exports.compareNS
 // An in-memory sorting function according to the
 // $sort special query parameter
 export function sorter ($sort: any) {
-  const criteria = Object.keys($sort).map(key => {
-    const direction = $sort[key];
+  let sortLevels = false; // True if $sort has tags with '.' i.e. '{a: 1, b: -1, "c.x.z": 1}'
 
-    return { key, direction };
-  });
+  const getVal = (a: any, sortKeys: any[]) => {
+    let keys = sortKeys.map(key => key);
+    let val = a;
+    do {
+      let key = keys.shift();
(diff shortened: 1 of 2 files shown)


daffl:
Apologies for the delay, this makes sense, thank you!

Issue #2228, opened 2021-02-19
Typescript Quick Start example has type errors


SephReed:
### Steps to reproduce

1. Npm install all the listed imports:
```
// these four, npm install each
import feathers from '@feathersjs/feathers';
import '@feathersjs/transport-commons';
import express from '@feathersjs/express';
import socketio from '@feathersjs/socketio';
```

2. Copy /paste the given code

### Expected behavior
There should be no type errors

### Actual behavior
Tons of type errors
<img width="496" alt="Screen Shot 2021-02-18 at 9 23 44 PM" src="https://user-images.githubusercontent.com/14002040/108457321-099dbe80-7238-11eb-9c97-ea18797828a0.png">

SephReed:
It's this same issue I ran into a while ago. (https://github.com/feathersjs/feathers/issues/2120)

### tsconfig.json MUST have `"allowSyntheticDefaultImports": true,`

This *really* should be put in big giant bold letters somewhere on the getting started page.  It literally breaks the entire thing in typescript if you don't have that one tsconfig setting.

SephReed:
https://github.com/feathersjs/docs/pull/1518

daffl:
This happens when you do not follow the [first TypeScript step](https://docs.feathersjs.com/guides/basics/starting.html#quick-start) in the guide, that specifies to use `tsc --init --target es2018`:

<img width="772" alt="Screen Shot 2021-02-18 at 8 36 00 PM" src="https://user-images.githubusercontent.com/338316/108458181-f505fa00-7228-11eb-98aa-152492c8bf74.png">

SephReed:
Ah.  I've done the setup for a typescript project so many times that I did all the "create an npm project and install typescript" steps in the normal way that works for every other project.

I see that you're correct in that following these steps exactly would work.  Nothing about the code snippet above implies the importance of "allowSyntheticDefaultImports" coming from the line `npm init --yes`.

I don't see any reason why such important configuration should be so obfuscated.

daffl:
I don't remember what difference the ES2018 setting makes (imho it should be the default) but I think it had something to do with how imports are handled (it also doesn't transpile async/await and things like that since it's available in all supported Node versions).

Something I've been hoping to get some help with for a while is to store the last language tab (JS/TS) you selected (or have a language choose) in localstorage and always select that language by default. I think that would make going through the guide in that regard a little more clear.

SephReed:
That would be a nice feature for sure. 

[...]

Issue #2147, opened 2020-12-06
Hook typing and find method parsing issue


soullivaneuh:
I use the `Hook` type when declaring my hook:

```typescript
type PushToken = {
  token: string;
  applicationSlug: string;
  createdAt?: Date;
}

const handleCreate: Hook<PushToken> = ({ data, service, result }) => {
  if (!data) {
    return;
  }
  const { token, applicationSlug } = data;

  return service
    .find({ query: { token }, paginate: false })
    .then((tokens) => tokens.length)
  ;
}
```

Works fine, except for the last `then` callback.

The `tokens.length` throw the following typescript error:

```
Property 'length' does not exist on type 'PushToken | PushToken[] | Paginated<PushToken>'.
  Property 'length' does not exist on type 'PushToken'.ts(2339)
```

This is caused by the following signature:

```
find (params?: Params): Promise<T | T[] | Paginated<T>>;
```

I may have only one instance but also an array or a Paginated instance.

So I tried to force the typing like this:

```typescript
.then((tokens: PushToken[]) => tokens.length)
```

But then give me now an error on the function typing itself:

```
Argument of type '(tokens: PushToken[]) => number' is not assignable to parameter of type '(value: PushToken | PushToken[] | Paginated<PushToken>) => number | PromiseLike<number>'.
  Types of parameters 'tokens' and 'value' are incompatible.
    Type 'PushToken | PushToken[] | Paginated<PushToken>' is not assignable to type 'PushToken[]'.
      Type 'PushToken' is missing the following properties from type 'PushToken[]': length, pop, push, concat, and 26 more.ts(2345)
```

The only way I have to get rid of this error is to set the `any` type to my `tokens` params, but I loose all the interest of typescript then.

Not sure if it's a bug, but I can't find any reliable example for that on the documentation.

soullivaneuh:
Got the same issue with `patch`:

```typescript
const handleCreate: Hook<PushTokensData, PushTokens> = (context) => {
  const { service, params, data } = context;
  if (!params.user) {
    throw new Error('This hook must be under authentication.');
  }
  if (!data) {
    return;
  }
  const { token } = data;
  data.userId = params?.user?._id;
  data.pingedAt = new Date();

  return service
    .find({ query: { token }, paginate: false })
    // @see https://github.com/feathersjs/feathers/issues/2147
    .then((resolvedTokens: any) => resolvedTokens.length ? resolvedTokens[0] : undefined)
[...]

Issue #2005, opened 2020-07-05
feathers service declaration


bwgjoseph:
Hi,

By default, when generating a new service (Post) using `feathers-cli` will result in this interface declaration.

```ts
// Add this service to the service type index
declare module '../../declarations' {
  interface ServiceTypes {
    'post': Post & ServiceAddons<any>;
  }
}
```

Doing so, will not give us the `overload` type that is defined.. so we lose the overload methods declaration..

Made a small tweak to include interface for it, to illustrate it better

```ts
interface PostInterface {
  title: string;
  message: string;
}

// Add this service to the service type index
declare module '../../declarations' {
  interface ServiceTypes {
    'post': Post & ServiceAddons<PostInterface>;
  }
}
```

![image](https://user-images.githubusercontent.com/37484538/86539046-51333c80-bf2c-11ea-962d-609d9bab52a2.png)

Notice that, it does not provide the `overload` methods by default as shown in the image above. 

Now, if we were to declare as such..

```ts
// Add this service to the service type index
declare module '../../declarations' {
  interface ServiceTypes {
    'post': Post & ServiceOverloads<PostInterface> & ServiceAddons<PostInterface>;
  }
}
```

![image](https://user-images.githubusercontent.com/37484538/86539075-83449e80-bf2c-11ea-8076-eaf77e386538.png)

IDE will be able to pick up the correct overload types and display as such (shown in the image above)

So I thought that the default, should as such instead of `Post & ServiceAddons<PostInterface>`

To add-on, I'm not sure if there's actually any difference between declaring

```ts
// this
declare module '../../declarations' {
  interface ServiceTypes {
    'post': Post & ServiceOverloads<PostInterface> & ServiceAddons<PostInterface>;
  }
}

// and this
declare module '../../declarations' {
  interface ServiceTypes {
    'post': Post & Service<PostInterface>;
  }
}
```

Since that `Service` is defined as such `type Service<T> = ServiceOverloads<T> & ServiceAddons<T> & ServiceMethods<T>;`

---

So my proposal, is to switch from using

`'post': Post & ServiceAddons<PostInterface>;`

to 

~~`'post': Post & Service<PostInterface>;`~~
`'post': Service<PostInterface> & Post;`

update: and because of the way it's defined, the order matters (due to overloading?, or intersecting types?), if define `Post & Service<PostInterface>`, it will not "see" the `ServiceOverloads<T>` first. Which will cause this to happen.

note: the interface is different but the idea remains.

[...]

Issue #1952, opened 2020-05-10
Feathers Typescript - Typed HookContext


EmirGluhbegovic:
### Steps to reproduce

There is an issue titled similar to mine, but it drifts to typing the sequelize models.

My goals: 
- Be able to do `app.get([typed options appear here])`
- Be able to do `app.service([typed options appear here])` -> this does not happen when using HookContext

In my attempt to solve this I tried the following:

in the declarations.d.ts file:

```
imports....
  import config from '@config/default.json';
 declarations...

// A mapping of service names to types. Will be extended in service files.
export interface EntityServices {
	users: Service<UserInterface>;
	subscribers: Service<SubscriberInterface>;
	permissions: Service<PermissionInterface>;
}

type Config = typeof config;

interface ModifiedConfig extends Config {
	sequelizeClient: Sequelize.Sequelize;
	'view engine': 'ejs';
}

export interface ServiceTypes extends EntityServices {}

type initialApplication = ExpressFeathers<ServiceTypes>;

type AppGet = <T extends keyof ModifiedConfig>(path: T) => ModifiedConfig[T];
type AppSet = <T extends keyof ModifiedConfig>(
	path: T,
	config: ModifiedConfig[T],
) => initialApplication;

// The application instance type that will be used everywhere else
export interface Application extends Omit<initialApplication, 'get' | 'set'> {
	get: AppGet;
	set: AppSet;
}

export type Paths = keyof ServiceTypes;
export type EntityPaths = keyof EntityServices;

export interface APIHookContext extends FeathersHookContext {
	app: Application;
	path: Paths;
}

export interface HookContext extends FeathersHookContext {}

```

This initially works perfectly but the issue comes in the use:

In a hook function if I change:

This:

`
export const sendNewUserEmail = async (context: HookContext) => {
	[some code...]
	return context;
};
`

To this:

`
export const sendNewUserEmail = async (context: APIHookContext) => {
	[some code...]
	return context;
};
`

I get all of typing BUT, all hell breaks loose with some of the following:

The `service.hooks(hooks);` call, now complains of a mismatch between the application and paths object of APIHookContext and HookContext. 

On deeper inspection I see within the feathers typing, HookContext refers to "this", which I believe is the uninitialized application. Hence my object is populated while "this" is empty.

I tried further by extracting the `service.hooks()` function and updating its type but then the issue arises with feathers imports such as "feathers-hooks-common" functions. They pass HookContext which then does not match with APIHookContext.

[...]

PR #2255, opened 2021-03-18
feat: Feathers v5 core refactoring and features

### LINKED ISSUE #509: Asynchronous initialization of Feathers services
This is real issue that I've recently had, but I guess this concerns the framework in whole. I think that it is natural for Feathers services to be asynchronously initialized. They can pick their configuration from asynchronous sources, or it can be something else.

In most cases synchronous application initialization is presumed, but I've encountered design problems when started using Sequelize.

While Mongoose does some good job on accumulating promises inside for initialization actions (connection, index creation), so queries are automatically chained to initial promises, Sequelize does nothing like that, e.g. when `sync()` is called.

I would expect that SQL database is ready at the moment when web server has been started, but when the application is defined in synchronous manner, it will not. And this is what happens with typical Feathers setup. In some cases async initialization may take several ms, in other cases it may take seconds, while the server will pollute logs with errors that should have been never happen normally.

Another problem is error handling. If there are uncaught errors during async service initialization, the reasonable move would be to halt app initialization and catch errors in a single place.

Here's a small example that explains the idea and is based on the structure proposed by Feathers generator template.

app.js:

```js
const app = require('feathers')();

// THIS!
app.set('feathersServicePromises', []);

app
.use(require('body-parser').json())
.configure(require('feathers-hooks'))
.configure(require('feathers-rest'))
.configure(require('./middleware'))
.configure(require('./services'));

// THIS!
Promise.all(app.get('feathersServicePromises'))
.then(() => app.listen(3000), console.error);
```

services/index.js:

```js
module.exports = () => {
    const app = this;

    app
    .configure(require('./foo-async-service')
    .configure(require('./bar-async-service')
    .configure(require('./baz-sync-service')
};
```

services/foo-async-service.js:

```js
const sequelize = require('./common/sequelize');

module.exports = () => {
    const app = this;
    const fooModel = sequelize.model('foo');

    // AND THIS!
    const initPromise = fooModel.sync();
    app.get('feathersServicePromises').push(initPromise);

    app.use('/foo', require('feathers-sequelize')({ Model: fooModel  }));
    app.service('/foo');
};
```

The places of interest here are `initPromise` and `Promise.all(...)`. It would be convenient to just return `initPromise` from service function, like it is done in hooks. And `Promise.all(...)` could probably be lazily executed with `app` object `toPromise()` method or `promise` getter property. So it could be

```js
module.exports = () => {
    ...
    return fooModel.sync();
};
```

in service function, and

```js
app
...
.configure(require('./services'))
.promise
.then(() => app.listen(3000), console.error);
```

in entry point.

I suppose that the pattern I've used for async initialization is already simple and clean, but it would be great if the framework would take this job.

My suggestion is to place this feature under consideration. 

I would appreciate any thoughts on the subject.

### LINKED ISSUE #853: app.setup() should be async
I have feathers app that uses sequelize for storage. For the tests, I use sqlite database, which is removed before the test starts, and is created  in sequelize.js file as generated by feathers generate service. This all worked fine until the number of models increased beyond 10 ( the number is dependent on how fast the machine is).  Now every time I run the full test suite the first couple of tests fail because the DB sync hasn't completed yet and it happens so that the first tests that get ran are the last tables being created. 

As a simple workaround , I changed the before of app.test.js(it's the first test being executed)  like so :

```diff
--- a/core/test/app.test.js
+++ b/core/test/app.test.js
@@ -18,7 +18,7 @@ const getUrl = pathname => url.format({
 describe('Feathers application tests', () => {
   before(function(done) {
     this.server = app.listen(port);
-    this.server.once('listening', () => done());
+    this.server.once('listening', () => setTimeout( ()=> done(), 1000));
   });
```

This fixes it for now until the number of models grows of the test is executed on a slower machine.
The real problem is that app.listen() calls  app.setup(), but doesn't wait for it to complete before firing the listening event. The database is created in app.setup and can take some time.

I think app.setup() should be async and app.listen() should wait for it to complete, or else there will always be possibility for race conditions between setup()  and service requiring the setup() to complete before it's functional.

### LINKED ISSUE #932: Koa style middleware: Hooks for any asynchronous method

### LINKED ISSUE #1160: FeathersError stripped of dynamic properties in Winston logger
### Steps to reproduce

I believe this is related to #653 that was closed by #697. I'm seeing similar results on v3.3.5.

In logger.js:
```js
const { createLogger, format, transports } = require('winston');
const { combine, timestamp, label } = format;

const env = process.env.NODE_ENV || 'development';
const options = {
  host: 'localhost',
  port: env === 'development' ? 3061 : 3031,
  path: 'logs',
  auth: 'None',
  ssl: false
};

// Configure the Winston logger. For the complete documentation see https://github.com/winstonjs/winston
const logger = createLogger({
  // To see more detailed errors, change this to 'debug'
  level: env === 'development' ? 'debug' : 'info',
  format: combine(
    label({ label: 'connector'}),
    timestamp(),
    format.splat(),
    format.simple()
  ),
  transports: [
    new transports.Console(),
    new transports.Http(options),
  ],
});

module.exports = logger;
```
In app.js:
```js
const logger = require('./logger');
.
.
.
app.use(express.errorHandler({ logger }));
```

### Expected behavior
When I manually use the logger, this is the JSON received by the HTTP server:
```js
{ level: 'info',
    message: 'TEST Feathers application started on http://localhost:3060',
    label: 'connector',
    timestamp: '2019-01-06T22:24:08.403Z' },
  { query: {}, route: {}, provider: 'rest' } 
```
Notice the level, label, and timestamp that were added by winston.

### Actual behavior
When running the predefined Feathers unit tests for 404, an internal FeathersError is thrown. This is what is received by the HTTP server:
```js
{ name: 'NotFound',
    message: 'Page not found',
    code: 404,
    className: 'not-found',
    data: { url: '/path/to/nowhere' },
    errors: {} },
  { query: {}, route: {}, provider: 'rest' }
```
The level, label, and timestamp have been stripped off. Debugging the http transport in Winston, at this line `req.end(Buffer.from(JSON.stringify(options), 'utf8'));`, `options` is the full FeathersError object and has the level, label, and timestamp properties. But I believe the FeathersError.prototype.toJson() function is not adding the dynamic properties, as mentioned in #653, and thus they get stripped off as they are logged. Is this what #697 was supposed to resolve?

### System configuration

**Module versions** (especially the part that's not working): feathers-errors@3.3.5, winston@3.1.0

**NodeJS version**: 8.9.4

**Operating System**: Mac

### LINKED ISSUE #1262: Error format
### Steps to reproduce

We are using @feathersjs/feathers version 3.1.3 with @feathersjs/errors version 3.3.0. 

We configure authentication service with some hooks:
```ts
app.service('authentication').hooks(authHooks(app));

function authHooks(app) {
  const config = app.get('authentication');
  return {
    before: {
      create: [validateUserAccess, authentication.hooks.authenticate(config.strategies)],
      remove: [authentication.hooks.authenticate('jwt')],
    },
  };
}
```
When we throw a Forbidden error object inside the validateUserAccess hook then we get a json object back (called by rest&ws). When we throw a Forbidden error object inside service hook we get html back without error code (rest&ws). Why is that? 

We would like to get the same json output for hooks before service calls and hooks for authentication.

### LINKED ISSUE #1274: Errors result in GeneralError without specifics of the cause
### Steps to reproduce

I might be doing something wrong, but in a lot of cases when errors are thrown and converted to a Feathers GeneralError the details of the actual error that was thrown are missing. This is very frustrating when it is an error in say production, as you have no clue what is causing the error.

What I have noticed in my own code: when an error is thrown without the `new` keyword (throw Error versus throw new Error) the Feathers GeneralError does not contain the error details. So in my own code I have changed this everywhere to use the `new` keyword. However I do not have control over external modules I am using.

### Expected behavior
I would expect errors to be caught by Feathers, converted to a General Error and for the original error details to be added to it.

### Actual behavior
The error is caught and converted to a Feathers GeneralError but without the original error details.

### System configuration

@feathersjs/feathers: 3.1.7
@feathersjs/errors: 3.3.0

**NodeJS version**: 10.9.0

**Operating System**: Windows 10 and Linuix 18.1

**Browser Version**: n/a

**React Native Version**: n/a

**Module Loader**: n/a

### LINKED ISSUE #1548: [FR] Support reverse application hooks
### Steps to reproduce
If you need before hooks to run after all services, you'd have to copy paste the hook to every single service, or hack it into an extended adapter.

### Solutions:
1. add applications hooks: afterBefore, beforeAfter, beforeError that can handle the reverse scenario.
2. make the cli generate apps with a hook filter on the service that can facility support for these cases. Hook middleware in a way.

### My current solution: 
Usage in service: service.hooks(proxyHooks(hooks))
```
// Application hooks that run for every service

const proxiedHooks = {
  before: {
    all: [],
    find: [],
    get: [],
    create: [ ],
    update: [ ],
    patch: [ ],
    remove: []
  },

  after: {
    all: [ ],
    find: [],
    get: [],
    create: [],
    update: [],
    patch: [],
    remove: []
  },

  error: {
    all: [],
    find: [],
    get: [],
    create: [],
    update: [],
    patch: [],
    remove: []
  }
}

modules.exports = function proxyHooks (hooks) {
  for (const baeK of Object.keys(hooks)) {
    const cruds = hooks[baeK]
    for (const crudK of Object.keys(cruds)) {
      const hook = cruds[crudK]
      if (Array.isArray(hook)) {
        cruds[crudK] = baeK === 'before'
          ? [...hook, ...proxiedHooks[baeK][crudK]]
          : [...proxiedHooks[baeK][crudK], ...hook]
      }
    }
  }
  return hooks
}
```

Conversation that inspired this: https://pastebin.com/fQh43HhT

### LINKED ISSUE #1723: Running a global hook after service hooks
**Is there a way to run a global before hook, for example `setField({ from: 'params.user.id', as: 'data.createdById' })`, after the service level before hooks have run?**

The services should in this case manage authentication and decide which authentication strategies that are allowed so the `authenticate` hook cannot be global and the global hook needs to be run after authentication.

I'm open for any workarounds, but I'd rather not add the `createdById` hook to all existing and new services.

### LINKED ISSUE #1736: Sequelize and upgrading to feathers V4 with TypeScript
### Steps to reproduce

First of all, thank you for the amazing feathers V4 upgrade that can support typescript!

### The problem

Upgrading to feathers.js and using the suggested feathers sequelize migrations setup

which suggests having a migrations folder next to the src folder)

Only way I could prove it was to look into the feathers sequelize docs (screenshot below)
image

This can give 2 problems:

#### 1. There are javascript files which are not under the rootdir

> error TS6059: File '**/migrations/scripts/2019****.js' is not under 'rootDir' '**/src'. 'rootDir' is expected to contain all source files.

How I solved it? Add it to the 'exclude' in tsconfig.ts

#### 2. The migrations script depends on ***/src/app.ts a raw typescript file that is not transpiled yet.
The code boils down to this. 
copyshot

So it will require the ts file, but it's the raw ts version which is not runnable by node itself

Now if you want to solve this you either have to:

A. Go back to javascript (not desired)
B. Transpile migration typescript files and run `sequelize db:migrate` on the transpiled files
C. Use a third party sequelize-cli that does support typescript 

### System configuration

**NodeJS version**: 12

**Operating System**: Linux

### LINKED ISSUE #1738: Feathers v4 upgrade declarations file missing
#### What is the problem?

When I try to upgrade feathers v3 to feathers v4 it creates a new file named `authentication.ts`
in this file it requires the following file:
```
import { Application } from '../../declarations';
```
Which does not exist
So i tried a clean install of feathers v4
In there, it has a `declarations.d.ts` file, which is missing from the v3 to v4 application.

### How did I solve it?
I copied `declarations.d.ts` file from an new feathers v4 application to an old feathers v3 moved to v4 application.
I changed
```
import { Application } from '../../declarations';
```
Into
```
import { Application } from './declarations';

declare module './declarations' {
  interface ServiceTypes {
    'authentication': AuthenticationService & ServiceAddons<any>;
  }
}
```

daffl:
I'm not a huge fan of single massive PRs especially if they will break things but it turned out that there wasn't really a way around it since some of the changes were pretty substantial and overarching. Overall documented API usage is mostly backwards compatible. Breaking changes are being documented [in the v5 migration guide](https://dove.docs.feathersjs.com/guides/migrating.html#deprecations-and-breaking-changes).

This pull request addresses several long standing feature request and issues broken down as follows:

## Asynchronous application initialization

`app.setup` and `app.listen` are now asynchronous. This allows to initialize databases, connections and can be combined with the new [@feathersjs/hooks](https://github.com/feathersjs/hooks) for more complex workflows.

- Closes https://github.com/feathersjs/feathers/issues/509
- Closes https://github.com/feathersjs/feathers/issues/853

## Custom methods

This lays the groundwork for custom methods laid out in https://github.com/feathersjs/feathers/issues/1976 but still needs tests.

## Asynchronous middleware with `@feathersjs/hooks`

This pull request builds on top of @bertho-zero's work and should finalize the migration to the new asynchronous `@featherjs/hooks`. It contains a backwards compatible wrappers for `before`, `after` and `error` service and application hooks but makes asynchronous hooks a first class citizen.

- Closes https://github.com/feathersjs/feathers/issues/509
- Closes https://github.com/feathersjs/feathers/issues/932
- Closes https://github.com/feathersjs/feathers/issues/1548
- Closes https://github.com/feathersjs/feathers/issues/1723

This also allows improved error handling, including better stack traces.

- Closes https://github.com/feathersjs/feathers/issues/1160
- Closes https://github.com/feathersjs/feathers/issues/1262
- Closes https://github.com/feathersjs/feathers/issues/1274

## TypeScript improvements

Also included are major TypeScript improvements which should also allow better usability and code completion in editors like VSCode for non-TypeScript users.

- Typed `app.get` and `app.set`
- Improved application service declaration
- Fully typed hook context including proper application and service type
- Individual service request and response types

Addresses the following issues:

- Closes https://github.com/feathersjs/feathers/pull/2125
- Closes https://github.com/feathersjs/feathers/issues/1738
- Closes https://github.com/feathersjs/feathers/issues/1736
- Closes https://github.com/feathersjs/feathers/issues/1741
- Closes https://github.com/feathersjs/feathers/issues/1742
- Closes https://github.com/feathersjs/feathers/issues/1952
- Closes https://github.com/feathersjs/feathers/issues/2005
- Closes https://github.com/feathersjs/feathers/issues/2147
- Closes https://github.com/feathersjs/feathers/issues/2228


ismail-codinglab:
Damn that's one hell of a typescript improvement, no more wild guesses or lots of `any` workaround. 👍 🥳 

I don't mind this being a big pr as long as it's covered by tests, which it look like it is.

Do you know when or in which version/tag this will be available?

Again thanks for all that you do to make our lives easier

PR #1566, opened 2019-09-19
feat: application service types default to any

vonagam:
In #1364 `ServiceTypes` default changed from `any` to `{}` with an intention to make `app.service()` act more strict when `ServiceTypes` is provided. Desired outcome and expected type resolution (specified in second comment from that PR) were right, but the solution had a flaw.

When you make `ServiceTypes` default to `{}`, `Application` without parameters (like in [`HookContext`](https://github.com/feathersjs/feathers/blob/13b3dbdf8133bce2833a56bfbd0f0799b0420a8a/packages/feathers/index.d.ts#L59)) becomes `Application< {} >`. And you can't assign `Application< {} >` to `Application< MyServiceTypes >`:

```ts
const app: Application< MyServiceTypes > = context.app; // error
const app: Application< MyServiceTypes > = context.app as Application< any >; // works
```

As shown `Application< any >` does not have such drawback.

To achieve desired outcome right solution is to simply remove second catch-all `service()` overload, because when `ServiceTypes` is `any` first definition is already catch-all and when `ServiceTypes` is not `any` then first definition acts as a strict one.

There is a little breaking change (that why it is not a fix, but a feat) since now `{}` stops acting as wildcard and not a default value.

PR #2260, opened 2021-03-21
fix: Resolve some type problems

vonagam:
Continues #1566.

Don't know why types started to fail only after my changes, but i found and fixed reason for failures:

In [`LegacyHookFunction`](https://github.com/feathersjs/feathers/blob/2fb4cfa19b35f54ed5ff372233aeb50ca6c65dfe/packages/feathers/src/declarations.ts#L311) `app` property on potentially returned context was required to be of the same type as on passed in context. And because of this it was creating variance problems between simple feathers app and express wrapped app, through such forbidden chain:

```ts
(expressApp as FeathersApp).service('some').hooks({after: (context) => ({app: feathersApp, ...otherStuff})});
```

Since we don't care about returned context, simply relaxed type there. In turn such relaxation allowed to improve type of options in `hooks` method of app itself (and not on service through which the problem was exposed).

Other bonus small changes:

- Updated `ExpressUseHandler` as you mentioned. I missed that one, but it was not a cause for the failings. Also created separate overload for when `path` is a regexp.
 
- Keep `ServiceTypes` and `AppSettings` from feathers app during express wrapping.

 - In `app.use` `| Application` should be inside ternary expression, not outside.

### DIFF
--- a/packages/express/src/declarations.ts
+++ b/packages/express/src/declarations.ts
@@ -6,13 +6,14 @@ import {
 } from '@feathersjs/feathers';
 
 interface ExpressUseHandler<T, ServiceTypes> {
-  <L extends keyof ServiceTypes> (
-    path: ServiceTypes[L] extends never ? string|RegExp : L,
+  <L extends keyof ServiceTypes & string> (
+    path: L,
     ...middlewareOrService: (
       Express|express.RequestHandler|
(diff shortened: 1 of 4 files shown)


daffl:
Perfect, thank you! I was actually wondering why it was working before, too when I was looking into it. This looks great.

PR #2270, opened 2021-03-27
feat(core): Public custom service methods

### LINKED ISSUE #1976: Custom methods the Feathers way

daffl:
This pull request implements tests and client side functionality for custom service methods outlined in and closes #1976.

On the server a custom service method is added like this:

```js
class MyMessageService {
  async create (data, params) {}

  async findOne (data, params) {}

  async resendNotification (data, params) {}
}

app.use('/messages', new MyMessageService(), {
  methods: [ 'create', 'findOne', 'resendNotification' ]
});
```

It can be used via an HTTP and socket API directly as described in #1976.

On the Feathers client it is used like this:

```js
const feathers = require('@feathersjs/feathers');
const socketio = require('@feathersjs/socketio-client');
const io = require('socket.io-client');

const socket = io('http://api.feathersjs.com');
const app = feathers();

// Set up Socket.io client with the socket
app.configure(socketio(socket));

// Since there is no introspection, custom methods need to be initialized on a service
app.service('messages').methods('customMethod', 'findOne');

const message = await app.service('messages').findOne({ userId: 1 });
```

TypeScript has a CustomMethod and client interface:

```ts
import { feathers, CustomMethod } from '@feathersjs/feathers';
import socketio, { SocketClient } from '@feathersjs/socketio-client';
import io from 'socket.io-client';

type ServiceTypes = {
  messages: SocketClient & CustomMethod<'customMethod'|'findOne'>
}

const socket = io('http://api.feathersjs.com');
const app = feathers<ServiceTypes>();

// Set up Socket.io client with the socket
app.configure(socketio(socket));

// Since there is no introspection, custom methods need to be initialized on a service
app.service('messages').methods('customMethod', 'findOne');

const message = await app.service('messages').findOne({ userId: 1 });
```

### DIFF
--- a/packages/express/src/declarations.ts
+++ b/packages/express/src/declarations.ts
@@ -33,7 +33,7 @@ export type Application<ServiceTypes = any, AppSettings = any> =
 
 declare module '@feathersjs/feathers/lib/declarations' {
   export interface ServiceOptions {
-    middleware: {
+    middleware?: {
       before: express.RequestHandler[],
       after: express.RequestHandler[]
     }
--- a/packages/express/src/index.ts
+++ b/packages/express/src/index.ts
@@ -35,12 +35,15 @@ export default function feathersExpress<S = any, C = any> (feathersApp?: Feather
   const mixin: any = {
     use (location: string, ...rest: any[]) {
       let service: any;
-      const middleware = Array.from(arguments).slice(1)
-        .reduce(function (middleware, arg) {
+      let options = {};
+
+      const middleware = rest.reduce(function (middleware, arg) {
           if (typeof arg === 'function' || Array.isArray(arg)) {
             middleware[service ? 'after' : 'before'].push(arg);
           } else if (!service) {
             service = arg;
+          } else if (arg.methods || arg.events) {
+            options = arg;
           } else {
             throw new Error('Invalid options passed to app.use');
           }
@@ -62,7 +65,10 @@ export default function feathersExpress<S = any, C = any> (feathersApp?: Feather
 
       debug('Registering service with middleware', middleware);
       // Since this is a service, call Feathers `.use`
-      (feathersApp as FeathersApplication).use.call(this, location, service, { middleware });
+      (feathersApp as FeathersApplication).use.call(this, location, service, {
+        ...options,
+        middleware
+      });
 
       return this;
     },
--- a/packages/express/src/rest.ts
+++ b/packages/express/src/rest.ts
@@ -1,7 +1,7 @@
 import Debug from 'debug';
 import { MethodNotAllowed } from '@feathersjs/errors';
 import { HookContext } from '@feathersjs/hooks';
-import { createContext, getServiceOptions, NullableId, Params } from '@feathersjs/feathers';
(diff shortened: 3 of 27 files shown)


bertho-zero:
Great ! The only thing that bothers me is that the arguments are not free.

The signature of a method must be `[data, params]` and I currently have custom methods that expect `[id, params]` or `[id, data, params]`.

Maybe we can add flexibly if we give an array of argument to body? Which could be empty and to which we would concatenate `params` from the url.

daffl:
Oh, I forgot to mention, that I added a middleware that should still allow to register any other kinds of custom methods:

```js
const { createContext } = require('@feathersjs/feathers');
const { serviceMiddleware } = require('@feathersjs/express');

app.use('/myservice', myservice);

// With custom method handler
app.get('/myservice/:__feathersId/my-method', serviceMiddleware(async ({ req, res, options }) => {
  // options.id - __feathersId
  // options.params - Service params
  // options.data - body

  const { params, id } = options;

  return app.service('my-service').myCustomMethod(id, params);
}));

// With different params, using the body and returning the context (when using it with hooks)
app.post('/myservice/:myId/myMethod', serviceMiddleware(async ({ req, res, options }) => {
  const { myId } = req.params;
  const { params, data } = options;
  const context = createContext(service, method);

  return app.service('my-service').myCustomMethod(myId, data, params, context);
}));
```

It's not as declarative but I'm hoping it to be more flexible and explicit.

PR #2380, opened 2021-06-01
fix: update repository urls

lukashass:
### Summary

While looking at

- https://www.npmjs.com/package/@feathersjs/adapter-commons
- https://www.npmjs.com/package/@feathersjs/adapter-tests

I noticed the repository links where pointing to [a no longer existing repository](https://github.com/feathersjs/databases).

### DIFF
--- a/packages/adapter-commons/CHANGELOG.md
+++ b/packages/adapter-commons/CHANGELOG.md
@@ -3,29 +3,29 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
-# [5.0.0-pre.4](https://github.com/feathersjs/databases/compare/v5.0.0-pre.3...v5.0.0-pre.4) (2021-05-13)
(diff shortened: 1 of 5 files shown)


lukashass:
The changelogs of `adapter-commons`, `adapter-tests` and `memory` also include links to <https://github.com/feathersjs/databases>. I'm happy to change those as well, if you like.

daffl:
Thank you! Makes sense to probably change those as well since the repository has been removed completely.

lukashass:
I changed the urls in the changelogs, but left some occurrences of `https://github.com/feathersjs/databases` as they were. They would otherwise point to an existing, but wrong issue. For example [these](https://github.com/lukashass/feathers/blob/fix-repository-urls/packages/adapter-commons/CHANGELOG.md#453-2020-09-24) urls.

daffl:
Great, thank you! I guess it continued to generate wrong Changelog links because of the wrong reference package.json.

PR #2381, opened 2021-06-03
Support being a built-in CodeSandbox sandbox

FossPrime:
### Summary

Basically Support being part of the environment drop down in configs and template searches

Screenshot from 2021-06-02 19-10-53

https://codesandbox.io/s/

Based on: https://github.com/nuxt/codesandbox-nuxt/blob/master/sandbox.config.json

Full list of Templates: https://github.com/codesandbox/codesandbox-client/tree/master/packages/common/src/templates

### DIFF
--- a/sandbox.config.json
+++ b/sandbox.config.json
@@ -1,0 +1,8 @@
+{
+  "infiniteLoopProtection": true,
+  "hardReloadOnChange": false,
+  "view": "terminal",
+  "container": {
+    "node": "14"
+  }
+}


daffl:
Nice! Assuming with this merged it'll just show up in that list?

FossPrime:
Next step is to submit a PR with a link in the PWA client to an official git cloneable repo that has a package.json dev script. The first port exposed is the one used by the built in browser... the terminal gives you iptables 2 ss access which is unusually generous.

The bar for being included in the home page with the "isPopular" attribute is probably higher, but Adonis and some of their pet projects are on there, I don't expect much push back on that.

daffl:
Done. Once it is up maybe we can even include it in some parts of the docs.

FossPrime:
Prepping the PR for CodeSandbox...
- [ ] Make SVG TypeScript icon https://github.com/codesandbox/codesandbox-client/blob/master/packages/template-icons/src/TypeScriptIcon.tsx
- [ ] Add spec file in https://github.com/codesandbox/codesandbox-client/tree/master/packages/common/src/templates
  First file to open should probably be a Hook... but I'll leave it as ReadMe for now
- [x] Test the sandbox in the node Template
  - [ ] submit it as proof-of-concept https://github.com/codesandbox/codesandbox-client/blob/389073613e06eee944231f4aeef9dfa746c1b947/contributor-docs/adding-template.md#5-test-the-template
- [ ] Submit a PR for their repository importer https://github.com/codesandbox/codesandbox-importers/blob/master/packages/import-utils/src/create-sandbox/templates.ts

## July 20th Update:

Findings:
- 500 file limit
- no executable bits will transfer, so use bash or sh instead of `./`
- We have to define helmet frame ancestors, or disable it
  - https://stackoverflow.com/a/66058361/370238
- domain of server is unpredictable, usually a random subdomain of `.sse.codesandbox.io` for container sandboxes and csp.app for webpack ones.
- Changing more than 50 files in 2 seconds will disable file sync
- Container debugging doesn't seem to be supported... They have limited support for react
- Integration has to happen with GitHub, no GitLab, Gitter, Bitbucket etc. Changes are saved as Pull Requests.
- The default branch has to be configured to `main` in GitHub for the exporter to work
- Server restarts with a successful `yarn start` are necessary to accept container configs
- Using unsanctioned node versions is easy, NVM and bash are preinstalled

I have 3 projects to convert into CSP before I can get to this... I'll get to this after. Hopefully late this month.

FossPrime:
I got a fully loaded feathers-client example app running directory off Github, as a Github Sandbox

githubbox.com/rayfoss/feathers-pinia (It's just a 2 line edit on https://github.com/marshallswain/feathers-pinia ... i've submitted it as PR 5)

I'm not sure how to do the same with the Feathers Server... but it's my task to find out over the next 2 days.

Two weeks ago, they added Node 16 support. Five days ago, they had draft subpath Imports support in. Making it easier to get real production projects running.

Given the new monorepo situation, **it seems the best approach is to have a dedicated repo for sandbox deployment**, with Glitch and CodeSandbox support. With some thought given to how to auto-generate updates to it.

PR #2398, opened 2021-06-23
Fix additional query parameters causing issues

mrfrase3:
### Summary

the expressOauth code allows passing through more parameters than `redirect`, which all get lumped into an object called `query`, this then gets put into the params object on the other end of the oauth flow for convenience.

e.g. `https://mybackend.com/oauth/google?redirect=/checkout&cartId=1234`
results in the following in params being available `{ redirect: '/checkout', query: { cartId: 1234 } }`

The issue is that these params are passed directly into the patch/get/create service method calls in the strategy, so then the query is used to filter the patch request and an error is returned where the id is not found.

Omitting the query from the params fixes this issue.

### DIFF
--- a/packages/authentication-oauth/src/strategy.ts
+++ b/packages/authentication-oauth/src/strategy.ts
@@ -109,7 +109,7 @@ export class OAuthStrategy extends AuthenticationBaseStrategy {
 
     debug('createEntity with data', data);
 
-    return this.entityService.create(data, params);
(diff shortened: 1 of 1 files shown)


daffl:
That makes sense, I believe the import should be `import { createDebug, _ } from '@feathersjs/commons';` and then `_.omit` though.

daffl:
K, got it fixed, published in `v5.0.0-pre.5`

Issue #2414, opened 2020-11-08
A next generation CLI


daffl:
Maintaining a fully featured single application generator has proven very difficult and time consuming. Every new feature or prompt adds an exponential growth in code and test complexity. Templates are difficult to customize since they are buried in a third party module and difficult to maintain, especially when having to account for all kinds of customizations somebody would want to do.

A more flexible approach is used by [hygen.io](http://www.hygen.io/) which puts generator templates right into your application. If you want to change it, edit it to what you need and all future generator calls will reflect it.

To further improve modularity, each Feathers module will ship with the templates that can be used to generate its integration in a standard JavaScript/TypeScript Feathers setup. You can either just run it or eject the templates into your own application to customize.

For example

```
feathers generate authentication
```

Will look for templates that can be generated in the following modules:

- `./_templates/authentication` (locally)
- `@feathersjs/authentication/_templates`
- `feathers-authentication/_templates`
- `authentication/_templates`

Templates can be ejected into your application with

```
feathers customize authentication
```

This will allow any third party module to provide templates to generate its integration into a standard Feathers application setup and any user to customize what is being being generated to fit their own application needs (like creating templates that initialize a common set of protection hooks when generating a new service).

daffl:
This is now available in the v5 prerelease and can be used following the new guide at https://dove.feathersjs.com/guides. Closing this main issue, bugs and new features can be follow up with specific issues.

pkyeck-bling:
@daffl sorry to dig this up but was this somehow even possible in v4? or do we need to update to v5 to customize the templates?

daffl:
It is not but you can create your own templates with https://github.com/featherscloud/pinion

PR #2425, opened 2021-08-06
Feathers v5 CLI

daffl:
This pull request is the draft for a more modern and flexible generator and CLI tool based on [Hygen](http://www.hygen.io/) and a modernized and more concise new application structure where service class, new hooks and service registration are all together in the same file. It also uses ES modules for the JavaScript version and automatically generates and registers schemas and resolvers.

For more details on CLI usage see https://github.com/feathersjs/feathers/issues/2414

This is a big undertaking and the current progress is tracked here:

- [x] Fork Hygen to make it more embeddable (see https://github.com/feathersjs/hygen)
- [x] Create core CLI
- [x] Basic app generator (static files, package.json)
- [x] Test core generators in CI
- [ ] Eject templates via `feathers customize <name>`
- __JavaScript__
  - [x] App generator (Koa/Express app.js, service index, channels)
  - [x] Service base generator (import, hook registration etc.)
  - [x] Custom service generator
  - [ ] Memory service generator
  - [ ] Hook generator
  - [ ] Core authentication generator
  - [ ] Local authentication generator
  - [ ] oAuth authentication generator
  - [ ] Schema generator
- __TypeScript__
  - [x] App generator (Koa/Express app.ts, declaration, service index, channels etc.)
  - [x] Service base generator (import, hook registration etc.)
  - [x] Custom service generator
  - [ ] Memory service generator
  - [ ] Hook generator
  - [ ] Core authentication generator
  - [ ] Local authentication generator
  - [ ] oAuth authentication generator
  - [ ] Schema generator
- [ ] Legacy template generator


FossPrime:
Just a thought:

It sounds like hygen is more testable... aka, it can be easily configured programmatically without dealing with a file system. It would be really interesting to see what a stateless-ish call to the generator would look like in Runkit... That is... 

```
// path pwd, oauth credentials in env, production defaults. Assume reverse proxy handles SSL/DDOS/etc
new fjsCli({db: 'mongo://...', services:[{name: pog, hooks: [...] }]})
```

daffl:
If we do have file system access that could be possible. It is indeed much easier to use programmatically. Currently it looks like this: https://github.com/feathersjs/feathers/blob/hygen-cli/packages/cli/test/index.test.ts#L84-L96 - the `createPrompter` override is still a little clunky but since I'm using our own Hygen fork we can add what we need.

FossPrime:
- [package.json](https://github.com/feathersjs/feathers/blob/cad02961a433e38994947743c788e25438433496/packages/cli/_templates/app/base/ts/package.json.t#L13) shouldn't this be in tilde quotes?
- `feathers generate service` errors out with `TypeError: Cannot read property 'name' of undefined` on cli/_templates/service/base/index.js:8:20
  - is `feathers g custom service` the correct new syntax?
  - If so what should the comment in [line 80](https://github.com/feathersjs/feathers/pull/2425/files#diff-c987d2e848290b057cc104cf8979ff8315a502b6f65fedfe91f2115f0dfbb7faR80) of tests say?
  - it seems to be writing to `/services` on the root directory if directories.lib is not defined in package.json, not a huge deal... unless.
- ts-node is more up to date than ts-node-dev, using ts-node 10, and with better support for `"typescript": "^4.5.0-beta",`'s  tsconfig `"module": "nodenext",` https://codesandbox.io/s/ts-http-ygsxh?file=/tsconfig.json

https://codesandbox.io/s/feathers-dove-cs7qc?file=/src/services/index.ts

daffl:
This has been superseded by https://github.com/feathersjs/feathers/pull/2578

Issue #2312, opened 2021-04-22
Feathers Schema: A common data model definition format


daffl:
This issue consolidates the work from [feathersjs/schema](https://github.com/feathersjs/schema) and the discussion from https://github.com/feathersjs/schema/issues/20 into an up to date proposal in the main repository since I think for now it makes the most sense to include the `@feathersjs/schema` here (with the option to turn it into a standalone module later once it has been used for a bit).

## Problem

A very common problem we are seeing in web application development at the moment, especially in the NodeJS world, is that data model definitions are duplicated in many different places, be it ORM models, JSON schema validators, Swagger/OpenAPI specifications, TypeScript type definitions or GraphQL schemas. It is difficult to keep them synchronized and any new protocol or database is probably going to need the same work done again.

You can find converters from almost any specific format to any other format on npm but there are many challenges from relying on a compilation step (generating duplicate code that doesn't really have to exist in the first place) to only covering a small subset of functionality. It also appeared a majority of those modules is no longer actively maintained. ORMs are trying to address the challenge by locking you into their way of doing things with database or ORM specific idiosynchrasies dictating how flexible you can be in defining your model.

## Feathers schema

Feathers schema provides a common schema definition format for JavaScript and TypeScript that allows to target different formats like GraphQL, JSON schema (Swagger/OpenApi), SQL tables, validations and more with a single definition. Similar to how Feathers services allow different transport mechanisms to access their API without having to change your application code, schemas are Feathers answer to address the ever expanding list of data schema definition and validation formats.

It does not only address the definition of data types but also how to resolve them within the context of your application. *This is a different approach to most ORMs where you define your data model based on the database (or other ORM specific convention).* Some examples where schemas are useful:

- Ability to add read- and write permissions on the property level
- Complex validations and type conversions
- Shared model definition types between the client and server (when using TypeScript)
- Associations and loader optimizations
- Query string conversion, validation (no more type mismatches) and protections
- Automatic API docs

## How it works

`@feathersjs/schema` uses [JSON schema](https://json-schema.org/) as the main definition format with the addition of *property resolvers*. Resolvers are functions that are called with the application context and return the actual value of a property. This can be anything from the user associated to a message to the hashed password that should be stored in the database.

### Schema definitions

```js
import { schema } from '@feathersjs/schema';

const UserSchema = schema({
  type: 'object',
  required: [ 'email', 'password' ],
  additionalProperties: false,
  properties: {
    id: { type: 'number' },
[...]

PR #2441, opened 2021-08-30
feat(schema): Initial version of schema definitions and resolvers

daffl:
This pull request implements the first version of `@feathersjs/schema` as discussed in https://github.com/feathersjs/feathers/issues/2312. It features schema definitions based on JSON schema with static TypeScript type inference as well as a dynamic resolver system. Specifically:

- Schemas are used for data type definitions and simple (synchronous) validations
- Resolvers are used for modifying data based on the context like
  - Populating associations
  - Securing queries and e.g. limiting requests to a user
  - Removing protected properties for external requests or users without the right permissions
  - Hashing passwords and validating dynamic password policies etc.

There may be further iterations necessary but I'm planning on including it in the next pre-release because it is required for further CLI work (done in https://github.com/feathersjs/feathers/pull/2425).

Here is an example for a user schema definition and resolvers with protected passwords and query limitations (non-admins can only see and change their own data):

```ts
import {
  schema, resolve, Infer, resolveResult,
  propertyQuery, validateQuery, resolveQuery
} from '@feathersjs/schema';

export const userSchema = schema({
  type: 'object',
  additionalProperties: false,
  required: ['email', 'password'],
  properties: {
    id: { type: 'number' },
    email: { type: 'string' },
    password: { type: 'string' }
  }
} as const);

export type User = Infer<typeof userSchema>;

export const userDataResolver = resolve<User, HookContext<Application>>({
  properties: {
    password: async (value) => {
      // Return a hashed version of the password before storing it in the database
      return bcrypt(value);
    }
  }
});

export const userResultResolver = resolve<User, HookContext<Application>>({
  properties: {
    password: async (value, _user, context) => {
      // Do not return the password for external requests
      return context.params.provider ? undefined : value;
    }
  }
});

export const userQuerySchema = schema({
  type: 'object',
  additionalProperties: false,
  properties: {
    $limit: {
      type: 'number',
      minimum: 0,
      maximum: 100
    },
    $skip: {
      type: 'number'
    },
    // This will allow all of the Feathers query syntax
    id: propertyQuery({
      type: 'number'
    })
  }
} as const);

export type UserQuery = Infer<typeof userQuerySchema>;

export const userQueryResolver = resolve<User, HookContext<Application>>({
  properties: {
    id: async (value, _user, context) => {
      // Non admins can only query and modify their own data
      if(context.params?.user && context.params.user.role !== 'admin') {
        return context.params.user.id;
      }

      return value;
    }
  }
});

app.service('users').hooks([
  validateQuery(userQuerySchema),
  resolveQuery(userQueryResolver),
  resolveResult(userResultResolver
]);
```


deskoh:
@daffl Is it possible to allow a singleton `ajv` instance to be configured? Custom AJV instance would be useful for options like [`removeAdditional`](https://ajv.js.org/options.html#removeadditional) and can be used for data sanitization (e.g. `const sanitizedData = myschema.validate(data)`).

daffl:
@deskoh Passing a custom AJV instance is not possible and documented in https://dove.docs.feathersjs.com/api/schema/schema.html#options

PR #2456, opened 2021-10-03
feat(basic hooks) allow groups by method then type

marshallswain:
### Summary

- Introduces the terminology of "Basic Hooks" and "Advanced Hooks"
   - Any reference of "Legacy Hooks" to "Basic Hooks".
   - Renames any reference of "Async Hooks" to "Advanced Hooks"
- Introduces support for registering basic hook objects  grouped by method.  To prevent a lot of code changes, it converts the new format to the old format, internally.

There are now two supported Basic Hook registration formats.  The new format prevents having to scoll past all other methods' hooks in order to see the after hooks.

```ts
const  newFormat = {
  find: {
    before: [hook1, hook2, hook3],
    after: [hook4, hook5, hook6],
    error: [error1, error2, error3]
  },
  get: {
    before: [hook1, hook2, hook3],
    after: [hook4, hook5, hook6],
    error: [error1, error2, error3]
  },
}
```

### Other Information

The majority of the changes in this PR are for renaming the hooks from `legacy` to `basic`.

The new hook format feature is added in `application.ts`, `hooks/basic.ts`, and the test file.
- The [`app.hooks()` method](https://github.com/feathersjs/feathers/pull/2456/files#diff-52e395dc6b28077aefd1ed36163fd5c50f7e23ca71dd56c8a8a08d393fd46684R123-R126) now checks for an object that has nested hook types.  Please double-check this.
- A new [`groupBasicHooksByType` private utility](https://github.com/feathersjs/feathers/pull/2456/files#diff-5ca7fe5525ed3155322fac494f81c2be465060933eb1b8030616e329a3a9eca9R85-R97) has been added.  It's "private" in the sense that it isn't exported from the main module, but just the basic hooks module.
- The [`service.hooks` method](https://github.com/feathersjs/feathers/pull/2456/files#diff-9dd442ca294f7dba145a85914271f3ff3b2ffbca851c4ef1eaeb63329bba30eaR87-R91) detects the new format and uses the above utility to convert the hooks.
- I added [two tests](https://github.com/feathersjs/feathers/pull/2456/files#diff-8a706c65373f4cc3169dcfb691543e66e6ecc78a568259801517018a23c8b44cR105-R200)
   - A unit test for the `groupBasicHooksByType` method.
   - A unit test for services to make sure hooks are registered and run correctly.

### Pending Work

- [ ] Update the Deno version.  Is this done by a build process or is it manual?
- [ ] Documentation
   - [ ] Clarify differences between Basic and Advanced Hooks
   - [ ] Show the new Basic Hook format.
- [ ] Maybe update the `BasicHookOptions` type to be the new format, since it makes more sense and doesn't introduce any negative effects or mental overhead.

### DIFF
--- a/deno/_feathers/application.ts
+++ b/deno/_feathers/application.ts
@@ -17,9 +17,9 @@ import {
   HookOptions,
   FeathersService,
   HookMap,
-  LegacyHookMap
+  BasicHookMap
 } from './declarations.ts';
-import { enableLegacyHooks } from './hooks/legacy.ts';
+import { enableBasicHooks } from './hooks/basic.ts';
 
 const debug = createDebug('@feathersjs/feathers');
 
@@ -33,11 +33,11 @@ export class Feathers<ServiceTypes, AppSettings> extends EventEmitter implements
     [HOOKS]: [ (eventHook as any) ]
   };
 
-  private legacyHooks: (this: any, allHooks: any) => any;
+  private basicHooks: (this: any, allHooks: any) => any;
 
   constructor () {
     super();
-    this.legacyHooks = enableLegacyHooks(this);
+    this.basicHooks = enableBasicHooks(this);
   }
 
   get<L extends keyof AppSettings & string> (name: L): AppSettings[L] {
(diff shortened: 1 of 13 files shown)


daffl:
I think the renaming from legacy to basic hooks makes definitely sense. Can we split the renaming and the new registration format into two different PRs?

The reason why I suggested having a "wrapper" function was that this could then also work when using [@feathersjs/hooks](https://github.com/feathersjs/hooks#class-hooks) directly on a service class which I personally would recommend over using `app.service('name').hooks`:

```js
import { hooks } from '@feathersjs/hooks'

class MyService {
  async get (id) {
    return { id }
  }
}

hooks(MyService, {
  get: [
    basicHooks({
      before: [],
      after: [],
      error: []
    }),
    myNewAsyncHook()
  ]
})
```

marshallswain:
@daffl that makes sense.  Closing this in favor of a solution built into `@feathersjs/hooks`

PR #2457, opened 2021-10-04
[dove/hygen-cli] Fix capitalized Yarn command issue

FossPrime:
# Bug fix for

hygen generator won't run when Yarn is selected...

Screenshot 2021-10-04 14 23 59

when using these settings:
Screenshot 2021-10-04 14 23 28

# Details

## Why the tests are failing

The time out after 5 minutes, on CI tools it takes 3m30s just to get through lerna bootstrap... leaving almost no time for tests to run in CI=true env's. Solution could be to remove the automatic full monorepo boot... and only boot say, CLI with `lerna bootstrap --ignore-scripts --scope "@feathersjs/cli" --include-dependencies`

## Why I had to mess with the semicolon

Which if we only fixed the capitalization would yield:

```
/home/sandbox/.nvm/versions/node/v16.8.0/bin/node /home/sandbox/.nvm/versions/node/v16.8.0/bin/yarn add @feathersjs/feathers@^5.0.0-pre.6 @feathersjs/errors@^5.0.0-pre.6 @feathersjs/configuration@^5.0.0-pre.6 @feathersjs/authentication@^5.0.0-pre.6 @feathersjs/transport-commons@^5.0.0-pre.6 winston@^3.3.3 @feathersjs/socketio@^5.0.0-pre.6 @feathersjs/koa@^5.0.0-pre.6 koa-static@^5.0.0 ; yarn add nodemon@^2.0.12 axios@^0.21.4 mocha@^9.1.1 @types/mocha@^9.0.0 @types/koa-static @types/node@^16.9.4 ts-node-dev typescript@^4.4.3 shx@^0.3.3 --dev
```

and `Error: https://registry.yarnpkg.com/;: Not found`.

Hygen prefers the fail deadly approach of `&&` which we could employ by swapping the semicolon in `packages/cli/_templates/app/base/js/package.json.t` line 4

`sh: "<%= h.install(...dependencies) %>; <%= h.installDev(...devDependencies) %>"`

# Guru meditation:

Upon further testing, I see no benefit to having the generator use Yarn, it's not appreciably faster than NPM 7 for anything the generator would install. The generator runs in about 15 seconds on a shared CSB container with NPM... I'm in favor of dropping the whole choose your packager shenanigans. You can still use Yarn or PNPM after generating with minimal effort.

Also I'm in favor of running a yearly doodle to set defaults, Typescript should really be the first choice. And feathers-memory or [@seald-io/nedb](https://www.npmjs.com/package/@seald-io/nedb) for the database adapter, there is currently no DB adapter option that doesn't require spinning up a goliath.

### Summary

here](https://github.com/feathersjs/feathers/blob/crow/.github/contributing.md#pull-requests))

### DIFF
--- a/packages/cli/src/utils.ts
+++ b/packages/cli/src/utils.ts
@@ -54,8 +54,13 @@ export async function getHelpers (pkg: PackageJSON, self: PackageJSON, _logger: 
           ? `${name}@${self.devDependencies[name]}`
           : name
       );
-      const { packager } = pkg.feathers;
-      const command = `${packager} install ${deps.join(' ')} --${dev ? 'save-dev' : 'save'}`;
+      const packagerName = pkg.feathers.packager
+      const packagerCmd = packagerName.toLowerCase()
+      const isYarn = packagerName === 'Yarn'
+      const installCmd = isYarn ? 'add' : 'install'
+      const devArg = isYarn ? '--dev' : '--save-dev'
+      const cmdArgs = dev ? ' ' + devArg : ''
+      const command = `${packagerCmd} ${installCmd} ${deps.join(' ')}${cmdArgs}`
 
       return command;
     },


daffl:
Thank you for putting that up! I'm always in favour of reducing options and complexity especially since it indeed no longer has a real speed advantage. Question would be if this causes other issues if a generator wants to install dependencies.

FossPrime:
This PR does retain Yarn support... and fixes it.


## If we wanted to get rid of the Yarn prompt, yarn dep install gates and package.json scripts gates

we would at some point need to update `yarn.lock` if you are using Yarn... but that's something that can be done as a post/after hook, by calling a quick `yarn` command, which defaults to `yarn install`. We can do that if we see a `yarn.lock` file and don't need to ask the user.

package-lock.json would still exist with a default create generator... but that should be the job of `npm config set package-lock false` to deal with. The only safe way to delete it is to check it doesn't exist before we start. It would still get created during an early abort scenario.

yarn reminds me of TypeScript... it is not canonical to import 'foo.ts' or publish raw typescript to NPM, the canonical thing to do is to pretend it doesn't exist when possible.

## Alternative PR that adds passive yarn support

In package.json.t, we would call:  
`<%= h.install(...dependencies) %>; <%= h.installDev(...devDependencies) %>`  
but append ` ; h.yarnInstall()` that would reference a getHelpers function that returned `yarn` if yarn.lock is readable.

That looks like this
`<%= h.install(...dependencies) %> ; <%= h.installDev(...devDependencies) ; h.yarnInstall() %>`  
on every generator... such as feathers-oauth

I'd prefer

`<%= h.install(dependencies, devDependencies) %>` which handles yarn/pnpm for us and generators/templates don't have to think about it.

daffl:
I like that idea. I'm all for reducing prompts from the initial generator (who knows what package manager/language/linting configuration will be hip tomorrow 🤷). Since templates are easy to add and customize now I'd propose a

```
feathers generate yarn
```

Which removes your `package-lock.json`, adds `yarn` as an engine and reinstalls everything with the yarn command. Although it's a little hacky, internally we could just check for `yarn.lock` as you suggested and adjusts the install command accordingly.

daffl:
I also had to do some messing around with the installation since with the shell command for some reason all dependencies ended up in either `dependencies` or `devDependencies`. I think it was some `npm` synchronization issue or something so I changed it to work like this: https://github.com/feathersjs/feathers/blob/hygen-cli/packages/cli/_templates/app/base/index.js#L61

Issue #1443, opened 2019-07-06
Hooks returning new context


vonagam:
I think that returning new context object from a hook should be depreciated.

Upsides:

1) Clearer expectations of a hook - it just a function that mutates a context.
1) Clearer expectations of a context - it stays the same all the way for a call.
1) Less bugs, less code, clearer code, more performant code for users and feathers.
1) Hooks returning new object lose their context changes on a throw unlike mutating ones.
1) Migration from returning new object to mutating existing one should have no problems.

Downsides:

1) No backward compatibility for such hooks, need to migrate such hooks (even if this is easy).

Example: look at snippets in #932, they will produce unexpected behaviour if some hook down the chain replaces a context. (Even though those snippets are just illustrations, it is relevant since errors came from not right expectations, but if replacing hooks were depreciated there would be no problems besides unnecessary return statements).

Error right in the first snippet (same problem in others):
```js
async (context, next) => {
  try {
    // ...

    // next() may return new context instance
    // but context variable not updated
    await next();
    
    // ...
    
    // will return initial passed context
    // possibly replacing new instance, losing everything
    return context;
  } catch(error) {
    // ...
  }
}
```

So you wouldn't be able to write `const { result } = await next()` if you want to write `return context`. You would need to write:

```js
const { result } = context = await next();
// or if you replace context
const { result } = context = await next(context);
```

What are your thoughts? Are there any use cases where replacing context with new instance is easier than mutating existing one? Any justification to keep such feature besides backward compatibility?

daffl:
I'm going back and forth about this as well. I kind of agree, on the other hand, having hooks be more functional and immutable would help prevent many subtle and confusing issues. More specifically, the problem is that a hook can modify the calling objects in ways you usually wouldn't expect which is especially difficult when other services are called. For example with the following:

```js
const otherHook = context => {
  context.params.query.name = 'david';
}

const myHook = context => {
  context.params.query.name = 'nobody';
  await context.app.service('otherservice').find(context.params);
}

app.service('otherservice').hooks(otherHook);
app.service('myservice').hooks(myHook);
```

Used like this:

```js
const params = { query: { name: 'somebody' } };

await app.service('myservice').find(params);

console.log(params.query.name);
```

I'd definitely expect `console.log(params.query.name)` to show `'somebody'` but it will actually be `'david'` set in a totally different part of the application two levels down. Usually these kinds of issues are easily prevented by making a copy of the query like this:

```js
const otherHook = context => {
[...]

PR #2462, opened 2021-10-05
fix(core): Allow to return a new hook context in basic hooks

### LINKED ISSUE #2451: returned hook context not used anymore in workflow
### Steps to reproduce

As long as feathers dove should behave like feathers crow, it is currently not possible to create a new context object which will returned and reused in the following hooks.

If I create a new context object with spread operators during the hook, changing some query parameters and returning this context at the end of the hook. It will be used in the following hooks in feathers v4, in feathers v5 the original context from the function parameter will be reused in the next hooks and not the object what was return by the previous hooks. That means the return object isn't available in the following workflow (service and hooks).

Example:

```js
const worksInV4AndV5 = (context) => {
  context.params.query.foo = 'worksInV4AndV5';
  return context; 
};

const worksInV4NotInV5 = (context) => {
  return {
    ...context,
    params: {
      ...context.params,
      query: {
        ...context.params.query,
        foo: 'worksInV4NotInV5'
      }
    }
  };
};

const logHook = (context) => {
    // logs in v4  'worksInV4NotInV5' 
    // logs in v5  'worksInV4AndV5' 
    console.log(context.params.query.foo);
};

const hooks = {
    before: {
      all: [],
      get: [
        worksInV4AndV5,
        worksInV4NotInV5,
        logHook
      ],
      find: [
        worksInV4AndV5,
        worksInV4NotInV5,
        logHook
      ],
      create: [],
      update: [],
      patch: [],
      remove: [],
    },
    after: {
      all: [],
      get: [],
      find: [],
      create: [],
      update: [],
      patch: [],
      remove: [],
    },
    error: {
      all: [],
      get: [],
      find: [],
      create: [],
      update: [],
      patch: [],
      remove: [],
    }
  };

...

service.hooks(hooks);

```

If I run the application and making a get on the service, it will log in v4  'worksInV4NotInV5' and in version 5  'worksInV4AndV5'.

### Expected behaviour

The returned context should be used in the following hooks, like in version 4 of feathers.

### Actual behaviour

Currently, in version 5 it doesn't matter if we will return a new context object, the current or undefined. The original context object will be used in the following hooks. That means if we don't manipulate the original object, those changes aren't available in the following hooks.

### System configuration

feathers dove (Version 5 pre.9 and pre.10)

**NodeJS version**:

Version v12.22.6, v14.17.6 and v16.9.1

**Operating System**:

macOS Catalina Version 10.15.7

daffl:
This fixes a regression in backwards compatibility that is difficult to debug.

Closes #2451

### DIFF
--- a/packages/feathers/src/hooks/legacy.ts
+++ b/packages/feathers/src/hooks/legacy.ts
@@ -2,16 +2,20 @@ import { _ } from '../dependencies';
 import { LegacyHookFunction } from '../declarations';
 
 const { each } = _;
+const mergeContext = (context: any) => (res: any) => {
+  if (res && res !== context) {
+    Object.assign(context, res);
+  }
+  return res;
+}
 
 export function fromBeforeHook (hook: LegacyHookFunction) {
   return (context: any, next: any) => {
     context.type = 'before';
 
     return Promise.resolve(hook.call(context.self, context))
-      .then(ctx => {
-        if (ctx && ctx !== context) {
-          Object.assign(context, ctx);
-        }
+      .then(mergeContext(context))
+      .then(() => {
         context.type = null;
         return next();
       });
@@ -20,15 +24,15 @@ export function fromBeforeHook (hook: LegacyHookFunction) {
 
 export function fromAfterHook (hook: LegacyHookFunction) {
   return (context: any, next: any) => {
-    return next().then(() => {
-      context.type = 'after';
-      return hook.call(context.self, context)
-    }).then((ctx: any) => {
-      if (ctx && ctx !== context) {
-        Object.assign(context, ctx);
-      }
-      context.type = null;
-    });
+    return next()
+      .then(() => {
+        context.type = 'after';
+        return hook.call(context.self, context)
+      })
+      .then(mergeContext(context))
+      .then(() => {
+        context.type = null;
+      });
   }
 }
 
@@ -45,11 +49,7 @@ export function fromErrorHooks (hooks: LegacyHookFunction[]) {
 
       for (const hook of hooks) {
         promise = promise.then(() => hook.call(context.self, context))
-          .then(ctx => {
-            if (ctx && ctx !== context) {
-              Object.assign(context, ctx);
-            }
-          })
+          .then(mergeContext(context))
       }
 
       return promise.then(() => {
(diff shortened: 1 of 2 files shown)


vonagam:
Wait, but it wasn't an unexpected regression, it was a deliberate choice for depreciation in next major version advocated by me in #1443.

This current fix, while small and easy creates a different behavior from v4. Here you just `Object.assign` returned object, before discarding it, to original context. Examples of problems that may lead to: 
- If somebody to `delete` property (custom or `error`, `id`) from a new context, that change won't be transferred by `Object.assign` and will be lost.
- `Map`/`WeakMap` with context as a key will not work, since while one will think that they have a reference to new current context, what they actually have is an update to original one.
- As before, if you start making changes to new context object but then something throws an error, all your changes are lost, which creates difference in behavior with non-returning hooks.
- There is also a composition problem in v4 that i did not mention before - if you work with arbitrary passed-in hooks then you yourself should handle cases when they return contexts and the only consistent working way is to start returning too, which leads to propagation of a bad practice. 

And migration is easy, they could have done the same that was done here - instead of `return {...context, ...update}` just write `Object.assign(context, update)` - it works (and more proper) both in v4 and v5.

Anyway, you can ignore this message, just had to write since it was me who proposed depreciation.

daffl:
Ah yes, sorry I remember now. The problem with this approach was that I heard from a couple of beta testers now that they are ensuring immutability in their projects and did something like:

```js
context => {
  return {
    ...context,
    params
  }
}
```

The problem with the new implementation was that it just did nothing and you had no idea why. You may be right in that it makes more sense to throw an error if a different context than the original one is returned.

vonagam:
> had no idea why

Legacy hook [return type](https://github.com/feathersjs/feathers/blob/b7143d4c0fbe961e714f79512be04449b9bbd7d9/packages/feathers/src/declarations.ts#L322) should have been updated to reflect that returning is not an option anymore. Then they could see (at least those who use typescript) an error at writing time instead of compile/runtime one.

> ensuring immutability

Good intention, but stems from miscommunication (docs and types). Originally returning new context was added (as you said) with an idea of a pure hook (receive input, return output without side effects), but it was never actually implemented like that in the end. So no profits, only costs for that mirage...

Another smallish thing is that migration from void hooks to async hooks are easier than from returning ones. So if you plan to force migration to new hooks in the distant future (one or two versions down the line), they would still need to update.

PR #2469, opened 2021-10-13
feat(authentication-oauth): Allow dynamic oAuth redirect

### LINKED ISSUE #2285: OAuth2 dynamic redirect doesn't work
### Steps to reproduce

As per documentation, OAuth2 dynamic redirect may be set as:

    <a href='http://myserver.url/oauth/google?redirect=login'>

However, it is ignored.  If `authentication.oauth.redirect` config key is present, it is taken into account, if not, JSON is returned. The query string is never honored.

In general, I wonder what is the optimal scenario for handling redirects, as the API server may be serving varios clients, that shouldn't be obliged to pass their own URLs for redirects, but the server has no means of differentiating between them.

The documentation says that 

> ```
> <a href="/oauth/github?redirect=dashboard">
>   Login with GitHub
> </a>
> 
> ```
> With the above configuration will redirect to `https://app.mydomain.com/dashboard` after the OAuth flow.

Strictly speaking, it will redirect me to `/dashboard` on the server, not to `mydomain.com` which probably initiated the request.

**Module versions**

`4.5.11` and `5.0.0-beta.0`

### LINKED ISSUE #2430: Allow multiple oAuth redirect domains
This is an issue brought up in https://github.com/feathersjs/feathers/issues/2430 and by @cliffvick and confirmed by @mrfrase3: Currently oAuth authentication only supports redirecting to a single frontend URL after a successful login. However, sometimes you may want to authenticate the same API with different frontends using [dynamic redirects](https://docs.feathersjs.com/api/authentication/oauth.html#redirects). This can already be done by [customizing the oAuth strategy `getRedirect`](https://docs.feathersjs.com/api/authentication/oauth.html#customization) like this:

https://gist.github.com/daffl/d0dee9cda7eee3d270bc99a1f1c67d9b

It might be worth adding an additional configuration option of `allowedRedirects` that takes an array of allowed URLs and checks the dynamic query `redirect` against that list similar to the above Gist.

daffl:
This pull request allows dynamic oAuth redirects by adding an `origins` option to the oAuth configuration. This can be added _instead_ of the `redirect` option to include a list of URLs that are allowed to make oAuth requests. It will use the `Referer` HTTP header from the original request to redirect back to. For example a configuration like

```json
{
  "authentication": {
    "oauth": {
      "origins": [
        "https://feathersjs.com",
        "https://feathers.cloud"
      ]
    }
  }
}
```

Would allow and redirect any request made from those domains.

- Closes #2285
- Closes #2430

### DIFF
--- a/packages/authentication-oauth/src/express.ts
+++ b/packages/authentication-oauth/src/express.ts
@@ -20,6 +20,7 @@ declare module 'express-session' {
       accessToken: string;
       query: { [key: string]: any };
       grant: { [key: string]: any };
+      headers: { [key: string]: any };
   }
 }
 
@@ -54,13 +55,14 @@ export default (options: OauthSetupSettings) => {
       }
       req.session.redirect = redirect as string;
       req.session.query = query;
+      req.session.headers = req.headers;
 
       next()
     });
 
     authApp.get('/:name/authenticate', async (req: Request, res: Response, next: NextFunction) => {
       const { name } = req.params ;
-      const { accessToken, grant, query = {}, redirect } = req.session;
+      const { accessToken, grant, query = {}, redirect, headers } = req.session;
       const service = app.defaultAuthentication(authService);
       const [ strategy ] = service.getStrategies(name) as OAuthStrategy[];
       const params = {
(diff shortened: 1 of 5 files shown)


emmanuelgeoffray:
Hi @daffl 
Thank you for this nice feature.
I think this new `origins` option should be omitted when authentification-oauth is parsing each strategy.
https://github.com/feathersjs/feathers/blob/dove/packages/authentication-oauth/src/index.ts#L51

daffl:
@emmanuelgeoffray Can you open a new issue with a more detailed description?

PR #2472, opened 2021-10-16
fix(adapter-tests): Add tests for pagination in multi updates

fratzinger:
### Background

This PR is coming from https://github.com/feathersjs-ecosystem/feathers-sequelize/pull/364 and https://github.com/feathersjs-ecosystem/feathers-sequelize/issues/363 . It's a discussion between me, @daffl 
 and @DaddyWarbucks if, on a service with default pagination, a `patch:multi` without any paginate in `params` should patch only a subset of items (default pagination) or all items (`paginate: false`). The same applies to `remove:multi`.

In my opinion there are two points:
1. What is the most anticipated behavior? How many items do you expect to patch, when you do something like `service.patch(null, { isConfirmed: true }, { query: { canBeConfirmed: true, isConfirmed: false }}`? All items that fulfill this criteria or a maximum of 10 items (most default pagination). Default pagination for `find` method is a no-brainer, because the structure changes from `{ data, skip, total }` to `any[]` and you can check wether `data.length` is the same as `total`. For `patch:multi` and `remove:multi` you only have `any[]` either way. You don't know, if you patched/removed *all* items or not. This is a point for `paginate: false` as default for `patch:multi` and `remove:multi`
2. For `patch:multi` and `remove:multi` sometimes you want to find all items that will be changed/removed. I need that in my applications for permissions, changelogs (need to know the old value before the change). So I would expect that a `find` and a `patch`/`remove` return the exact same items. This is a point for `paginate.default` as default for `patch.multi` and `remove:multi`

As point 1 as for `paginate: false`, point 2 is stronger to me. Either way I'll add a hook to my `app.hooks.js` file that ensures the same pagination behavior for `find`, `patch:multi` and `remove:multi`

### Solution

While the discussion for or against `paginate: false` as default for `patch:multi` and `remove:multi` is somewhat ongoing., we should ensure that all adapters act the same. That's where this PR comes into play. I added a standardized adapter-test to check this behavior. As `feathers-sequelize` has `paginate.default` as default `patch:multi` and `remove:multi`, the test exactly checks against this behavior. I added the test to the internal `@feathersjs/memory` and it fails because `getEntries` is used in `_patch` which uses `paginate: false` by default. So the behavior of `feathers-sequelize` and `feathers-memory` is different in this case.

I also added typings for `@feathersjs/adapter-tests` with `AdapterTestName` replacing `string` to prevent typos.

### Todos:

- [x] Decide wether `paginate: false` or `paginate.default` should be the default behavior. @daffl
- [x] When the behavior is clear, change the test to ensure this behavior @fratzinger
- [x] Add the test to `@feathersjs/memory` (currently `paginate: false` overwrites anything _confirmed_) @fratzinger
- [x] ~Add a note in docs for `patch:multi` and `remove:multi` so it's clear for everybody @fratzinger~ *not needed for no pagination - since it's intuitive*
- [x] Add a test for `remove:multi` @fratzinger
- [ ] Release a new `adapter-tests` package @daffl
- [ ] Add the tests to `feathers-sequelize` (currently `paginate.default` does not overwrite anything _confirmed_) @fratzinger
- [ ] Add the tests to `feathers-objection` (currently `paginate.default` does not overwrite anything _first glance_) @fratzinger
- [ ] Add the tests to `feathers-knex` (currently `paginate: false` overwrites anything _first glance_) @fratzinger
- [ ] Add the tests to `feathers-mongoose` (currently `paginate: false` overwrites anything _first glance_) @fratzinger
- [ ] Add the tests to `feathers-mongodb` (currently `paginate: false` overwrites anything _first glance_) @fratzinger
- [ ] Add the tests to `feathers-nedb` (currently `paginate: false` overwrites anything _first glance_) @fratzinger
- [x] ~Add a hook to `feathers-hooks-common` to ensure the same behavior for `find`, `patch:multi` and `remove:multi` @fratzinger~ *not needed*


daffl:
Sorry just revisiting this. It's actually a good question. I always looked at it that multi `patch` and `remove` should change and return everything since getting paginated results seems counter intuitive. On the other hand, the entire reason why there is a default for pagination is to avoid huge server loads and result payloads.

daffl:
Following up again, the problem is that you can't paginate a request that actually modifies things. So it definitely should return everything.

DaddyWarbucks:
I agree that the adapters should patch/remove everything. I think I was the one that initially mentioned "Not patching everything is probably good practice, thats why pagination exists and the user can disable it if needed". But I also agree it is counter intuitive.   `feathers-sequelize` uses a `$returning` param to indicate if the results should actually be returned. See: https://github.com/feathersjs-ecosystem/feathers-sequelize/blob/db4b5ff9b89df9b163f9cd5d8c32de8c915ff71c/lib/index.js#L245
This seems like a good practice. It allows the patch/remove to actually modify all records as the user would expect, but also serves a similar purpose with pagination in avoiding huge server loads.

daffl:
It might make sense to allow turning it off. Problem is that you won't get real-time updates then. 

Just to confirm though, this PR should be good to merge now right? Also, nice addition on the adapter name typings!

fratzinger:
Yes, I think so. It's weird, that I changed the adapter-test from testing against default pagination to no pagination and the test is still passing for `@feathersjs/memory`. So https://github.com/feathersjs/feathers/pull/2472/commits/430545938ea819a98b516a22ec212043656c4c4c (see above in this PR) was meant to fail, but has passed. Don't know why.

Other thing. I thought about backporting the two tests to crow adapter-tests so we can add the tests to all common adapters sooner. What do you think?

PR #2478, opened 2021-10-24
fix: Small improvements

vonagam:
- Added missing dependencies.
- Renamed `Application` type parameters `ServiceTypes` to `Services` and `AppSettings` to `Settings`. That's the change that does not solve anything and a breakish one (since when somebody augment a module they have to use same names). Can remove.
- Default `T` to `any` in `Service` type (and its relatives), so that you can simply write `Service` instead of `Service<any>`. There is some slight difference between `Service`/`Service<any>` and `Service<any, any>` - first one has `D` as `Partial<any>`. If this change comes up somewhere, there is a way to add new type later to default `D` to `any` if `T` is `any`.
- `authentication` package had a problem with [`lib/index.d.ts`](https://unpkg.com/@feathersjs/authentication@5.0.0-pre.14/lib/index.d.ts) where there was an import of `@feathersjs/hooks` which is not a direct dependency of the package, fixed by tweaking exports.
- in `transport-commons` fix types for `Publisher` (wrong type for `context` arg) and `Application.channel` (not supporting an argument of type `string | string[]`).
- Rename `CustomMethod` type to `CustomMethods` and allow to easily specify `data` and return type or keep it `any` as it currently is.
- Pass hook enabled methods to `enableLegacyHooks` (so include custom and filter out unused ones).
- In `koa` add headers to params (not sure if should be cloned as query is, in `express` neither query nor headers are cloned).
- In `koa` add types for feathers props on koa context - `feathers` and `hook`.
- In `transport-commons` and `express` remove `hooks` dependency and use correct type for context - it isn't supposed to work with generic hook context, only with feathers one.

### DIFF
--- a/packages/adapter-commons/test/service.test.ts
+++ b/packages/adapter-commons/test/service.test.ts
@@ -27,7 +27,7 @@ describe('@feathersjs/adapter-commons/service', () => {
   });
 
   describe('works when methods exist', () => {
-    class MethodService extends AdapterService implements InternalServiceMethods<any> {
+    class MethodService extends AdapterService implements InternalServiceMethods {
       _find (_params?: Params) {
         return Promise.resolve([]);
       }
--- a/packages/adapter-tests/package.json
+++ b/packages/adapter-tests/package.json
@@ -48,6 +48,7 @@
     "access": "public"
   },
   "devDependencies": {
+    "@types/mocha": "^9.0.0",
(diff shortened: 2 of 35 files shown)


daffl:
These are great, thank you! I think the CustomMethods change makes sense, we'll just have to update the docs. I wonder if there is a better way in TypeScript to do that. We're having a similar problem making service method typings more flexible.

PR #2482, opened 2021-10-27
feat(transport-commons): Ability to register routes with custom params

vonagam:
Allows to add routes with some predetermined route params which are not placeholder based.

```ts
// context.params.route.some will be 'param' for '/some/path' 
app.routes.insert('/some/path', { service, params: { some: 'param' } });
```

### DIFF
--- a/packages/transport-commons/src/routing/index.ts
+++ b/packages/transport-commons/src/routing/index.ts
@@ -1,4 +1,4 @@
-import { Application, Service } from '@feathersjs/feathers';
+import { Application, Service, ServiceOptions } from '@feathersjs/feathers';
 import { Router } from './router';
 
 declare module '@feathersjs/feathers/lib/declarations' {
@@ -41,8 +41,10 @@ export const routing = () => (app: Application) => {
   app.lookup = lookup;
 
   // Add a mixin that registers a service on the router
-  app.mixins.push((service: Service, path: string) => {
-    app.routes.insert(path, { service });
-    app.routes.insert(`${path}/:__id`, { service });
+  app.mixins.push((service: Service, path: string, options: ServiceOptions) => {
+    const { routeParams: params = {} } = options;
+
+    app.routes.insert(path, { service, params });
+    app.routes.insert(`${path}/:__id`, { service, params });
   });
 };
--- a/packages/transport-commons/src/routing/index.ts
+++ b/packages/transport-commons/src/routing/index.ts
@@ -4,42 +4,45 @@ import { Router } from './router';
 declare module '@feathersjs/feathers/lib/declarations' {
   interface RouteLookup {
     service: Service,
-    params: { [key: string]: string }
+    params: { [key: string]: any }
   }
 
   interface Application<Services, Settings> {  // eslint-disable-line
-    routes: Router;
+    routes: Router<{
+      service: Service,
+      params?: { [key: string]: any }
+    }>;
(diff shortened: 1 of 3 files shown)


daffl:
Yep, can definitely see that being useful. I think we can also add it to the new service options (https://dove.docs.feathersjs.com/api/application.html#use-path-service-options) to make it more integrated:

```js
app.use('/messages', new MyService(), {
  methods: [ 'get', 'doSomething' ],
  events: [ 'something' ],
  params: { something: 'here' }
});
```

packages/transport-commons/src/routing/index.ts
```diff
@@ -4,42 +4,45 @@ import { Router } from './router';
...
+  }
 
-      return result;
-    }
-  });
+  app.routes = new Router();
+  app.lookup = lookup;
 
   // Add a mixin that registers a service on the router
   app.mixins.push((service: Service, path: string) => {
```
daffl: Essentially this would become a

```ts
app.mixins.push((service: Service, path: string, options?: ServiceOptions) => {
  const { params = {} } = options;

  app.routes.insert(path, { service, params });
```

vonagam: Make sense, but wouldn't `route` be a better name for that option?

vonagam:
One question that i had when looking through code:

Is it possible to register same instance of a service on different routes? [This line](https://github.com/feathersjs/feathers/blob/b8eb804158556d9651a8607e3c3fda15e0bfd110/packages/feathers/src/service.ts#L74) makes me think that this possibility was thought about at some point.

If yes then it means that you can pass server options only during first `use` call since `wrapService` does not support having multiple settings stored. So if you pass options second time they will be silently ignored which is a bug.

If no then shouldn't `wrapService` throw an error (instead of returning) when it receives already wrapped service?

packages/transport-commons/src/routing/index.ts
vonagam: Makes sense. But i think that `params` can be confused with `params` on context, meaning that it looks like a value will be merged directly into `params`, not into `params.route`. The clearest name is `routeParams`, but it is two words.

vonagam: Added `routeParams` to service options. Also fixed `ServiceMixin` type a little (`options` argument can't be undefined) and updated `route` on `params` to reflect that it can contain any values now, not only string ones.

Unresolve the conversation if you think that `params` is better name (I don't have strong opinion on this and ok with whatever).

daffl:
Great, thank you! This looks good to me.

Good point about re-using the options. Can we put up a test for this to see what would be a good way to go about that? The goal was basically to

- Re-use existing services (but possibly with different options when passed)
- Have custom or plugin services provide their own options (e.g. custom methods and events) without somebody having to explicitly define them on registration

vonagam:
Hm... if plugins will provide their own options using `SERVICE` symbol then it means that options provided on a registration will be ignored even on first one. But I think that this is a separate issue and should be handled in another PR.

Other question: 
Are there any plans to use router from transport-commons for express? Right now koa and socket transports use the router, but not express, so this change will be useful for them but not for express.

daffl:
You're right, the service options should be a different PR/issue. 

The problem with changing the router in Express is that things like custom service middleware (and any other middleware matching Express routes which are not compatible with this router) won't work anymore.

PR #2484, opened 2021-10-27
feat(http): Add `redirect` property to hook context

vonagam:
Adds standard way to specify a redirect location.

Before it was ok (not perfect, but ok) to do it with an express middleware but since there is now koa transport (and potentially in the future there will be other bare bone http one) it would be more beneficial to have a proper way for common functionality. 

To do it consistently had to add `encodeurl` dependency to `transport-commons` - both `koa` and `express` use it to escape location, but `koa` does not have `.location()` function (unlike express), only `.redirect()` which also does bunch of other stuff with body. 

If `statusCode` is not specified and `redirect` is present, then status defaults to 303. Default should be either that or 302 (former says to use GET request, later should be for temporary redirects with same request method).

Made code in rest between `koa` and `express` a little more consistent.

Also since a hook manager can theoretically (if overwritten) return a different context from one that was passed in to `initializeContext` have to set `res.hook`/`ctx.hook` twice - first when `base` is created (in case a method throws), second after a final context returned. (Not that anyone would do it, just a pedantic change.)

Still need to add some tests.

### DIFF
--- a/packages/express/src/rest.ts
+++ b/packages/express/src/rest.ts
@@ -40,11 +40,16 @@ export const serviceMiddleware = (callback: ServiceCallback) =>
       const { query, body: data } = req;
       const { __feathersId: id = null, ...route } = req.params;
       const params = { query, route, ...req.feathers };
+
       const context = await callback(req, res, { id, data, params });
+
       const result = http.getData(context);
+      const statusCode = http.getStatusCode(context, result);
+      const location = http.getLocation(context, req.get('Referrer'));
 
       res.data = result;
-      res.status(http.getStatusCode(context, result));
+      res.status(statusCode);
+      if (location) res.set('Location', location);
 
       next();
     } catch (error: any) {
@@ -66,11 +71,13 @@ export const serviceMethodHandler = (
   }
 
   const args = getArgs(options);
-  const context = createContext(service, method);
+  const contextBase = createContext(service, method);
+  res.hook = contextBase;
 
+  const context = await service[method](...args, contextBase);
   res.hook = context;
 
-  return service[method](...args, context);
+  return context;
 });
 
 export function rest (handler: RequestHandler = formatter) {
--- a/packages/feathers/src/declarations.ts
+++ b/packages/feathers/src/declarations.ts
@@ -313,6 +313,10 @@ export interface HookContext<A = Application, S = any> extends BaseHookContext<S
    */
   statusCode?: number;
   /**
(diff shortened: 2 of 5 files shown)


daffl:
Thank you for putting that up! I'm a little torn of adding HTTP specific things (this could be done in a Koa middleware as well right?) into the hook context although granted this has already snuck in with the `statusCode` property. The question would be how this should behave with different transports (like websockets). With `statusCode` e.g. a websocket connection can just ignore it but I'm not sure what should happen if `redirect` is et.

vonagam:
Yeah, it could be implemented in a custom middleware. But if you create a library and don't know which transport end-user will have then you will need to implement a custom middleware for each http transport (before there was only one official, now there are two, in future there can be more). And there will be no standardization, meaning that if some other library or user himself want to work with redirects then it might lead to conflicts between solutions.

For me redirects seem like a really essential part of http protocol that it is strange to see no way to specify them in feathers. 

With websocket `redirect` should be ignored just like `statusCode`. Ideally it might have made sense to move `context.statusCode` into `context.http.status` and `context.redirect` into `context.http.redirect` (and maybe add other http stuff, like ability to set response headers), but it would be a breaking change now and as i understand you want to keep migration as smooth as possible.

daffl:
I was actually thinking exactly the same thing with `context.http` when reading your comment. I think we can still make the `statusCode` part backwards compatible (e.g. `const staus = context.statusCode || context.http.status || 200`). Should we do that instead? Then it would be easy to add something like `context.http.reponseHeaders` as well.

vonagam:
To make `statusCode` backwards compatible I think it makes sense to implement it as getter/setter for `http.status`, so that end user don't have to do `context.statusCode || context.http.status` and avoid situations when values go out of sync with different hooks using different approaches. It is also possible to add a depreciation message in getter/setter about move to `http.status` and removal of root `statusCode` in the distant future.

daffl:
https://github.com/feathersjs/feathers/pull/2496 has been merged so we probably want to update this one accordingly.

vonagam:
Once`http.responseHeaders` was mentioned the need for `redirect` property vanished since a redirect is just a combination of a status and a header. Redirect functionality in feathers can be exposed (if there is interest in it) as some utility function.

So I rebased #2486 and after (if) it is merged `responseHeaders` can be introduced. Then this PR can be converted to an addition of a utility function or closed.

vonagam:
#2524 is merged and this redirect functionality should be supported now with setting `context.http.location`.

PR #2488, opened 2021-11-14
Mongo like sort

mhillerstrom:
Fixed a problem with detection of wether to sort embedded objects or ordinary single-level sort. Added tests.


daffl:
This makes sense, thank you!

PR #2492, opened 2021-11-21
fix(adapter-commons): clean up in sort.ts and select function

vonagam:
After last merged PR looked into `sort.ts` and saw some things that can be cleaned up. Did so.

Also in `select` function there was [`push()`](https://github.com/feathersjs/feathers/blob/9c22f70a838cb6341775d91705a7527c8fc5590e/packages/adapter-commons/src/index.ts#L14) into `params.query.$select` if they were present and `otherFields` provided. Should be concated into new array instead.


daffl:
Great call. I copied the sorting from NeDB a while ago (this is also why it doesn't have that many tests) and this is definitely nicee. Using the `exports.` variable probably only still worked by accident.

PR #2496, opened 2021-11-25
feat(core): add `context.http` and move `statusCode` there

vonagam:
This is `context.http` from #2484.

Here it contains only `statusCode` (can be renamed to `status` or not, don't have an opinion one way or another so kept previous name), with the idea to add `responseHeaders` later (`redirect` is not needed since it will be covered by those two).

I used getter/setter to prevent having two sources of truth and things getting out of sync. Also ideally a depreciation message should be added to those accessors that will write a warning first time one of them is accessed (a helper function for such depreciations can be added to `@feathersjs/commons`).

### DIFF
--- a/packages/feathers/src/declarations.ts
+++ b/packages/feathers/src/declarations.ts
@@ -240,6 +240,14 @@ export interface Params {
   [key: string]: any; // (JL) not sure if we want this
 }
 
+export interface Http {
+  /**
+   * A writeable, optional property that allows to override the standard HTTP status
+   * code that should be returned.
+   */
+  statusCode?: number;
+}
+
 export interface HookContext<A = Application, S = any> extends BaseHookContext<ServiceGenericType<S>> {
   /**
    * A read only property that contains the Feathers application object. This can be used to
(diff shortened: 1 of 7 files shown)


packages/feathers/src/hooks/index.ts
```diff
@@ -80,7 +80,13 @@ export function hookMixin<A> (
         method,
         service,
         event: null,
-        type: null
+        type: null,
```
daffl, line 83: If we hadd `http: {}` here we won't have to add it in the `createContext` sections of the HTTP handlers.

packages/feathers/src/declarations.ts
```diff
@@ -310,8 +318,14 @@ export interface HookContext<A = Application, S = any> extends BaseHookContext<S
   /**
    * A writeable, optional property that allows to override the standard HTTP status
    * code that should be returned.
+   *
+   * @deprecated
```
daffl: I believe if we add a description here like `Use http.statusCode instead` it will show up accordingly in editors that support it.

daffl (commented):
This looks good, thank you! Just two small somments.

packages/feathers/src/declarations.ts
vonagam: Done.

packages/feathers/src/hooks/index.ts
vonagam: Just feels not right to add `http` to every context from the start, even internal ones.

Is there any problem with `createContext`? Seemed like a proper place/way to add a transport-specific property to a context.

daffl: Great point. Agreed.

BenPortner:
@daffl @vonagam this PR deprecates the `Context.statusCode` property and refers to `http.statusCode` instead. However, there is not http property for websocket transport. How are websocket statusCodes to be handled in the future?

daffl:
You can still set the status code, however it will be ignored for websockets since they have no effect. Feathers errors will still be converted with their respective error statuses though.

PR #2498, opened 2021-11-26
fix(express): missing express types for Request, Response

marshallswain:
Fixes this issue:
Screen Shot 2021-11-26 at 9 47 53 AM


daffl:
As I mentioned, we might want to look into why this didn't break the tests before. I _think_ it's because `Response` and `Request` are also global variables in the Web API. If so we should disable them here.

Issue #2500, opened 2021-11-28
Asynchronous app initialization and setup improvements


daffl:
PR https://github.com/feathersjs/feathers/pull/2255 adds an asyncronous `app.setup` and `service.setup` which can be used with new hooks. However, when working with @marshallswain we noticed that this is still not ideal when configuring your application. For example the MongoDB service needs the asynchronous model in the `configure` function *before* its `setup` method is called. We are now looking at adding the following improvements:

- `app.configure` should allow `async` functions. They will run in the order `app.configure` is called and only when the previous function returns. *If* there is any asynchronous `app.configure` function, `await app.setup()` or `app.listen()` must be called *before* any services (calling `app.service(name)`) can be used.
- As an additional improvement and to avoid having the app in an unexpected state (e.g. in tests), if *any* service implements a `setup` method,  `await app.setup()` or `app.listen()` must also be called *before* any services (calling `app.service(name)`) can be used.

Both changes will work compatible with existing applications, specifically we wanted to avoid having to unnecessarily call `app.setup` on client side Feathers apps.

Related to https://github.com/feathersjs/feathers/issues/509 and https://github.com/feathersjs/feathers/issues/853

marshallswain:
Makes sense to me.  To clarify, "using a service" means calling `app.service('service-name')` to get a reference to the service, right?  So calling that *before* all async setup completes will throw an error?

daffl:
Correct. I updated the comment. We should be able to make the error message fairly descriptive. In the docs and generated apps we recommend getting the reference as late as possible (e.g. in a specific test instead of storing it early),

vonagam:
Change to `app.configure` sounds strange. Right now it is no more than a pretty way to call a function. Which means that there is no difference and expectations are clear.

Now that code will become improper:
```ts
app.configure(someConfigure);
app.configure(changeFooSetting);
console.log(app.settings.foo);
```
The outputted result will depend on whenever `someConfigure` (which is unrelated to `changeFooSetting`) is async or not.

Though I understand the need for a way to add async setup steps (wrote and use a simple utility of my own for that). I would recommend to introduce a new app method (`addSetup` or something) which will collect funcs for running during setup. So there will be clear separation between simple sugar that runs immediately and setup funcs that execute later.

As for mandatory `await app.setup()` - why *if*s? Why not just make it a requirement?

daffl:
True, having different behavious based on what kind of function is passed (which you may not even know) makes it quite unpredictable.

[...]

PR #2510, opened 2021-12-13
feat(core): add `app.addSetup` for asynchronous configuration steps

vonagam:
Connected to #2500 and contains small fixes from #1611.

Adds `app.addSetup()` which allows adding asynchronous configuration/initialization steps.  
Those will run with services setups during `app.setup()` call. 

Services use this method to add their own setup calls to queue,
so one can add a step that runs before or after a specific service setup.

This is for cases when you don't have an appropriate service to implement `setup` method in.

Will add tests if that sounds as a reasonable addition.

### DIFF
--- a/packages/feathers/src/application.ts
+++ b/packages/feathers/src/application.ts
@@ -15,7 +15,8 @@ import {
   HookOptions,
   FeathersService,
   HookMap,
-  RegularHookMap
+  RegularHookMap,
+  Setup
 } from './declarations';
 import { enableRegularHooks } from './hooks/regular';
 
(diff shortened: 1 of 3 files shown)


marshallswain:
I like this idea.  It nicely solves the concerns you had in https://github.com/feathersjs/feathers/issues/2500#issuecomment-995094009.  Looks good to me.

vonagam:
I will close this one since recent hooks for `setup`/`teardown` from #2585 should do the trick. 

But will note that setting [`this._isSetup = true`](https://github.com/feathersjs/feathers/blob/a268f86da92a8ada14ed11ab456aac0a4bba5bb0/packages/feathers/src/application.ts#L161) should be done before services begin to register, not after, otherwise dynamically registered services from some `setup` will not get their own `setup` called. (It was fixed in #1611, but then #2585 put it back because it was based on a code before the fix.)

daffl:
Ah yes, I meant to get back to ask if that new one may not cover everything but I'm hoping it will suffice. Ugh, I thought I kept your `_isSetup` change in there because that made sense. Will have to revert that part.

PR #2511, opened 2021-12-18
Draft: move from hygen to ts-template-strings generator

fratzinger:
### Introduction

Discussed on slack before. This is based on an idea of @daffl to move from `hygen` to a new generator based on ts & template strings. This was the initial sketch from @daffl, which is the base of this new generator:

```ts
import { GeneratorContext, Prompter } from '@feathersjs/cli'
import { join } from 'path'

type Variables {
  name: string
}

export async function prompt (prompter: Prompter) {
  return prompter.prompt([{
    type: 'input',
    name: 'name',
    message: 'What is the name of your app?'
  }])
}

export async function render (context: GeneratorContext<Variables>) {
  const to = join(context.lib, 'app.ts')
  const body =
`
// This is the ${context.name} app
import feathers from '@feathersjs/feathers'

const app = feathers()

export default app
`

  return { body, to }
}
```

Related: #2414, #2425

### Summary

Proof of concept. Moved from `hygen` with ejs to new custom generator based on ts & template strings. This is 'just' a refactor. Globally spoken, nothing added or changed. It was fairly easy and has a lot in common with hygen. I enjoyed the template strings very much!

I think this could be something!

### Things changed:
- added almost all files from [`@feathersjs/hygen`](https://github.com/feathersjs/hygen) to this repo, and tweaked obviously. (see folder `nextgen`)
- changed all template files to new `ts` files with render function
- front-matter attributes become exported members of the `render` function (see type `RenderAttributes`)
- `prompt.js` and `index.js` files for `hygen` became `prompt.ts` and `index.ts`
- changed (compared to `@feathersjs/hygen` lazy `require()` to lazy `await import()`
- removed dependencies `@feathersjs/hygen`
- removed (compared to `@feathersjs/hygen`): dependencies `ejs`, `front-matter`
- changed (compared to `@feathersjs/hygen`): moved from `chalk` to `picocolors` and `colorize-template` cause perf

### Benefits:
- full typescript support. type suggestion, auto complete, type checking

### Downsides:
- no code formatting in template strings
- escape template strings are little bit annoying:
```ts
const body = `
app.listen(port).then(() => {
  logger.info(\`Feathers app listening on http://\${host}:\${port}\`);
});
`
```

### Things to be done:
- [ ] There is a bug while generating the custom service in the tests. I looked at it for a while but cannot find the problem. It has to be solved, but doesn't change the proof of concept for now.

### Discussion:

David, what do you think? Where to go from this? Do you think it's worth investigating?

I have some thoughts for it and am happy to discuss them:
1. Make it a separate (new) generator (independent from @feathers/cli - like hygen is). Good idea?
2. Add tests for the new generator
3. docs for the generator (new generator and @feathersjs/cli)
4. deno?
5. improve types


daffl:
This is amazing! I'll have a closer look over the holidays but this looks like it could totally work.

daffl:
This has been finalized in the first version via https://github.com/feathersjs/feathers/pull/2578

Issue #2514, opened 2021-12-22
OAuth redirect lost sometimes due to session store race


mterrel:
### Steps to reproduce

1. Initiate an OAuth login with the `redirect` query parameter set, for example to `https://example.com/oauth/github?redirect=dashboard`
2. Go through the OAuth flow to authorize the user.
3. At the end of the flow, the browser should eventually be redirected to `https://example.com/dashboard#access_token=xyz123`. This happens correctly **most** of the time. But occasionally, the browser is redirected to `https://example.com#access_token=xyz123` instead.

Note that this behavior is difficult to reproduce and appears to require specific timing of external resources (the session store needs to be slow) to trigger the failure.

### Expected behavior

At the end of the OAuth flow, the browser should get redirected to the path requested by the initial `redirect` query parameter 100% of the time.

### Actual behavior

At the end of the OAuth flow, the browser gets redirected to the root of the site about 5% of the time (about 1 of every 20 requests redirects incorrectly).

### System configuration

Tell us about the applicable parts of your setup.

**Module versions** (especially the part that's not working):

Feathers 4.5.x:
```
├─ @feathersjs/adapter-commons@4.5.2
├─ @feathersjs/authentication-client@4.5.8
├─ @feathersjs/authentication-local@4.5.4
├─ @feathersjs/authentication-oauth@4.5.4
├─ @feathersjs/authentication@4.5.8
├─ @feathersjs/commons@4.5.11
├─ @feathersjs/configuration@4.5.3
├─ @feathersjs/errors@4.5.11
├─ @feathersjs/express@4.5.4
├─ @feathersjs/feathers@4.5.8
├─ @feathersjs/rest-client@4.5.11
├─ @feathersjs/socketio-client@4.5.11
├─ @feathersjs/socketio@4.5.11
├─ @feathersjs/transport-commons@4.5.11
├─ feathers-hooks-common@5.0.3
└─ feathers-sequelize@6.2.0
├─ connect-session-sequelize@7.1.1
└─ express-session@1.17.2
```

**NodeJS version**:

14.18

**Operating System**:

Linux

**Browser Version**:

Chrome 96.0.4664.110 / Win10

**React Native Version**:
N/A

**Module Loader**:
None

## Analysis

[...]

PR #2515, opened 2021-12-23
OAuth redirect lost sometimes due to session store race (#2514)

mterrel:
### Summary

This PR solves the issue (described in greater detail in #2514) where Feathers OAuth does not always redirect to the path requested by the `redirect` query parameter and instead redirects to the root of the site specified by `authentication.oauth.redirect`. This happens only a small percentage of the time and only when using an external Express session store that is somewhat slow.

The issue is due to a race condition where the Express session created during a request to `/oauth/:name` does not get saved before the browser makes a subsequent request in the OAuth flow (to `/oauth/connect/:name`)  and that causes Express to create a new session, effectively discarding the original session which contains the `redirect` parameter.

The fix is for Feathers to explicitly save the Express session in the `/oauth/:name` handler before responding with the redirect to `/oauth/connect/:name`.

This fix is consistent with guidance from Express session's [docs](https://github.com/expressjs/session#sessionsavecallback) (see bolded phrase):

> **Session.save(callback)**
> Save the session back to the store, replacing the contents on the store with the contents in memory (though a store may do something else--consult the store's documentation for exact behavior).
> This method is automatically called at the end of the HTTP response if the session data has been altered (though this behavior can be altered with various options in the middleware constructor). Because of this, typically this method does not need to be called.
> **There are some cases where it is useful to call this method, for example, redirects,** long-lived requests or in WebSockets.
 
Also see issues like https://github.com/expressjs/session/issues/660

Fixes: #2514 

### Other Information

**Tests run**

- `npm run test`

**Note on TypeScript version**

The newer version of TypeScript that gets installed with a fresh clone of this repo does not compile successfully due to TS changing the default type of the error in catch blocks to `unknown`. So I ran my tests by pinning TypeScript to an older version.

### DIFF
--- a/packages/authentication-oauth/src/express.ts
+++ b/packages/authentication-oauth/src/express.ts
@@ -57,7 +57,13 @@ export default (options: OauthSetupSettings) => {
       req.session.redirect = redirect as string;
       req.session.query = query;
 
-      res.redirect(`${path}/connect/${name}?${qs.stringify(query as any)}`);
+      req.session.save((err: any) => {
+        if (err) {
+          res.status(500).send(`Error storing session: ${err}`);
+        } else {
+          res.redirect(`${path}/connect/${name}?${qs.stringify(query as any)}`);
+        }
+      });
     });
 
     authApp.get('/:name/callback', (req: any, res: any) => {


mterrel:
@daffl In order to get CI to run, you may want to look at adding `"useUnknownInCatchVariables": false` to `tsconfig.json`. Or, alternately, you'll need to go touch pretty much all your catch blocks on the crow branch. Let me know if I can be of help.

daffl:
Released in 4.5.12. Thank you for digging into this.

PR #2524, opened 2022-01-09
feat(transport-commons): add `context.http.response`

vonagam:
Allows to specify any response headers for http transports.
Related to #2484.

Not sure about naming though. About `response` part.
If `http` is only about outputs (and any inputs go into `params`) then there is no need to specify that this is response headers.
If `http` can contain something besides outputs, then status part should be maybe renamed to `responseStatus` for consistency.

I think the first option is more attractive to me with short names - so `status`, `headers` and possibly `location` and `cookies` (those are headers too, but can be beneficial to add easier way to set them).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,6 +13,7 @@
 				"@types/axios": "^0.14.0",
 				"@types/bcryptjs": "^2.4.2",
 				"@types/config": "^0.0.41",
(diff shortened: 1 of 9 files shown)


daffl:
I think this one will work great. Since this allows for redirects, can we close #2484?

vonagam:
Seems so. Will do.

PR #2545, opened 2022-01-27
fix(schema): Allow compound oneOf for queryProperty

marshallswain:
Currently we use `oneOf` to create the `queryProperty` validations.  This works fine until you nest it inside of another `oneOf`.  In this case it fails every time.  I've switched it to `anyOf` and have also added two tests to make sure it's both passing and failing correctly.


marshallswain:
Also, I should note that if we don't change to `anyOf`, the first test I added would fail.

PR #2585, opened 2022-03-29
feat(core): Add app.setup and app.teardown hook support

daffl:
This pull request adds built-in support for application `setup` and `teardown` hooks. They are special hooks that don't run on the service level but instead directly on `app.setup` and `app.teardown`

```ts
app.hooks({
  setup: [ mongodbConnection ],
  teardown: [ closeMongoConnection ]
});
```


marshallswain:
:shipit: 🚀

Issue #2522, opened 2022-01-07
Feature Request: Server Bundle, like @feathersjs/client


FossPrime:
### Steps to reproduce

A typical dove deployment will have the following dependencies
```
   "@feathersjs/authentication": "5.0.0-pre.15",
    "@feathersjs/configuration": "^5.0.0-pre.15",
    "@feathersjs/errors": "^5.0.0-pre.15",
    "@feathersjs/feathers": "^5.0.0-pre.15",
    "@feathersjs/socketio": "^5.0.0-pre.15",
    "@feathersjs/transport-commons": "^5.0.0-pre.15",
```

This adds a lot of unnecessary noise to package.json in an environment where storage isn't critical.

To ease developer mental load and improve appearances, I propose doing the same thing we do for @feathersjs/client and merging the above package into:
```
   "@feathersjs/server": "5.0.0-pre.15"
```

This might also be a solution to the version locked weirdness, where 9 modules look like they need updates and QA, when only one changed. I get the idea for major versions, but it seems silly for minor ones.

### Expected behavior
A modern framework that's easy to update and needs minimal to no additional dependencies for basic apps.

### Actual behavior
Bare-bones projects require the developer to manage over a dozen dependencies and their **perceived** inter-compatibility woes.

### System configuration

**Module versions**: Dove 5.0.0-pre.15  
**NodeJS version**: 16 LTS  
**Operating System**: Debian  
**Module Loader**: NPM 8.10

FossPrime:
superseeded

PR #2586, opened 2022-03-30
Document the need for third-party dependencies

FossPrime:
Document the need for third-party dependencies in ReadMe. Some of them are not obvious and seem a little too lazy, like install and shx, whose use can typically be easily avoided.

I've also found feathers-sequelize getting installed by the generator for seemingly no reason.

This is closely related to #2522... I encountered a couple mystery dependencies while merging a 3 month old Dove server to a vite project without using monorepo tools... being clear about why dependencies are there really matters in that situation, especially when you're battling code rot with TypeScript/vite.


daffl:
Hm, I can see this being useful somewhere in the docs but I'd like to keep the readme short with just a reference to the docs and the generator. Thank you for the pull request though!

PR #2587, opened 2022-03-30
chore: Fix all package types fields and unify test command

FossPrime:
The NPM package page thinks this package is missing types... it's not. This PR makes it show the TS symbol next to the name. 

https://www.npmjs.com/package/@feathersjs/koa

Should have the same TS symbol that the main package does...
Screenshot 2022-03-30 11 17 14


daffl:
Whoops, thanks! I think I at least forgot it in `schema` and `memory` as well. Would you mind checking all the `package.json` and add it anywhere else it may have been missed?

PR #2594, opened 2022-04-07
fix(schema): result resolver correctly resolves paginated find result

idaho:
### Summary

This pull request will resolves an issue with the resultResolver in combination with a paginated services.

Currently the schema resultResolver resolves the result on context.data. This will end up in an issue if you have a service configured with pagination enabled. 

Before the resultResolver is running we have the properties `total`, `limit`, `skip` and `data` on `context.result`. After all properties are resolved, the result of it will set to `context.result`. That means, that the paginated result structure gets los.


daffl:
Great, thank you!

PR #2607, opened 2022-04-24
fix(adapter-commons): Clarify adapter query filtering

### LINKED ISSUE #1161: About filterQuery
In `filterQuery` we have `operators` and `filters`. They are kinda on equal ground. `operators` work on `query` part of result, and `filters` define `filters` part.

But in `AdapterService` we see inequality in treatment of those two. [Here](https://github.com/feathersjs/feathers/blob/96ad17714c0f01adfe8816bb78a217f8a0e9cc03/packages/adapter-commons/lib/service.js#L42) we see how `operators` value comes from `options.whitelist` (and they have unneeded default value), while `filters` come from `options.filters`. And while they are both actually lists of additional keys, naming moves focus to `whitelist/operators/query` part for adapter developers.

Result of this focus you can see, for example, in `feathers-mongoose`. `$populate` is a filter type (as it was before changes), but now it is told to go in `whitelist` and, so, to `query` part. It results in those manual removals of `$populate` from `query` ([1](https://github.com/feathersjs-ecosystem/feathers-mongoose/blob/2f86c1d888caef81b4500d802253e37a5902589a/lib/service.js#L46), [2](https://github.com/feathersjs-ecosystem/feathers-mongoose/blob/2f86c1d888caef81b4500d802253e37a5902589a/lib/service.js#L114), [3](https://github.com/feathersjs-ecosystem/feathers-mongoose/blob/2f86c1d888caef81b4500d802253e37a5902589a/lib/service.js#L189), [4](https://github.com/feathersjs-ecosystem/feathers-mongoose/blob/2f86c1d888caef81b4500d802253e37a5902589a/lib/service.js#L265)), which would not be needed if instead of `whitelist`, `$populate` was expected in `filters`.

Ideally, i think options should have been named `operatorsWhitelist/filtersWhitelist`, `additionalOperators/additionalFilters`, `whitelistedOperators/whitelistedFilters` or something like this. Plus it should be clearly stated somewhere for adapter developers what is the difference between a filter and an operator: the most simple rule is if something has to be at root level and can't be nested in `$or` or `$and` (like `$populate` for example) then it is a filter, otherwise operator.

For starters, we can add aliases to `whitelist/filters` pair and in next major release remove `whitelist/filters`.

I know that this is not too important, but still it creates some inconvenience when you can't rely on `service.filterQuery` to really separate query from filters.

What are your thoughts?

### LINKED ISSUE #2389: [adapter-commons] filters and operators
Hi,

Continuing the discussion from Slack

<details>
 <summary>Conversation from Slack</summary>

```js
Hey @daffl, I see that the filterQuery supports additional filters, and operators but how do I go about using this? https://github.com/feathersjs/feathers/blob/dove/packages/adapter-commons/src/filter-query.ts#L103

I have a query that is

// can be as a options to ignore all fields
service.find({ query: { name: 'POST', description: 'hello', $ignoreCase: true } });
// or can be per field
service.find({ query: { name: { $eq: 'POST', $ignoreCase: true }, description: 'hello' } });

Is there a way for filterQuery to extract the split out to filters and so on correct by passing in the additional options ? I hope there is at least ways pass in as the options (first query above) with filterQuery splitting out as filters/operators. as for per field, I think I can still do post-processing after filterQuery split out the query

this question is for database service adapter..

Thanks

---

I think I got it working for the filters

const { filters, paginate, query } = this.filterQuery(params, { filters: ['$ignoreCase'] });

Initially, I was confused because I saw that you set the default as {} https://github.com/feathersjs/feathers/blob/3f4db68d6700c7d9023ecd17d0d39893f75a19fd/packages/adapter-commons/src/filter-query.ts#L105

So I'm not sure how to pass it in.. So I just tried to pass in as a array, and it seem to work but not sure if that is correct. If you could let me know that would be great! As I'm not quite sure why additionalFilters is default to {} and additionalOperators is default to []

---

And it seem like this is all I need for to support both filters and operators

for as a option, and as individual field

const { filters, paginate, query } = this.filterQuery(params, { filters: ['$ignoreCase'], operators: ['$eq', '$ignoreCase'] });

---

ok, I'm probably thinking too complex, for additional operators. can just pass in when init the service with whitelist options. So it's not the adapter that needs to allow this explicitly.

But I still need to pass in $ignoreCase as the additional filters since this can't be passed in via the service options.

---

Daffl

I think some of this is actually broken or at least not working very well (e.g. discussion in https://github.com/feathersjs/feathers/issues/1971) in what order these are happening, what specifically the difference between operators and filters is etc.

So your questions are totally valid. Maybe we should move this into an issue to see what can be done to improve the semantics and API for v5.
```
</details>

## Summary

This issue is those who use `filterQuery` from `adapter-commons` package. And I assume, mostly used by `database-adapter-service package` (e.g `feathers-mongoose`).

`filterQuery` provides a way to extract out the known `operators and filters` and also allows to pass in additional filters and operators. For example, we can pass in `whitelist` in the service options to the service to allow that service to accept the database-specific operators. But there is no way to pass in additional `filters` from the `service options`. Hence, if the `database-adapter` does not support it either by allowing it explicitly in the adapter or provide a way to let the client pass in via `options` then the adapter could be considered a missing feature. Should that be the job of the adapter to provide the feature out of the box, or should it always allow the client to determine what to pass through?

There is also confusion around what `operators` and `filters` should be

## Related

https://github.com/feathersjs/feathers/issues/1971
https://github.com/feathersjs-ecosystem/feathers-mongoose/issues/391

---

## Problem Statement

I think the issue here would be:

1. To identify what does `filter` and `operator` really mean/is?
2. To provide a way to pass in additional `filters` and `operators` via service options so that the `adapter` can pass down to `filterQuery`
3. To identify if it's the responsibility of the client or database adapter to provide database-specific operators/filters functionality (see https://github.com/feathersjs/feathers/issues/1971)

---

`filters` and `operators` meant (roughly) the same thing right now, and I think there is a need to properly differentiate it. If we look at `.find` method in general, it probably would look like this

```js
.find(filters, options);
```

Some example/variation of find would look like this

```js
.find({ name: 'joseph'}, { limit: 10 });

.find({ name: { $in: ['joseph', 'david'] }, { limit: 10 });
```

So, `filters` are just a more general term to indicate how you like to search for your data set, and you can use `operators` to help you with that. And then we pass in the `options` to affect (think `sort, limit`) the final look (of the result) before returning back to the caller.

However, in `feathers` world, we pass in everything within `query` and `filterQuery` method will interpret and breakdown before passing to the actual driver API call through whichever ODM/ORM you are using (correct me if I'm wrong)

This is how we will write it in term of `feathers` query

```js
.find({ query: { name: { $in: ['joseph', 'david'] }, $limit: 10 });
```

Which breaks down to 

```js
query: { name: { $in: ['joseph', 'david'] }
filters: { $limit: 10 }
```

This is different from how I describe above where it would look like this

```js
filters: { name: { $in: ['joseph', 'david'] } // operators is $in
options: { $limit: 10 }
```

I guess you can have `operators` for both `filters` and `options` which translate to

```js
filters: { name: { $in: ['joseph', 'david'] } // operators is $in
options: { $limit: 10 } // operators is $limit
```

## Proposal

First, with regards to how the actual term and usage of `filters`, `operators`, or `options` will need further discussion to see how there could be a better way to term this or come to a common understanding of what it means at least to `Feathers Database Common API`. Since that (maybe) different ODM/ORM has different terminology of those, it might be hard to define one that is valid across all different types of ODM/ORM.

The `feathers database common API` can then describe what are the default list of `operators` it allows for both `filters` and `options` like the list that is in [querying](https://docs.feathersjs.com/api/databases/querying.html) right now

My suggestion is to call it `filters` and `options` but also keep the `operators` term.

```js
.find(filters, options)

.find({ query: { name: { $isNull: true }, $ignoreCase: true });

// filters
{ name: { $isNull: true }
// options
{ $ignoreCase: true }
```

`operators` can be classify as `filterOperators` and `optionOperators`

```js
// service options
const options = {
  optionOperators: ['$ignoreCase'],
  filterOperators: ['$isMissing', '$inNull'],
};
```

Second, to allow to pass in `additionalFilters` via the service options, because, as far as I understand, `whitelist` in the service options is only meant for `Operators` and not `Filters`

```js
const options = {
  Model: getModel('posts'),
  // could be `adapter` to be generic
  ottoman: {
    optionOperators: ['$ignoreCase'],
    filterOperators: ['$inMissing', '$inNull'],
    lean: true,
    consistency: SearchConsistency.GLOBAL,
  },
  events: ['testing'],
};
```

However, doing so has a disadvantage which is putting the responsibility on the client, and where not all clients are advanced users and know the different operators/filters to allow/disallow via the service options. On the other hand, if the client needs to use such `operator/filter`, then it's not difficult for the client to just allow it via the service options rather than the adapter providing it out of the box. So long the documentation states clearly.

## Bonus

I was wondering if there is a way to standardize how `database-adapter` accept options when initializing service.

For example, in [feathers-ottoman](), 

```js
const options = {
  Model: getModel('posts'),
  // could be `adapter` to be generic
  ottoman: {
    lean: true,
    consistency: SearchConsistency.GLOBAL,
  },
  events: ['testing'],
};
```

Any `adapter` specifics options to go into `ottoman` object. This way, it makes it easier to identify/differentiate the `options` for `feathers-service` and for the `adapters`

---

Not sure if this is more of a discussion, if so, do move over to discussion and we can continue from there.

Let me know if any points are unclear, and I can clear those up, or re-words those.

### LINKED ISSUE #2445: [Suggestion] Apply operator whitelist option only for "external" provider by default
Could be possible in the next version `5.x` or current, to block all non-standard query operators for "external" provider and allow internally?

Migration could be frustrating if on API you use non-standard query operators internally (not client-side). 🥵 

I understand that `params` on `find`, `get`, `create`, `patch` and `remove`, receive the `provider` key can easily identify if is secure allow non-whitelisted query operators.

### LINKED ISSUE #2546: [Suggestion] ability to allow `multi` queries for internal calls only
Currently by default, a query like
```js
await context.app.service('profiles').remove(null, {
  query: {
    user_id: context.id,
  },
});
```
won't work unless `profiles` service has `multi` option allowing multi delete. Yet, having that allowed, Feathers "authomatically" allows for multi delete *external* call on profiles, which is suboptimal. I think it could be useful if there were not only "true" and "false" options for `multi`, but also either a list of providers allowed for `multi`, or some other option to allow for internal `multi` call but disallow external one.

One could use feathers-hooks-common `disable-multi-item-change` conditionally with `iff(isProvider(...))`, but [`disable-multi-item-change` is deprecated](https://hooks-common.feathersjs.com/migrating.html#deprecations) since feathers-hooks-common@5.0.0.

### LINKED ISSUE #2550: Suggestion: Have whitelist_client and whitelist_server instead of whitelist
I'm new to feathers, so my suggestion may be the same as https://github.com/feathersjs/feathers/issues/2445

Please close this request if this is the case.

I need to use  $populate on the server. I do not want to allow the client to send a query using $populate.

As it is, I need to whitelist $populate in the service so I can use it on the server.  This also allows the client to send a query with $populate.  

Extra coding has to be done to ensure the client does not send a $populate parameter. I would like to easily not allow the client to use $populate, but still use it on the server.

daffl:
This pull request tries to address some inconsistencies and flexibility issues when implementing and using database adapter.

# The problem

There are several issues that mention a lack of clarity when using the previous `filterQuery` method. What is a `filter` and what isn't? How are they customized? Another recurring problem is that you may not want strict query filtering when using a database adapter on the server.

# Improvements

## Remove `filterQuery` and add `sanitizeQuery` and `sanitizeData`

This pull request removes the `filterQuery` adapter base service method and adds a `sanitizeQuery` and `sanitizeData` method instead. `sanitizeQuery` applies the previous query filters but then returns them as a single query object and it is up to the adapter implementation what to do with that. `sanitizeData` does nothing by default but could be used to e.g. filter out protected fields (or things like `$push` for MongoDB).

## Add `$` prefixed *unsafe* internal service methods

It also changes the semantics of the `_find`, `_get`, `_patch`, `_update` and `_remove` internal (hook-less) service methods to sanitize the query and data and check if multi updates are allowed. It has been moved from the service methods so that it is easier to extend from a database adapter and implement the service methods you need instead of being limited by a type interface (also see https://github.com/feathersjs/feathers/pull/2605).

For better flexibility, it additionally introduces the `$find`, `$get`, `$patch`, `$update` and `$remove` *unsafe internal service methods*. This is what a database adapter now implements and it shouldn't do any query or data sanitization. Calling these methods directly lets you do any query the database adapter supports without having to explicitly declare allowed operators, filters or `multi` settings.

## Clarify `filters` and `operators`

Finally all parts of this PR follow a more strict definition of `filters` and `operators`:

- A `filter` is a special top level query property starting with a `$`. This can be used for pagination (`$limit`, `$skip`), modifying the result set (e.g. `{ $populate: true }` etc.
- An `operator` is a special property starting with a `$` that can be used to query properties, e.g. `{ age: { $gte: 21 } }`

This definition has also been reflected in the [Querying API documentation](https://dove.docs.feathersjs.com/api/databases/querying.html).

The adapter settings have been updated and can now be used like this:

```js
service({
  filters: {
    $populate: true,
    $ignoreCase: value => value === 'true' || value === true ? true : false
  },
  operators: [ '$eq' ] // allows `{ name: { $eq: 'David' } }`
});
```

Ideally I'd like to see a lot of this responsibility moving to using a query schema and resolvers instead which can handle all those complexities (type coercion, setting dynamic properties etc.) much more cleanly and securely but this is hopefully a step in a better direction.

- Closes https://github.com/feathersjs/feathers/issues/1161
- Closes https://github.com/feathersjs/feathers/issues/2389
- Closes https://github.com/feathersjs/feathers/issues/2445
- Closes https://github.com/feathersjs/feathers/issues/2546
- Closes https://github.com/feathersjs/feathers/issues/2550


matiaslopezd:
ref [#1161 ](https://github.com/feathersjs/feathers/issues/1161#issuecomment-1108957504)

PR #2610, opened 2022-04-24
feat(mongodb): Add and update feathers-mongodb as @feathersjs/mongodb

daffl:
This pull request converts https://github.com/feathersjs-ecosystem/feathers-mongodb to `@feathersjs/mongodb` and pulls it into core.

Since the new `@feathersjs/schema`, especially in combination with Feathers services, covers essentially most things a traditional ODM (in this case specifically Mongoose) or ORM would do (but with more flexibility and likely better performance since it works with plain objects instead of model instances and can be used in combination with existing Feathers caching and dataloading plugins), it is a good time to integrate the more low level database adapters closer into Feathers core.

They will provide the default integrations for databases in the CLI and should be maintained together with everything else (including documentation). SQL databases will be supported through a migrated https://github.com/feathersjs-ecosystem/feathers-knex

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -34,7 +34,7 @@ plugins:
   duplication:
     enabled: true
     config:
-      count_threshold: 3
+      count_threshold: 4
 exclude_patterns:
   - "**/test/*"	
   - "**/adapter-tests/*"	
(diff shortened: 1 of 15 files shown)


noor-tg:
👍🏼

PR #2614, opened 2022-04-28
fix(authentication-oauth): Make oAuth authentication work with cookie-session

cciocov:
This PR make authentication-oauth module work with cookie-session instead of express-session. The former does not have a save() or destroy() method, so this PR adds a check before trying to call req.session.save(). It also fixes a bug with a previous PR where a call to req.session.destroy() would still be made, even if there is no such method.


daffl:
Thank you! Will go out with the next prerelease :smile:

PR #2617, opened 2022-05-02
fix(authentication-oauth): Don't send origins in Grant's config

cciocov:
Don't send "origins" (similar to how we don't send "redirect") in Grant's config, as it will be considered another auth provider by Grant.


daffl:
Thank you for for fixing those things!

Issue #2630, opened 2022-05-10
For version 5.0.0-pre.20, authentication callback and redirect_uri are incorrect


johnheenan:
A fix has been proposed. The problem was not the redirect url, which is for the final phase back to the user browser. The problem was the callback and redirect_uri for authentication providerss

In version 4.5.11 the following `config/default.json` works fine for the specified redirect:

```
{
  "host": "localhost",
  "port": 3030,
  ...
  },
  "ts": true,
  "authentication": {
    "oauth": {
      "redirect": "http://localhost:3000",

```

In version 5.0.0-pre.20, the redirect is taken from host and port.

Tried adding v5 "origins" without "redirect". Same problem

The node server is behind a nginx proxy server

John Heenan

johnheenan:
Some more information. I took a build of code for http://localhost:3000 and placed it on the same server as the version 5.0.0-pre.20 feathersjs server, with "redirect" left at "/". This allows the reverse nginx proxy server to apply a generic `proxy_redirect default` and narrows down the scope.

## Using GitHub for authentication

An attempted login with github just redirects to home page with feathersjs messages including
```
express:router <anonymous>  : /oauth/github/callback?error=redirect_uri_mismatch&error_description=The+redirect_uri+MUST+match+the+registered+callback+URL+for+this+application.&error_uri=https%3A%2F%2Fdocs.github.com%2Fapps%2Fmanaging-oauth-apps%2Ftroubleshooting-authorization-request-errors%2F%23redirect-uri-mismatch +0ms

express:router session  : /oauth/github/authenticate +0ms
  express-session fetching 30Qd5ebzb6QdItunVClTedMOmXIPP7VS +50ms
  express-session session found +0ms
  knex:client acquired connection from pool: __knexUid4 +1m
  knex:bindings [ undefined ] +1m
  knex:client releasing connection to pool: __knexUid4 +0ms
  @feathersjs/errors GeneralError(500): Undefined binding(s) detected when compiling SELECT. Undefined column(s): [githubId] query: select count(`users`.`id`) as `total` from `users` where `githubId` = ? +1m
```

## Using Google for authentication

An attempted login with google resulted in the following redirect from google to what should be feathersjs, which shows what is being passed to oauth providers. Providers should not be told to use http://localhost:3030/.
[...]

PR #2631, opened 2022-05-11
Fix authentication-oauth regression using incorrect callback and redirect_uri

johnheenan:
### Summary

Please see https://github.com/feathersjs/feathers/issues/2630

Call to defaultsDeep parameters reordered to ensure protocol and host values from both both oauth.defaults and from local variable can be distinguished, with local values as fallback.

Previous version was coding local values instead of oauth.default values. This matters if they are not the same, such as when using reverse proxy server. The local host value is from top host value of configuration file defualt.json.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1810,9 +1810,9 @@
 			}
 		},
 		"node_modules/@jridgewell/set-array": {
-			"version": "1.1.0",
-			"resolved": "https://registry.npmjs.org/@jridgewell/set-array/-/set-array-1.1.0.tgz",
-			"integrity": "sha512-SfJxIxNVYLTsKwzB3MoOQ1yxf4w/E6MdkvTgrgAt1bfxjSrLUoHMKrDOykwN14q65waezZIdqDneUIPh4/sKxg==",
+			"version": "1.1.1",
+			"resolved": "https://registry.npmjs.org/@jridgewell/set-array/-/set-array-1.1.1.tgz",
+			"integrity": "sha512-Ct5MqZkLGEXTVmQYbGtx9SVqD2fqwvdubdps5D3djjAkgkKwT918VNOz65pEHFaYTeWcukmJmH5SwsA9Tn2ObQ==",
 			"engines": {
 				"node": ">=6.0.0"
 			}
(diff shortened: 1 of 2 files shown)


johnheenan:
I should let you know that I tried the following with oauth last, instead of second following `{}` as in the existing solution, but it did not work. The results are the same as the existing solution. I accept the existing solution , with oath second should work.
```
const grant = defaultsDeep({}, {
    defaults: {
      prefix,
      origin: `${protocol}://${host}`,
      transport: 'session',
      response: ['tokens', 'raw', 'profile']
    }
  }, omit(oauth, ['redirect', 'origins']));
```
This is the suggested solution that works, but it should have an addition.
```
const grant = defaultsDeep({
    defaults: {
      prefix,
      origin: `${oauth?.defaults?.protocol ?? protocol}://${oauth?.defaults?.host ?? host}`,
      transport: 'session',
      response: ['tokens', 'raw', 'profile']
    }
  }, omit(oauth, ['redirect', 'origins']));
```
 The following is better to ensure the transport from the configuration is used. I will add another commit to the PR.
```
  const grant = defaultsDeep({
    defaults: {
      prefix,
      origin: `${oauth?.defaults?.protocol ?? protocol}://${oauth?.defaults?.host ?? host}`,
      transport:  oauth?.defaults?.transport ?? 'session',
      response: ['tokens', 'raw', 'profile']
    }
  }, omit(oauth, ['redirect', 'origins']));

```

daffl:
Thank you for the pull request! When comparing it to v4 I'm thinking the order is just wrong which is what you changed but then we can keep the old host settings. I'll quickly update and it will get out with the new release tomorrow and you can let me know if that fixes it.

johnheenan:
Thanks for the commit. No, changing the order does not fix the problem with my config/default.json. My fix does fix the problem with my config. I thoroughly tested your and my version, to the extent of checking the compiled module js used.

We need to step back a bit because we may be at cross purposes given the way I configure feathers for authentication. I need to state how I configure and why I configure this way. In short, I configure feathers on a 'need to know' basis and my configuration preserves this principle. I will put a sample configuration for discussion on the issue at https://github.com/feathersjs/feathers/issues/2630 so you and others can comment on this if they wish.

johnheenan:
Thanks for the commit. No, changing the order does not fix the problem with my config/default.json. My fix does fix the problem with my config. I thoroughly tested your and my version, to the extent of checking the compiled module js.

We need to step back a bit because we may be at cross purposes given the way I configure feathers for authentication. I need to state how I configure and why I configure this way. These reasons, with a sample configuration, are on https://github.com/feathersjs/feathers/issues/2630, so you and others can comment on this if they wish.

daffl:
No my bad, you were correct. I thought it would still fill in the correct settings. Will merge your original PR and put it out with the next release shortly.

cciocov:
@johnheenan @daffl 
This merge introduces a bug, in that you are no longer able to override the keys under "defaults". SInce we're using "defaultsDeep" instead of "merge", the order of parameters passed should be inverse (i.e.: higher priority parameters first, which are the ones coming in the "oauth" object).

Opened PR #2659 to fix this.

Issue #1809, opened 2020-02-14
 protect()-hook does not hide the protected field in realtime events when patching multiple records


Tr3ffel:
I'm currently building a feathersjs application where I need to patch multiple user inside some service method. Because `protect('password')` is set in the user hooks I would expect that the password would be hidden in the realtime events, which is not the case. (I'm using the mongoose feathers adapter)

### Steps to reproduce
users.hooks.js
```
...
after: {
    all: [ 
      // Make sure the password field is never sent to the client
      // Always must be the last hook
      protect('password')
    ],
....
```

users.service.js
```
...
const options = {
    Model,
    paginate,
    multi: true // allowed to patch multiple users
  };
  // Initialize our service with any options it requires
  app.use('/users', new Users(options, app));
...
```

some-service.class.js
```
exports.SomeService = class SomeService {
  setup(app) {
    this.app = app;
  }

  async create (data, params) {
    // realtime event returns user without password
    await this.app.service('users').patch(params.user._id, {
      name: 'test'
    });

    // realtime events returns user with password
    await this.app.service('users').patch(null, {
      name: 'qwertz'
    });

    return null;
  }
};
```

channel.js
```
...
app.on('login', (authResult, { connection }) => {
    if(connection) {
      const user = connection.user;
      app.channel('anonymous').leave(connection);
      app.channel('authenticated').join(connection);
      app.channel(`userIds/${user._id}`).join(connection);
    }
  });

  app.service('users').publish(user => {
    return [app.channel(`userIds/${user._id}`)];
  });
...
```
When calling `app.service('some-service').create(data, params)` the handler on the client side: 
```
app.service('/users').on('patched', (user, context) => { console.log('users was patched', user) })
```
returns the following user objects:

1. Event
![image](https://user-images.githubusercontent.com/61047871/74534403-ef017d00-4f33-11ea-9163-8d8aa11271f4.png)

2. Event
![image](https://user-images.githubusercontent.com/61047871/74534422-f4f75e00-4f33-11ea-85c3-438fe1452f68.png)

### Expected behavior
I'm not sure if I missed something but i would expect that the password field would also be hidden inside the 2. Event.

I've created a repo where the issue can be reproduced:
https://github.com/Tr3ffel/feathers-update-many-bug

GautierT:
Same problem. Is this an expected behavior ? 

Thanks.

marshallswain:
@GautierT it's not expected behavior.  

[...]

PR #2641, opened 2022-05-21
feat(schema): Add resolver for safe external data dispatching

daffl:
This pull request adds a `resolveDispatch` hook which makes sure that only a safe version of the data ever gets sent to the client. This includes real-time events from internal calls and properties populated in nested resolvers. For example, the following resolver:

```ts
export const userDispatchResolver = resolve<UserResult, HookContext<Application>>({
  schema: userResultSchema,
  properties: {
    password: () => undefined
  }
});

service.hooks([
  resolveDispatch(userDispatchResolver)
]);
```

Will make sure that the user password never gets sent to a client. This replaces the `protect` hook and addresses its shortcomings (and closing https://github.com/feathersjs/feathers/issues/1809)

### DIFF
--- a/packages/schema/src/hooks.ts
+++ b/packages/schema/src/hooks.ts
@@ -13,6 +13,13 @@ const getContext = <H extends HookContext> (context: H) => {
   }
 }
 
+const getData = <H extends HookContext> (context: H) => {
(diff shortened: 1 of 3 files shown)


marshallswain:
Nice and clean

PR #2669, opened 2022-06-18
feat(cli): Add typed client to a generated app

daffl:
This pull request adds a Feathers client module to the generated application. This means you can import your API in your client side React, VueJS etc. project and instantiate a client specifically for it. If you are using TypeScript it will use the service names and schemas declared on the server.

```ts
import io from 'socket.io-client'
import socketio from '@feathersjs/socketio-client'
import authentication from '@feathersjs/authentication-client'

import { createClient } from 'my-api'

const client = createClient(socketio(io))
  .configure(authentication)

const page = await client.service('users').find({
 query: {} // Get autocompletion for all query schema props
})

page.data // Will be UserResult[] type
```

### DIFF
--- a/packages/authentication-oauth/src/index.ts
+++ b/packages/authentication-oauth/src/index.ts
@@ -42,18 +42,14 @@ export const setup = (options: OauthSetupSettings) => (app: Application) => {
     }
   }
 
-  const grant = defaultsDeep(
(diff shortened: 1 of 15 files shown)


marshallswain:
:shipit:

PR #2670, opened 2022-06-18
Set types to `"./lib/index.d.ts"`

FossPrime:
Here's the recommended most compatible way to do it...
https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html

The way it is in some of our modules causes compatibility problems. The do is how Node ESM module resolution expects it. Deno does not support defaulting to index.js due to a community wide move away from problematic magic. Specifically this has caused problems with CodeSandbox Cloud IDE, StackBlitz and would probably cause problems in Deno, integrations, CI tools that check integrity, tools that automatically transform code and other environments. There is no downside to the PR approach and no substantial benefit to the current approach.

What users currently see:
Screenshot (15)

We stray three steps away from the recommended way.

{2546598D-DF23-42FF-ADE6-D10E11B359CD}

Our way: https://unpkg.com/browse/@feathersjs/schema@5.0.0-pre.23/lib/

### DIFF
--- a/packages/schema/package.json
+++ b/packages/schema/package.json
@@ -3,8 +3,8 @@
   "description": "A common data schema definition format",
   "version": "5.0.0-pre.23",
   "homepage": "https://feathersjs.com",
-  "main": "lib/",
-  "types": "lib/index.d.ts",
+  "main": "./lib/index.js",
+  "types": "./lib/index.d.ts",
   "keywords": [
     "feathers",
     "feathers-plugin"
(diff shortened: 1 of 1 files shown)


daffl:
That makes sense. We should probably do this for all packages.

FossPrime:
If you merge this ill do it for the rest... I just have to make sure every package is pointing to the right place

I would also try to add a test to make sure every package has resolvable definitions

daffl:
Can you update this pull request with the changes? It makes for a better changelog that way.

FossPrime:
Closing this as I managed to get the makers of the tools I use to support our style. If someone else encounters a problem with it, like Deno people... we can make a massive PR that fixes it wholesale.

PR #2671, opened 2022-06-19
feat(knex): Add KnexJS SQL database adapter to core

daffl:
Similar to https://github.com/feathersjs/feathers/pull/2610 this pull request adds an updated version of https://github.com/feathersjs-ecosystem/feathers-knex to core for built-in SQL database support. Similar to MongoDB, with the new Feathers schemas and resolvers, a more low level database adapter (and not a full fledged ORM) makes the most sense to use as the recommended go-to database adapter (others will continue to work as is of course).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,4 @@ dist/
 
 # TypeScript compiled files
 packages/**/lib
+*.sqlite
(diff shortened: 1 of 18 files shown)


noor-tg:
👍🏼

Issue #2681, opened 2022-06-30
[@feathersjs/schema] `schema` function is not pure and crashes in ES Modules


FossPrime:
### Steps to reproduce

- Call schema twice with the same parameters.

### Expected behavior
Don't crash.

### Actual behavior
"schema with key or id "UserPatch" already exists"

### System configuration

**Module versions**: dove pre
**Module Loader**: esbuild

## Reproduction:

use schema instead of safeSchema, wait for server restart, than modify a comment string. When the server is being restarted the bug will crash esbuild wasm as we are building in the same context and something global is conflicting... this might be unavoidable, so I propose wrapping schema with my fix when appropriate.

https://stackblitz.com/edit/bug-feathersjs-schema-2680?file=src%2Fschema%2Fusers.schema.ts,src%2Futils.ts

daffl:
I would see how AJV does this since the error is coming from there. We could work around it but I'm pretty sure someone needed to do hot module reloading with AJV before.

FossPrime:
I've ruled out upstream... I cannot reproduce the issue with json-schema-to-ts@2.5.4 and vite@3.0.1

https://stackblitz.com/edit/vitejs-vite-qdrpwy?file=BugReproduction.ts

With up to date @feathersjs/schema dependencies, I can no longer reproduce the issue at all!

https://stackblitz.com/edit/bug-feathersjs-schema-2680-vfz5ho?file=src%2Fschema%2Fusers.schema.ts,src%2FsafeSchema.ts,client.ts

FossPrime:
https://stackblitz.com/edit/feathers-vite-chat-ys59tp?file=src%2Fschema.ts,src%2Fschema%2Fusers.schema.ts,src%2Fschema%2Fmessages.schema.ts,client.ts

For some reason copying schema to your src makes the intellisense server not crash...
https://stackblitz.com/edit/feathers-vite-chat-ys59tp?file=src%2Fschema%2Fusers.schema.ts

FossPrime:
The issue only manifests when loading a module from node_modules, if the module is anywhere else in the project, DEFAULT_AJV will be discarded, cleaning up any polluting modules. See PR.

FossPrime:
Other situations I can think of where this is a problem:

- Running two feathers instances in the same context and both of them have a User schema.
- Running tests on Schemas, you have to be careful to not name a schema the same as another test if the context is shared.
- dynamically creating schemas from a feathers client
[...]

PR #2702, opened 2022-07-21
Fix for Ajv global collision bug #2681

FossPrime:
### Summary

Fixes HMR issues with Ajv's schemas #2681 

Sandbox with fix applied: https://stackblitz.com/edit/feathers-vite-chat-pcztgp?file=src%2Fschema%2Fusers.schema.ts,schema%2Fsrc%2Fschema.ts

Repo with fix: https://github.com/MarsOrBust24/feathers-schema-freedom
Chat: https://discord.com/channels/509848480760725514/989366511665819698/999577745010999426

Alternatively we could instantiate Ajv inside the constructor, but it would have a performance penalty.

P.S. Type coercion by default seems like a bad idea, it will hide problems until it's too late and make debugging more difficult.

### DIFF
--- a/packages/schema/test/fixture.ts
+++ b/packages/schema/test/fixture.ts
@@ -14,7 +14,8 @@ import {
   querySyntax,
   Combine,
   resolveDispatch,
-  resolveAll
+  resolveAll,
+  Ajv
 } from '../src'
 import { AdapterParams } from '../../memory/node_modules/@feathersjs/adapter-commons/lib'
 
--- a/packages/schema/test/fixture.ts
+++ b/packages/schema/test/fixture.ts
@@ -18,6 +18,11 @@ import {
 } from '../src'
 import { AdapterParams } from '../../memory/node_modules/@feathersjs/adapter-commons/lib'
 
+const fixtureAjv = new Ajv({
+  coerceTypes: true,
+  addUsedSchema: true // default
+})
+
 export const userSchema = schema({
   $id: 'UserData',
   type: 'object',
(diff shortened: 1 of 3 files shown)


packages/schema/test/fixture.ts
```diff
@@ -14,10 +14,16 @@ import {
   querySyntax,
   Combine,
   resolveDispatch,
-  resolveAll
+  resolveAll,
+  Ajv
 } from '../src'
 import { AdapterParams } from '../../memory/node_modules/@feathersjs/adapter-commons/lib'
 
+const fixtureAjv = new Ajv({
```
daffl, line 22: Do we need a separate instance if the default one has been updated?

FossPrime: We only need to own our Ajv context if we want to combine schemas by making use of schema `$ref`. As we took away some of the global side effects of calling schema.

We need a separate instance BECAUSE we need to update/mutate it. This update makes the default instance more immutable and side-effect free.

FossPrime:
As to how to better support refs, we could add an addSchema method to the returned class. Which would then allow users to combine schemas with an import tree. This is much closer to the syntax Ajv uses.

```ts
// users.schema.ts
export const schema = new FeathersSchema()
export const UserSchema = schema.addSchema(...)
const UserResponseSchema = UserSchema.fo(...)
---
// messages.schema.ts
import { schema } from './users.schema.ts'
export const MessageSchema = schema.addSchema(...)
```

In that example UserSchema shares the same Ajv with MessageSchema. FeathersSchema would be a small class that stores an Ajv instance and takes Ajv parameters in it's constructor.

This could be implemented side by side with the current API, with a deprecation warning for anyone using the old unsafe API.

daffl:
I agree on the type coercion - it's mainly intended for query schemas since they generally come in as strings for REST calls. Not sure what the best setup would be. `addSchema` sounds interesting. How does this usually work in AJV?

PR #2703, opened 2022-07-24
Add instructions for handling CORS for sockets

blackmann:
### Summary

Since `v3` of socket.io `cors` options need to be explicit. https://socket.io/docs/v3/handling-cors/

I spent about two hours figuring out what's happening; I think it'll be helpful to others in future if the docs stated how to handle it.


daffl:
Hm... this is a good point, I'm wondering if the CLI should do more here and ask for the origin urls.

PR #2712, opened 2022-08-03
Indicate how/when .parse of the strategy is used

blackmann:
This closes the loop on understanding how auth strategies work. If you happened to look at this doc first, it's not apparent that this method is called from the service.


daffl:
That makes sense. Thank you!

PR #2745, opened 2022-09-10
docs: init ecosystem page

fratzinger:
Screencast: https://we.tl/t-0ZnTvjMwHX

Todo:
- [ ] fix build error @fratzinger
- [ ] add coloring to element-plus
- [x] add category filter
- [x] add filter for old packages
- [ ] add copy-to-clipboard npm-package
- [x] decide wether to show npm package name or github repo name? @daffl
- [x] add npm package name to cloudflare output @daffl
- [x] add more packages to cloudflare @daffl

Sondervögel / special entries:
- https://pub.dev/packages/flutter_feathersjs
- https://github.com/edwardsmarkf/fastfeathers
- https://github.com/feathersjs-ecosystem/feathers-swift
- https://github.com/feathersjs-ecosystem/feathers-debugger
- https://github.com/feathersjs-ecosystem/feathers-debugger-service


fratzinger:
Current screencast of the page: https://we.tl/t-0ZnTvjMwHX

daffl:
This looks fantastic! The only other things I can think of are:

- Adding categories to filter by (like "Frontend", "React", "Database" etc. similar to the table of contents in the https://github.com/feathersjs/awesome-feathersjs page)
- Filtering out stale (modules that didn't have a release or GitHub activity in more than three years) and archived modules by default (similar to the "core" selection)

fratzinger:
1. Categories to filter: I also think that's necessary. I would like to take the `package.json` keywords, that are already in the list. I'll work on that and come up with something useful.
2. I'll come up with something.

daffl:
We can use the keywords but I think the list to select from should be curated. As for the other question, I think GitHub name is good. Could we add a copy-able `npm install <packagename>` code block section for the package name? I'll get to adding the package name to the API.

fratzinger:
> list to select from should be curated

That is what I was up to. 👍

> copy-able `npm install <packagename>`

Good idea. totally! But if maybe copying the packagename with the packager is better for `yarn`/`pnpm`.

fratzinger:
We gotta update all of the keywords 😅
image

fratzinger:
Two questions to discuss:
1. What happens to https://github.com/feathersjs/awesome-feathersjs ? It's redundant to this docs page for packages. But it has 'Resources' and 'Projects using Feathers'.
2. How do you add a new package? The workflow at https://github.com/feathersjs/awesome-feathersjs with a PR is sufficient.

My two cents:
1. We need to keep https://github.com/feathersjs/awesome-feathersjs for resources and projects using feathers. But I think it's best to remove the packages from the readme completely. We don't want to maintain both but promote the docs page. We'll add a note to the feathers docs page. We don't want to delete the awesome repo since it's a widely known resource.
2. We need a simple way to submit new packages. 
A form on the docs page is quite complex: Do we need a backend then? Does it send an email to hello@fea... ?
The PR workflow as for https://github.com/feathersjs/awesome-feathersjs is pretty nice.
Why not adding the code of `feathershq/feathers-ecosystem` to the https://github.com/feathersjs/awesome-feathersjs repo? a PR to ecosystem.json can easily be made. We just need a github action that updates the cloudflare worker on merge on `main`.

daffl:
I was thinking of having a thing where you log in and submit your project but you're right, moving the JSON file to the  https://github.com/feathersjs/awesome-feathersjs  repo and making a PR against that file should be more than sufficient. Then people also can add their own keywords and/or categories.

I don't think we need a list of websites using Feathers anymore. Websites go offline all the time. If someone wants to associate their company with the project they can become a sponsor.

For the examples, can we make it possible to just have just a GitHub repository (no npm) and maybe show it slightly different? That way we can have the same filter criteria with keywords and staleness.

fratzinger:
Closing in favor of #2838

PR #2755, opened 2022-09-16
feat(cli): Add ability to `npm init feathers`

### LINKED ISSUE #1501: Suggestion: npm create feathers
```
$ npm install -g @feathersjs/cli
$ mkdir my-new-app
$ cd my-new-app/
$ feathers generate app
$ npm start
```

can be replaced by

```
$ npm create @feathersjs/app my-new-app
$ cd my-new-app
$ npm start
```
or
```
$ npx @feathersjs/create-app my-new-app
$ cd my-new-app
$ npm start
```

daffl:
This pull request adds a `create-feathers` package which allows to create a brand new Feathers application by simply running

```
npm init feathers feathers-chat
```

It will create a new folder and then run the `@feathersjs/cli` application generator.

In the new app, the CLI will also be installed locally so all commands are now run via `npx feathers`. That way every developer uses the version of the CLI installed with the app.

Closes https://github.com/feathersjs/feathers/issues/1501

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,8 +5,85 @@
   "packages": {
     "": {
       "name": "feathers",
-      "hasInstallScript": true,
       "license": "MIT",
+      "dependencies": {
+        "@babel/core": "^7.19.0",
+        "@babel/preset-env": "^7.19.0",
+        "@feathershq/pinion": "^0.3.5",
+        "@feathersjs/hooks": "^0.7.5",
+        "@koa/cors": "^3.4.1",
+        "@types/bcryptjs": "^2.4.2",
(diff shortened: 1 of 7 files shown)


marshallswain:
Rad. I would prefer `npm init feathers` to generate an app.  It will be the most-used generator, compared to a hook or resolver generator, or anything else we could build.

daffl:
Agreed, it does look nicer. This will probably be the only `npm init` package since all the other generators will be then run with the locally installed `@feathersjs/cli`. So the flow would be:

```
npm init feathers chat
cd chat
npx feathers generate service
npm start
```

Issue #446, opened 2016-10-25
Unregistering services and hooks


Download:
Currently, once the app is running, it's pretty difficult to unregister Feathers services and hooks. 

We would like to be able to support unregistering all event handlers associated with a route, so we can recreate and re-register them while the server keeps running. If we do this right, it will enable people to use hot reloading techniques such as Webpack's Hot Module Replacement with their Feathers services and hooks.

API suggested by @daffl in Slack:

``` js
app.service('/myservice',  null)
```

would completely undo the effects of

``` js
app.use('/myservice', myservice)
```

...and thus allow us to use HMR like this:

**services/myservice/index.js**

``` js
'use strict';

let app; // hang on to the app object
module.exports = function(){
  app = this;
  register()
};

function register() {
  // re-require on every invocation to facilitate HMR
  // `require` caches modules once loaded, so second require is fast.
  const hooks = require('./hooks');
  const Service = require('./service');
  app.use('/myservice', new Service());
  const service = app.service('/myservice');
  service.before(hooks.before);
  service.after(hooks.after);
}

function unregister() {
  app.service('/myservice', null);
}

if (module.hot) {
  module.hot.accept(['./hooks', './service'], function(){
    unregister(); // unregister the old service and hooks
    register(); // ..and register the new ones. 
  })
}

// not sure if it's useful to keep this next line...  modules referencing 
// `Service` would need to make sure they get the new one somehow...
// module.exports.Service = Service;
```

ElegantSudo:
That's a bit complicated. Is there a way we can augment this, so that on each deploy these tasks are already being reviewed and executed?

Download:
@ElegantSudo Can you elaborate on what is complicated? Basically the contents from the example above come from: 

1. the feathers CLI generator (but moved the logic into a function `register`)
2. the Webpack HMR api (e.g. see the [example](https://github.com/webpack/docs/wiki/hot-module-replacement#example-1-hot-replace-request-handler-of-http-server) on their Wiki)
3. The `unregister` function, which is basically the essence of this proposal

I think most of this code would be generated for you by `feathers generate service`.

ElegantSudo:
@Download that's a valid point. What I really meant is: can we add this to feathers so that this process is automated upon server deploy for every application?

I think this is an essential feature for any developer with an application.

marshallswain:
[...]

PR #2756, opened 2022-09-16
feat(core): Allow to unregister services at runtime

### LINKED ISSUE #2035: Unregistering services at runtime
Our organisation is using Feathers as our primary server and are deep into the development of a microservice written with Feathers that connects to our primary. We expose our primary Feathers services to the microservice and use it as a gatekeeper for a particular type of client. We do this by registering all the discovered services as local whenever a connection is established between the primary server and the microservice.

The problem that we've just discovered is that `app.use(serviceName, service)` doesn't overwrite an existing service of the same name, it actually increases the number of events in the socket.io instance. Ideally registering a service with the same name multiple times would destroy the existing one and create a new one.

Without this functionality it makes our (arguably not niche) use case impossible without incurring an inevitable increase in memory whenever the remote server reconnects. Yes, we can check if the service already exists and only register it if it doesn't. But if the central server removes a Feathers service there'd be no way to remove it from the microservice.

#446 floated this idea but it never gained any traction, with the most recent reply in 2018.
In #1200 it was mentioned that this functionality is still not possible, in 2019.

Is this something the Feathers team has on their roadmap?
Is there an unofficial way that this can be handled by manually clearing the references to the events in memory or similar?

daffl:
This pull request adds the ability to unregister a service dynamically by calling `await app.unuse('servicename')`. It will call the services `teardown` method and unregister all routes (for all transports).

Related to https://github.com/feathersjs/feathers/issues/446
Closes https://github.com/feathersjs/feathers/issues/2035


marshallswain:
I like `unuse` for its explicit relationship to `use`.

:shipit:

steffanhalv:
Finally 🥳

@daffl - When will this be published into a new pre-release? I have been waiting for this in a while :)

PR #2768, opened 2022-09-28
Our NPM page has over 10 broken links an images

FossPrime:
Screenshot_20220928-211552

Our NPM page possibly the highest traffic, highest ranked, most consequential FeathersJs front page

...and yet it links to a broken image of a broken slack channel, has a date of 2019, and about 3 broken donation buttons...


daffl:
Thanks! I'll have to see when to cut this into a new release since v5 should become latest shortly but you are making a good point that the current one should be up to date.

Issue #2031, opened 2020-07-27
OAuth internal redirects discard API prefixes


emilselroos:
I am trying to implement Google OAuth authentication in my FeathersJS (4.5.7) application. I have an API listening to `/api/v1` prefix - done by wrapping the feathers application within another express application. (Similar to this [comment](https://github.com/feathersjs/feathers/issues/531#issuecomment-287153441))

### Expected behavior

Going to `/api/v1/oauth/google` should redirect users to `/api/v1/oauth/connect/google`.

### Actual behavior

Redirecting to `/oauth/connect/google`, discarding the API prefix and leading to `CANNOT GET /oauth/connect/google` page.

---

My authentication config in `config.json`

```
"authentication": {

	"path": "/authentication",
	"entity": "account",
	"entityId": "account_id",
	"service": "accounts",
	"secret": "SECRET",
	"authStrategies": [ "jwt", "oauth" ],

	"jwtOptions": {
	  "audience": "localhost",
	  "issuer": "feathers",
	  "algorithm": "HS256",
	  "expiresIn": "1d"
	},

	"oauth": {
		"redirect": "http://localhost:3000",
		
		"defaults": {
			"protocol": "http",
			"host": "localhost",
			"path": "/oauth",
			"transport": "session",
			"state": true
			
		},
		"google": {
			"key": "KEY",
			"secret": "SECRET",
			"scope": [ "email", "profile", "openid" ],
			"nonce": true,
			"redirect_uri": "http://localhost:3030/api/v1/oauth/google/callback"
		}
	}

  }
```

And my `authentication.js` file

```
const { AuthenticationService, JWTStrategy } = require('@feathersjs/authentication');
const { expressOauth, OAuthStrategy } = require('@feathersjs/authentication-oauth');

class GoogleStrategy extends OAuthStrategy {
	async getEntityData (profile) {
		const baseData = await super.getEntityData(profile);

		return {
			...baseData,
			google_id: profile.googleId,
			profile_picture: profile.picture,
			email: profile.email,
			name: profile.given_name + ' ' + profile.family_name,
		}
	}

	async getRedirect(data) {
		return '/frontend';
	}
}

module.exports = app => {

	const authentication = new AuthenticationService(app);

	authentication.register('jwt', new JWTStrategy());
	authentication.register('google', new GoogleStrategy());
	
	app.use('/authentication', authentication);
	app.configure( expressOauth() );

}
```

michaek:
[...]

PR #2773, opened 2022-10-04
fix(authentication-oauth): Prefix handling in OAuth

nsainaney:
### Summary
OAuth service path ignores grant `default.prefix`. I am not sure if this is intentional but here's a proposed PR to fix it.

The issue is described here: https://github.com/feathersjs/feathers/issues/2031#issuecomment-1266538838


daffl:
It definitely was not intentional but I was afraid I'd miss something like that in the refactoring. Thank you for this!

PR #2820, opened 2022-10-20
WIP: Docs "Troubleshooting" Guide

marshallswain:
Screen Shot 2022-10-20 at 2 30 13 PM

### DIFF
--- a/docs/.vitepress/config.ts
+++ b/docs/.vitepress/config.ts
@@ -231,6 +231,10 @@ export default defineConfig({
             {
               text: "What's new?",
               link: '/guides/whats-new.md'
+            },
+            {
+              text: 'Troubleshooting',
+              link: '/guides/troubleshooting.md'
             }
           ]
         }
--- a/docs/guides/troubleshooting.md
+++ b/docs/guides/troubleshooting.md
@@ -1,0 +1,127 @@
+---
+outline: deep
+---
+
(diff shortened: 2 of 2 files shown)


daffl:
We'll definitely revisit this once we have more questions and see what the common issues with the final version are. This is what I took from the current MongoDB docs and just going to put it here to not loose it:

```markdown

## Common Mistakes

Here are a couple of errors you might run into while using validators.

### unknown keyword: "convert"

You'll see an error like `"Error: strict mode: unknown keyword: "convert"` in a few scenarios:

- You fail to [Pass the Custom AJV Instance to every `schema`](#pass-the-custom-ajv-instance-to-schema). If you're using a custom AJV instance, be sure to provide it to **every** place where you call `schema()`.
- You try to use custom keywords in your schema without registering them, first.
- You make a typo in your schema. For example, it's common to forget to accidentally mis-document arrays and collapse the item `properties` up one level.

### unknown format "date-time"

You'll see an error like `Error: unknown format "date-time" ignored in schema at path "#/properties/createdAt"` in a few scenarios.

- You're attempting to use a formatter not built into AJV.
- You fail to [Pass the Custom AJV Instance to every `schema`](#pass-the-custom-ajv-instance-to-schema). If you're using a custom AJV instance, be sure to provide it to **every** place where you call `schema()`.
```

PR #2826, opened 2022-10-21
docs allow copying Typebox example code

marshallswain:
Instead of the wonky table, this organizes the code into sections that can be navigated in the table of contents and sidebar.  It was impossible to copy text out of the other table, before.

Since this is a long doc, I've added the table of contents for mobile devices at the to of the page:

Screen Shot 2022-10-21 at 1 07 49 AM

Here's what the code sections look like:

Screen Shot 2022-10-21 at 1 08 51 AM


daffl:
Thank you! I've been thinking about this but then was too lazy 😆

PR #2531, opened 2022-01-14
WIP: Port Core to Deno

marshallswain:
### Summary

These are the packages and tasks required to finish the port to Deno:

- [x] adapter-commons
   - [x] code cleanup
   - [x] tests
      - [x] commons
      - [x] filter-query
         - [x] mongodb ObjectId
      - [x] service
      - [x] sort
   - [x] Update with https://github.com/feathersjs/feathers/commit/9c22f70a838cb6341775d91705a7527c8fc5590e
   - [ ] Update with https://github.com/feathersjs/feathers/commit/c3ec8a418bdc85506e3c5100015720a45454d8d3
   - [ ] Update with https://github.com/feathersjs/feathers/commit/f56c9e2ce61e067a4eb28b8b8c8f5d1519784170
- [x] commons
   - [x] add `events` typescript port
   - [x] code cleanup
   - [x] tests
      - [x] debug
      - [x] module
      - [x] utils
- [x] errors
   - [x] code cleanup
   - [x] tests
- [ ] authentication
   - [ ] code cleanup
   - [ ] tests
- [ ] authentication-client
   - [ ] code cleanup
   - [ ] tests
- [ ] authentication-local
   - [ ] code cleanup
   - [ ] tests
- [ ] feathers
   - [ ] code cleanup
   - [ ] tests
- [ ] transport-commons
   - [ ] code cleanup
   - [ ] tests

### Other Packages

We need to discuss how to port these:

- [ ] adapter-tests
- [ ] authentication-oauth
- [ ] client
- [ ] configuration
- [ ] memory
   - [ ] sift: https://github.com/crcn/sift.js/issues/238 
- [ ] rest-client
- [ ] schema
   - [ ] ajv?
- [ ] socketio
- [ ] socketio-client
- [ ] tests

### Node-only Packages

- [ ] express
- [ ] koa

PR #2828, opened 2022-10-22
WIP: Dove deno new

apollo79:
### Summary
**Important:**
This PR is not finished, it is a WIP!

This PR updates the packages ported in #2531 to the newest versions.

I used the `dove` branch and copied most things. 
The main things I changed are using the **[events](https://deno.land/x/events) library as EventEmitter** and the **[std bdd testing library](https://deno.land/std/testing/bdd.ts) for tests**

And finally, as it is somewhat of a convention in deno, I added `mod.ts` files to every module.

### Other Information
I mentioned this PR on the discord yesterday, there was some conversation about how to go with deno before:
https://discord.com/channels/509848480760725514/930352418179391528/1033008949252857987


apollo79:
One thing to talk about would be which formatter to use as the formatter included in the [VSCode deno extension](https://marketplace.visualstudio.com/items?itemName=denoland.vscode-deno) doesn't format in the same way as running `deno fmt` does. Currently I used the deno extension, but if we'll maybe check formatting via a GitHub Action in the future we should use `deno fmt`

main/adapter-commons/src/service.ts
```diff
@@ -1,84 +1,206 @@
...
+  }
+
+  /**
+   * Sanitize the incoming data, e.g. removing invalid keywords etc.
+   *
+   * @param data The data to sanitize
+   * @param _params Service call parameters
+   * @returns The sanitized data
+   */
+  async sanitizeData<X = Partial<D>>(data: X, _params?: P) {
```
apollo79, line 98: For typescript I had to make `_params` optional as in the `_create` method, `sanitizeData` is called with params that are maybe undefined

main/adapter-commons/src/service.ts
```diff
@@ -1,84 +1,206 @@
...
    * @see {@link https://docs.feathersjs.com/api/services.html#get-id-params|Feathers API Documentation: .get(id, params)}
    */
-  _get (id: Id, params?: AdapterParams): Promise<T>;
+  async _get(id: Id, params?: P): Promise<T> {
+    const query = await this.sanitizeQuery(params);
+
+    return this.$get(id, {
+      ...params,
+      query,
+    } as P);
```
apollo79, line 166: Here (not only here but in almost every method (`_find`, `_create`, `_update`, ...)) I had to cast the typ as `P`, because TypeScript complained about the following

> Argument of type '{ query: Query; adapter?: Partial<AdapterServiceOptions> | undefined; paginate?: PaginationParams | undefined; provider?: string | undefined; route?: { ...; } | undefined; headers?: { ...; } | undefined; }' is not assignable to parameter of type 'P'.
  '{ query: Query; adapter?: Partial<AdapterServiceOptions> | undefined; paginate?: PaginationParams | undefined; provider?: string | undefined; route?: { ...; } | undefined; headers?: { ...; } | undefined; }' is assignable to the constraint of type 'P', but 'P' could be instantiated with a different subtype of constraint 'AdapterParams<AdapterQuery, Partial<AdapterServiceOptions>>'.deno-ts(2345)

I don't know if there is a way to get around this without type casting (what would of course be very cool).

Btw we could just cast `params` as `P`, but I don't know if it is more readable:
```ts
    return this.$get(id, {
      ...(params as P),
      query,
    });
```

daffl:
I had a quick look and this look great, thank you for doing that. My initial questions would be:

- Can we use the build script from `@feathersjs/hooks` to build the Deno packages into the `packages/` folder (and then `.gitignore` those because we will only need them for publishing)?
- What's the best way to run tests with Deno here?
- Can we open this PR against the latest Dove branch. I'm assuming eventually we can remove the `deno/` folder.

apollo79:
Thanks for the feedback!
About the questions:
>     * Can we use the build script from `@feathersjs/hooks` to build the Deno packages into the `packages/` folder (and then `.gitignore` those because we will only need them for publishing)?
I hope so, I think, it should be possible with a few changes.
The problem isthat I am not able to run the build script in `@feathersjs/hooks` and neither the one I created for this PR which is quite similar.
The build script of `@feathersjs/hooks` throws the error
```
error: Uncaught (in promise) 'Module not found "file:///C:/C:/Users/<path_to_feathersjs_hooks>/main/hooks/src/index.ts".'
```
I am on Windows and it seems that there is a problem with the `URL`, that represents `__dirname` or `path.join` or something like that (notice the `C:/C:/` in the file path).
> What's the best way to run tests with Deno here?
To run the tests just run
```
deno test
```
in the folder of the module / package you want to test or in `main` to test all modules

> Can we open this PR against the latest Dove branch. I'm assuming eventually we can remove the deno/ folder.

Yes, of course. Should I do this right now or later when some more packages are ported?

PR #2840, opened 2022-10-30
feat(mongodb) aggregation pipeline queries

marshallswain:
## Overview

- Queries use the MongoDB Aggregation Framework, by default.
- Use `params.pipeline` to append additional aggregation stages.
- Custom `$feathers` stage allows granular control over pipeline stages.
- Dynamically modify Feathers pipeline stages with the `$feathers.handleStages` utility function.
- Non-breaking, fully backward compatible: continues to support requests needing custom collation and the `collection.find` API.
- New service utility methods that turn Feathers `params` into raw MongoDB queries.
- Faster paginated `find` performance with parallel requests.
- Adds a `getModel` method to make service methods more DRY.

## Aggregation Pipeline Queries

Makes queries using the MongoDB Aggregation Pipeline!  Here's an example showcasing the new feature:

```ts
const result = await app.service('todos').find({
  query: { $sort: { name: 1 } },
  pipeline: [
    {
      $lookup: {
        from: 'users',
        localField: 'userId',
        foreignField: '_id',
        as: 'user'
      }
    },
    { $unwind: { path: '$user' } }
  ],
  paginate: false
})
```

The above is like doing a join, but without the data transforming overhead like you'd get with a SQL JOIN.

## `params.pipeline`

The `find` method has been updated so that queries not meeting the two criteria, above, are created using the aggregation pipeline, which is an array of the document objects representing the stages of the request. Here’s an example which shows how to populate the `user` onto a query for `todos`, assuming that each `todo` contains a `userId` property.

```ts
app.service(‘todos’).find({
  query: {
    $sort: { name: 1 },
    $skip: 1,
    $limit: 4
  }
})

// the internal aggregation request looks like this
const query = collection.aggregate([
  { $match: {} },
  { $sort: { name: 1 } },
  { $skip: 1 },
  { $limit: 4 },
  ...(params.pipeline || [])
])
```

Notice that it’s now possible to pass `params.pipeline`.  This means you can set params.pipeline as an array of additional stages in the aggregation pipeline and they will be executed during the same request.  This means that it’s now possible to perform any of the [Aggregation Stages](https://www.mongodb.com/docs/upcoming/reference/operator/aggregation-pipeline/) like `$lookup` and `$unwind`, integration with the normal Feathers queries.

In the first example, the `query` is added to the pipeline, first (under the hood) then additional stages are added in the `pipeline` option:

- The `$lookup` stage creates an array called `user` which contains any matches in `todo.userId`, so if `userId` were an array of ids, any matches would be in the `users` array.  However, in this example, the `userId` is a single id, so…
- The `$unwind` stage turns the array into a single object.

A couple of other notable query stages:

- `$graphLookup` lets you recursively pull in a tree of data from a single collection.
- `$search` lets you do full-text search on fields

All stages of the pipeline happen directly on the MongoDB server.

## Custom `$feathers` Aggregation Stage

This PR adds the ability to optionally provide a custom `$feathers` stage, which allows you to specify where in the pipeline you want to inject the Feathers query-related pipeline stages.  This allows you to run some stages BEFORE the Feathers query runs, which is really handy for when you want to filter based on populated data (all in a single request).  In practice, the `$feathers` stage looks like this:

```ts
const result = await app.service('todos').find({
  query: { $sort: { name: 1 }, name: /Bob/ },
  pipeline: [
    { $match: { name: /dishes/ } },
    { $feathers: {} }, // is replaced by the stages dynamically generated from the `query`
    {
      $lookup: {
        from: 'people',
        localField: 'userId',
        foreignField: '_id',
        as: 'person'
      }
    },
    { $unwind: { path: '$person' } }
  ],
  paginate: false
})
```

Aggregation Framework stages build on top of each other, so in the above example, the first `$match` stage will pull all documents containing the word `dishes` into the MongoDB server's memory.  Then the `$feathers` stage will apply the pipeline that's made by the `query`, which will then further limit the initial set of data to include only those items that also contain `Bob`.  This is a bit of a contrived example.  It's more valuable when you start managing multiple collections of data in a single pipeline.

## Backwards Compatibility

All previous functionality is maintained by automatically switching to `collection.find` instead of `collection.aggregate` if either of the following conditions is met.

- `params.mongodb` is provided. This object is for the `collection.find` method. It cannot be passed to `collection.aggregate`.
- `params.query.$limit` is set to `0`. The aggregation pipeline will not accept 0 as a value for the `$limit` stage.

## New Service Utility Methods

The `find` method has been split into separate utilities for converting params into different types of MongoDB requests:

- `asFindQuery` takes a set of params and converts them to a raw `collection.find` request.
- `asAggregateQuery` takes a set of Feathers params and converts them to a raw `collection.aggregate` request.
- `asPipeline` takes a set of Feathers params and converts them to a pipeline array, ready to pass to `collection.aggregate`. This utility comprises the bulk of the `asAggregateQuery` functionality.

The `asFindQuery` and `asAggregateQuery` both return the "raw" MongoDB request object, which makes it possible to use the returned MongoDB Cursor object for custom interaction with the database.  Having access to the MongoDB cursor isn't very useful for external API requests.  It's a feature that's more targeted to custom server scripts.

## Faster, paginated `find` performance

I noticed that for paginated Feathers queries, the request for data wasn't starting until after the count query returned from the database server.  The find method has been refactored to allow those requests to be made in parallel.  It's not very noticeable when testing on localhost. Tests run about 20-30ms faster. It effectively cuts transfer time in half.

## The `getModel` method

Service methods now all use a new `getModel` method for retrieving the model from options or params.  The `getModel` method reduces repetition of similar logic in each of the service interface methods.


marshallswain:
Looking at this with fresh eyes, I think the utility functions need to be renamed.  I named them within the context of them only being useful inside of the `find` method.  Now that they've been made useful as standalone utilities, the api seems weird:

```ts
app.service('todos').asFindQuery(params)
app.service('todos').asAggregateQuery(params)
```

Maybe we drop the `as`.  This is probably the most likely to be remembered.  I wanted to change `service.aggregateQuery` to just `service.aggregate`, but we can't do the same thing for `findQuery`, since `find` is taken.  Since they have similar functionality of turning params into a raw query, it makes sense to give them a similar name.

```ts
app.service('todos').findQuery(params)
app.service('todos').aggregateQuery(params)
```

Or maybe drop the `as` and reverse the words:

```ts
app.service('todos').queryFind(params)
app.service('todos').queryAggregate(params)
```

Or maybe we suffix with `Raw` or `Native`, since they return a raw, native MongoDB query object:

```ts
app.service('todos').findRaw(params)
app.service('todos').aggregateRaw(params)
```

```ts
app.service('todos').findNative(params)
app.service('todos').aggregateNative(params)
```

marshallswain:
I converted the PR to a draft.  While using it in my app, I've realized that it would be valuable to be able to specify aggregation pipeline operators before or after the Feathers ones.  So I'll take a look at ways to cleanly accomplish that sometime this week.

marshallswain:
I've marked this as ready to review.  Can we open a separate PR for the docs?  I think we still need to give the MongoDB docs some Dove love, anyway.

marshallswain:
@daffl, I've added docs.  Please take a look.

PR #2843, opened 2022-11-02
docs(awesome): use query-string url for search filters

fratzinger:
As discussed with daffl, the search values are synced with the url at https://dove.feathersjs.com/awesome/

### Todo:
- [ ] It's working on my local machine but only in dev, not in the build. I guess it has something to do with SSR. But I think I know what to do. I'll update this once I'll have figured it out.


fratzinger:
I think, I've finished this. I've absolutely no idea why netlify fails though...

PR #2844, opened 2022-11-03
fix(memory): Use for loop in _find() for better performance

DaddyWarbucks:
This PR increases the performance of `_find()` by optimizing the way data is matched/limited. This is accomplished by using a `for` loop to handle multiple operations in the same loop instead of using filter/map/slice. 

## Problem
```js
// Given a service with 10,000 results and default pagination 10

// Current implementation
// 1 - Filters 10,000 results (even though no query)
// 2 - Slices to 10 results
// 3 - Maps to _select
const users = app.service('users').find():

// NEW implementation
// 1 - Returns first 10 results with select
const users = app.service('users').find():

// With query

// Current implementation
// 1 - Filters 10,000 results with query
// 2 - Slices to 10 results
// 3 - Maps to _select
const users = app.service('users').find({ query: { name: 'Bob' } }):

// NEW implementation
// 1 - Filters/select up to 10 results with query
// 2 - Returns 10 results
const users = app.service('users').find({ query: { name: 'Bob' } }):

// With limit/skip

// Current implementation
// 1 - Filters 10,000 results with query
// 2 - Skips to 20 results
// 3 - Slices to 20 results
// 4 - Maps to _select
const users = app.service('users').find({ query: { name: 'Bob', $limit: 20, $skip: 20 } }):

// NEW implementation
// 1 - Filters/skip/limit/select up to 20 results with query
// 2 - Returns 20 results
const users = app.service('users').find({ query: { name: 'Bob', $limit: 20, $skip: 20 } }):


```

## Pros
- Unlike `filter`, the `for` loop allows us to `break` after `$limit` number of matches, which means we do not run the matcher on results that would have been discarded later via skip/limit.
- Does not require an extra map to select results. `values = values.map(value => _select(value, params))` is no longer needed because the `_select` also happens in the for loop.
- Less memory. This is negligible, but this approach also uses less arrays/memory by removing the multiple `array.slice` used to handle skip/limit.

## Cons
- Arguably more complicated. And because `feathers-memory` is often used as an example in the feathers ecosystem, this package should be simple.

If we don't choose to adopt this whole PR, I do feel strongly that we should at least implement [1cc816429b6d29fa33e15077a815fabf2c964a54 ](https://github.com/feathersjs/feathers/commit/ecf4fdb790385edfdc68356410527b6bd6c16e09)which is a very simple change that skips initial `filter` if there is no query.


DaddyWarbucks:
I've completed some basic tests of this implementation in an app with roughly 10,000 items (which is all localStorage could hold). The services have a default pagination of 30 and max 100, which means most tests were done with `$limit: 30`.

### Without Query
We basically just get the advantage of using the `for-loop` and `break` here to limit results. Less loops than the old implementation's filter method.
- No Sort - 20% faster.
- Sort - 12% faster.

### With Complex Query
Less loops, and each loop costs more. So more gains
- No Sort - 130% faster.
- Sort - 35% faster.

### With Simple Query
Less loops, and each loop costs more. So more gains
- No Sort - 100% faster.
- Sort - 35% faster.

The only instance in which the new implementation is slower is when you are using both a query and sort and there are no results matching the query. This is because the current implementation filers the array to 0 and then sorts the empty array, and the new implementation must do both its sort/find operations on the whole array.

When lowering the `$limit` there were even more performance gains. And that makes sense, the `for-loop` gets to break sooner.  And on services with fewer items these numbers were higher as well.

daffl:
This is great, thank you! Will go out with the next prerelease.

daffl:
Apologies for commenting on an old PR but I think we missed something here which the tests didn't cover as well which is that this won't get the `total` right. I believe the following test should pass but doesn't:

```ts
  it('using $limit still returns correct total', async () => {
    for (let i = 0; i < 10; i++) {
      await service.create({
        name: `Tester ${i}`,
        age: 19
      })
      await service.create({
        name: `Tester ${i}`,
        age: 20
      })
    }

    try {
      const results = await service.find({
        query: {
          $skip: 3,
          $limit: 5,
          age: 19
        }
      })

      assert.strictEqual(results.total, 10)
      assert.strictEqual(results.skip, 3)
      assert.strictEqual(results.limit, 5)
    } finally {
      await service.remove(null, {
        query: {
          age: {
            $in: [19, 20]
          }
        }
      })
    }
  })
```

It looks like some of the improvements may have come from returning the wrong total - we can still do this but only if pagination is disabled.

DaddyWarbucks:
Ruh roh! I can take a look at this Monday and get it fixed.

daffl:
I noticed it when porting the adapter to https://github.com/wingshq/wings, the new Feathers-independent home for the database adapters. I think we can still do the loop when pagination is disabled.

PR #2845, opened 2022-11-03
fix(cli): Fix MongoDB connection database name parsing

nicholaswhitney:
From the feathers discord:

The current parsing method fails the following MongoDB connection string:
```
mongodb://user:P%40ss@localhost:27017/databaseName?authSource=admin
```

Where it parses `databaseName?authSource=admin` as the database name instead of just `databaseName`.

The following change should parse all database connection strings correctly.

There don't seem to be any open issues related to this, and it isn't dependent on PRs in other repos.


daffl:
Great, thank you!

PR #2876, opened 2022-11-16
fix(adapter-commons): multiple issues in adapter-commons

TheNoim:
### Summary

here](https://github.com/feathersjs/feathers/blob/dove/.github/contributing.md#pull-requests))

- [X] Tell us about the problem your pull request is solving.
- [X] Are there any open issues that are related to this? No
- [X] Is this PR dependent on PRs in other repos? No

### Other Information

This PR solves the following issues:

- Usage of double Partial which results in the type `Partial<Partial<T>>`
- Corrects return of $find in the test fixture of adapter-commons
- Corrects return of $update in the test fixture of adapter-commons
- Changes `never[]` to `any[]` in service.test.ts of adapter-commons to prevent type error

### Feedback

It is hard to find the contributing.md file. I suggest adding a link in the README.md for better visibility.

### DIFF
--- a/packages/adapter-commons/test/fixture.ts
+++ b/packages/adapter-commons/test/fixture.ts
@@ -6,23 +6,22 @@ export type Data = {
 }
 
 export class MethodBase
-  extends AdapterBase<Data, Partial<Data>, AdapterParams>
+  extends AdapterBase<Data, Data, AdapterParams>
   implements InternalServiceMethods<Data>
 {
   async $find(_params?: AdapterParams & { paginate?: PaginationOptions }): Promise<Paginated<Data>>
   async $find(_params?: AdapterParams & { paginate: false }): Promise<Data[]>
-  async $find(params?: AdapterParams): Promise<Data | Data[] | Paginated<Data>>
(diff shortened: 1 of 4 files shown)


packages/adapter-commons/test/fixture.ts
```diff
@@ -51,7 +50,7 @@ export class MethodBase
   }
 
   async $update(id: NullableId, _data: Data, _params?: AdapterParams) {
```
fratzinger, line 52: As far as I know, `NullableId` is wrong for `update`. Only `patch` and `remove` are allowed to have a `NullableId`. Am I wrong @daffl?

marshallswain: Correct. It should just be `Id`.  If the interface were setup for `NullableId`, it would mean you could completely replace multiple documents with the same `id`, which isn't possible.

daffl: This was my bad. A bunch of those issues will be fixed in https://github.com/feathersjs/feathers/pull/2906

fratzinger: Nice, thanks!

PR #2904, opened 2022-12-03
fix(schema): Allow query schemas with no properties, error on unsupported types

daffl:
This pull request fixes an issue in the `querySyntax` helpers for JSON schema and TypeBox where it threw an error when passing an empty schema (no properties) which should still be possible. It now also throws a more descriptive error when trying to create a query syntax helper for a non-primitive type which is currently not possible.

### DIFF
--- a/packages/schema/src/json-schema.ts
+++ b/packages/schema/src/json-schema.ts
@@ -122,6 +122,8 @@ export const queryProperty = <T extends JSONSchema>(def: T) => {
   } as const
 }
 
+export const SUPPORTED_TYPES = ['string', 'number', 'integer', 'boolean', 'null']
+
 /**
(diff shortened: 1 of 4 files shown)


packages/schema/src/json-schema.ts
```diff
@@ -122,6 +122,8 @@ export const queryProperty = <T extends JSONSchema>(def: T) => {
   } as const
 }
 
+export const SUPPORTED_TYPES = ['string', 'number', 'integer', 'boolean', 'null']
```
marshallswain, line 125: For MongoDB we might need `array` in the supported types. MongoDB can handle a query like `{ userIds: { $in: [1, 2, 3] } }`.

daffl: That should already be the case. `$in` and `$nin` are arrays of the primitive type.

PR #2906, opened 2022-12-04
feat(adapter): Add patch data type to adapters and refactor AdapterBase usage

daffl:
This pull request adds the ability to use a `PatchData` type in the adapters which may be different than the usual data type.

It also refactors the `AdapterBase` to remove the confusing and convoluted `$` prefixed service methods. Instead an adapter just implements the underscored service methods as before. While for the top level usage backwards compatible this may be a breaking change in some hopefully still uncommon instances but it reduces a lot of the complexity when implementing a database adapter.

You can still extend directly from the adapter class to implement the actual service methods with your own signatures and return types. This will circumvent the legacy query sanitisation which is now done with query schemas anyway.


packages/adapter-commons/test/service.test.ts
```diff
@@ -154,7 +154,7 @@ describe('@feathersjs/adapter-commons/service', () => {
     })
   })
 
-  it('allowsMulti', () => {
+  it.only('allowsMulti', () => {
```
marshallswain: `it.only`

packages/knex/src/index.ts
```diff
@@ -16,32 +17,55 @@ export class KnexService<T = any, D = Partial<T>, P extends Params<any> = KnexAd
...
+      ...params,
+      query: await this.sanitizeQuery(params)
+    })
   }
 
   async create(data: D, params?: P): Promise<T>
   async create(data: D[], params?: P): Promise<T[]>
   async create(data: D | D[], params?: P): Promise<T | T[]> {
+    if (Array.isArray(data) && !this.allowsMulti('create', params)) {
+      throw new MethodNotAllowed('Can not create multiple entries')
```
marshallswain, line 37: "Can Not" --> "Cannot"

marshallswain (approved):
Looks good.  Will we make `sanitizeQuery` conditionally run in another PR, or did I miss it in this one?

daffl:
I just added it in https://github.com/feathersjs/feathers/pull/2906/files#diff-686d31b4a5c93b9ab0ae6b60817ed194db3eccfaff344cfec6382f2439ce1e03R100

marshallswain (approved):
New updates look good.

PR #2914, opened 2022-12-07
fix(schema): Check for undefined value in resolveQueryObjectId resolver

philipimperato:
### Summary

Insures that a undefined or null field within the mongodb objectId resolver (resolveQueryObjectId) doesn't throw an exception and returns undefined.

### DIFF
--- a/packages/mongodb/src/resolvers.ts
+++ b/packages/mongodb/src/resolvers.ts
@@ -19,6 +19,10 @@ export async function resolveQueryObjectId(
 ): Promise<IdQueryObject<ObjectId>>
 export async function resolveQueryObjectId(value: ObjectIdParam): Promise<ObjectId>
 export async function resolveQueryObjectId(value: ObjectIdParam | IdQueryObject<ObjectIdParam>) {
+  if (value === undefined || value === null) {
+    return undefined
+  }
+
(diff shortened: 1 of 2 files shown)


packages/mongodb/src/resolvers.ts
```diff
@@ -19,6 +19,10 @@ export async function resolveQueryObjectId(
 ): Promise<IdQueryObject<ObjectId>>
 export async function resolveQueryObjectId(value: ObjectIdParam): Promise<ObjectId>
 export async function resolveQueryObjectId(value: ObjectIdParam | IdQueryObject<ObjectIdParam>) {
+  if (value === undefined || value === null) {
```
marshallswain: I think this is a scenario where we can actually just do `if (!value)` since the value has to be a string or an object, and neither of those can be `0`.

daffl: I think this is ok. Otherwise it would still error if you pass `null` as the value.

philipimperato: Oops, I already made the changes prior to reading this. Resubmitted.

daffl:
Great, thank you! Will go out shortly.

PR #2925, opened 2022-12-13
chore(docs): Add note about `methods` option

AshotN:
Add warning that `undefined` `methods` property will enable all standard methods.

### DIFF
--- a/docs/api/application.md
+++ b/docs/api/application.md
@@ -107,6 +107,12 @@ app.use('messages', new MessageService(), {
 })
 ```
 
+<BlockQuote type="warning" label="Important">
+
+If the `methods` property is `undefined`, all standard methods will be enabled and accessibly externally.
+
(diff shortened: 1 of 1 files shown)


daffl:
Do you think the *default* section in the `options` documentation above is not enough?

> `methods` (default: `['find', 'get', 'create', 'patch', 'update','remove']`)

AshotN:
> Do you think the _default_ section in the `options` documentation above is not enough?
> 
> > `methods` (default: `['find', 'get', 'create', 'patch', 'update','remove']`)

My only worry would be that someone might expect overlook that and think Feathers will only allow methods explicitly mentioned. Since `undefined` is going to default to every method for the sake of backwards compatibility. I figure a explicit warning may help someone avoid that mistake.

daffl:
That's fair. Done, thank you!

PR #2936, opened 2022-12-16
fix(cli): More app structure guide content and minor generated app improvements

daffl:


packages/cli/src/service/templates/client.tpl.ts
```diff
@@ -18,13 +20,11 @@ export type {
   ${upperName}Data,
   ${upperName}Query
 }
-`
-
-const methodsTemplate = ({ camelName, upperName, className, type }: ServiceGeneratorContext) => `
-const ${camelName}ServiceMethods = ['find', 'get', 'create', 'update', 'patch', 'remove'] as const
-type ${upperName}ClientService = Pick<${className}${
+export const ${camelName}ServiceMethods = ['find', 'get', 'create', 'patch', 'remove'] as const
```
8BallBomBom, line 23: Missing `update`?
Noticed in a newly generated app.

daffl: You can add it if you are using it but we found that `patch` in most cases provides the exact same functionality, at least for the database adapters.

8BallBomBom: Totally understandable, makes sense and makes things easier 😃

PR #2937, opened 2022-12-16
chore: Update Algolia config for docs

marshallswain:
Enable the DocSearch account.  Now all we need is to configure the crawler.


daffl:
Awesome, glad to have this working now! :)

daffl:
Ah dang, I merged the wrong one 😆 Will remove the Algolia configuration in #2942

PR #2945, opened 2022-12-20
feat(databases): Add support for fast bulk operations

### LINKED ISSUE #219: Multi create/patch/remove performance
When using `multi: true`, the `create` method is slow. See: https://discord.com/channels/509848480760725514/1049386346030047293

Solutions
- This line: https://github.com/feathersjs-ecosystem/feathers-mongodb/blob/c8e894cd433b39e5dc92ac824a983944768d25eb/lib/index.js#L219 should probably use { _id: { $in: [result.insertedIds] } } 
- Implement some `$returning: false` or `$limit: 0` to disable Feathers from returning results.

daffl:
This pull request adds the ability to the built in database adapters to perform fast bulk operations for `create`, `patch` and `remove` by passing `params.bulk = true`. Bulk operations will use the fastest database call available and - when successful - always return no data (an empty array) and not send real time events.

Closes https://github.com/feathersjs-ecosystem/feathers-mongodb/issues/219


packages/memory/src/index.ts
```diff
@@ -145,18 +148,18 @@ export class MemoryAdapter<
...
   async _create(data: Partial<Data> | Partial<Data>[], _params?: ServiceParams): Promise<Result | Result[]>
   async _create(
-    data: Partial<Data> | Partial<Data>[],
+    _data: Partial<Data> | Partial<Data>[],
     params: ServiceParams = {} as ServiceParams
   ): Promise<Result | Result[]> {
-    if (Array.isArray(data)) {
-      return Promise.all(data.map((current) => this._create(current, params)))
-    }
+    const payload = Array.isArray(_data) ? _data : [_data]
```
marshallswain, line 154: To skip the extra processing for large arrays, we can add another check for `params.bulk` before running `.map`:

```ts
const payload = Array.isArray(_data) ? _data : [_data]
const results = (params.bulk ? [] : payload).map((value) => {
  const id = (value as any)[this.id] || this._uId++
  const current = _.extend({}, value, { [this.id]: id })

  return _select((this.store[id] = current), params, this.id)
})

return params.bulk ? [] : Array.isArray(_data) ? results : results[0]
```

marshallswain: It looks like we also need to address the scenario where we create a single record and also pass `params.bulk`:

```ts
const david = await service.create(
  {
    name: 'David',
    age: 3,
    created: true
  }, 
  { bulk: true }
)
```

We don't want to return an empty array for the above scenario since that breaks the request-to-response plurality mapping of the service interface.  So we probably ought to throw an error:

```ts
if (!Array.isArray(_data) && params.bulk) {
  throw new BadRequest()
}
```

packages/adapter-commons/src/service.ts
```diff
@@ -57,6 +57,10 @@ export abstract class AdapterBase<
    * @returns Wether or not multiple updates are allowed.
    */
   allowsMulti(method: string, params: ServiceParams = {} as ServiceParams) {
+    if (params.bulk) {
```
marshallswain, line 60: Maybe we can update `allowMulti` to accept `context.id` and have a centralized place to throw an error whenever `params.bulk === true` and

- For create: `data` is an object
- For patch and remove: `context.id` is null

marshallswain (changes requested):

e-kibet:
Lets also solve the merge conflicts

MarcGodard:
Any ETA on this? Running to issues where this would help.

MarcGodard:
This is more than a year old, can I make the requested changes?

PR #2950, opened 2022-12-23
docs: use deep outline for typebox sidebar

marshallswain:
Having the deep outline makes it much easier to navigate the massive TypeBox API. Mobile devices only see a short table of contents, but for larger screens, it’s much easier to see what’s available when the sidebar shows the full tree. Also, there’s so much in the document that it’s easy to get lost.  The deep outline solves that.

Screenshot 2022-12-29 at 7 38 37 PM

### DIFF
--- a/docs/api/schema/typebox.md
+++ b/docs/api/schema/typebox.md
@@ -35,6 +35,206 @@ const messageSchema = Type.Object(
 type Message = Static<typeof messageSchema>
 ```
 
+## Result and data schemas
+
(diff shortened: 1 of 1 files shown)


daffl:
I think I changed this because it was pushing down important information like the query schema and validator docs. Is there a way to define the level of nesting?

marshallswain:
@daffl There is a way, but it's broken and puts all levels into a single list without indentation.  I think we need to reorder the page to put the types last.

marshallswain:
@daffl, I've updated the page to put the Types last, so they don't push down other important stuff.

marshallswain:
I've updated the screenshot.

daffl:
That works. Thanks!

PR #2956, opened 2022-12-27
docs: Improve explanation for queryResolver

AshotN:
An attempt to provide for information about `queryResolver` and how it works and can be used


AshotN:
Here is an example I also built with Codesandbox. 

https://codesandbox.io/p/sandbox/feathers-query-resolver-5dmftg?file=%2Freadme.md

docs/api/schema/resolvers.md
```diff
@@ -355,6 +361,49 @@ app.service('users').hooks({
 })
 ```
 
+For a more complicated example. We will make a seperate `queryResolver`, called `companyFilterQueryResolver`, that will act as a ownership filter. We will have a `Company` service that is owned by a `User`. We will assume our app has two registered users and two companies. Each user owning one company. For simplicity, `User1` owns `Company1`, and `User2` owns `Company2`
```
marshallswain: "separate" instead of "seperate"

marshallswain (approved):
This looks ready to go after the typo is fixed.

PR #2975, opened 2023-01-06
chore(docs): Provide HTTPS setup for Koa

8BallBomBom:
Simple copy from the Express docs but made compatible.
Inline with the docs over [here](https://koajs.com/#app-listen-).


daffl:
This is great, thank you for adding that!

PR #3001, opened 2023-01-23
fix(typescript): Make main type names more discernible

daffl:
This pull request changes the built in type names so that they are more easily to differentiate from the actual application type names. The problem was that TypeScript IDEs ended up auto importing the generic `Application` and `HookContext` types instead of the application specific ones so you don't get the correct service and configuration types.


daffl:
This actually requires a few more changes, might revisit this later if it turns out to be a continued problem for people.

PR #3004, opened 2023-01-23
fix(core): Run after all hooks first, and then after method hooks

### LINKED ISSUE #3002: After all and after method specific hook order changed in V5
### Steps to reproduce

If a service has the following hooks:

```js
after: {
  all: [context => console.log('After all')],
  find: [context => console.log('After find')],
}
```

In V4 it outputs:

```
After all
After find
```

In V5 it outputs:

```
After find
After all
```

My best current understanding is that it's because of how the before/after/error hooks get collected in a single "collected" chain using the new around hook mechanic here: https://github.com/feathersjs/feathers/blob/5854dea7f610262121a49623ec5bbd474dcd3ef3/packages/feathers/src/hooks.ts#L60-L65

What happens is that the when we spread all and then method specific hooks:

```js
...(collected.all || []),
...(collected[method] || [])
```

The order for the after hooks becomes:

afterAll() {
  await next()
  context => console.log('After all')
}

afterFind() {
  await next()
  context => console.log('After find')
}

and when that unwinds we first yield after find and then after all in that (backwards incompatible) order.

### Expected behavior

V5 keeps backwards compatibility in the hooks and logs:

```
After all
After find
```

### System configuration

**Module versions**:
@feathersjs/feathers@5.0.0-pre.35

KidkArolis:
### Summary

Fix #3002 - bring back the "after all" and "after method" hook ordering from V4.

The implementation is not as neat as it was before (having just 1 set of collected hooks was a nice idea). Let me know if you have a better idea.

### DIFF
--- a/packages/feathers/src/hooks.ts
+++ b/packages/feathers/src/hooks.ts
@@ -26,7 +26,8 @@ type HookStore = {
   before: { [method: string]: HookFunction[] }
   after: { [method: string]: HookFunction[] }
   error: { [method: string]: HookFunction[] }
-  collected: { [method: string]: AroundHookFunction[] }
+  collected: { [method: string]: AroundHookFunction[] },
+  collectedAll: { before?: AroundHookFunction[], after?: AroundHookFunction[] }
 }
 
 type HookEnabled = { __hooks: HookStore }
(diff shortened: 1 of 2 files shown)


daffl (approved):

daffl:
Thank you for looking into this. I can't believe we didn't have a test covering that. I was trying a few things to see if there was a different fix but I think this will have to be it for now. The fun and not-so-fun fact here is that a lot of that code is for backwards compatibility for a very specific case which is being able to call `service.hooks` multiple times.

PR #3005, opened 2023-01-24
fix(schema): Ensure all types of nested data are securely dispatched

### LINKED ISSUE #3003: Virtual properties of nested fields do not use external resolver
### Steps to reproduce

Resolve a virtual property inside a virtual property. The property that is 2 levels deep will not use its external resolver.

Example of 1 level deep working on external request, this is the result from the `roomOwners` service:
```
{
    "id": 11,
    "roomId": "11",
    "userId": "26",
    "user": {
        "id": 26,
        "tenantId": "26",
        "email": "test@test.com",
        "auth0Id": null,
        "name": null,
        "avatar": null
    }
}
```

Example of 2 levels deep not working on external request, this is the result from the `rooms` service:
```
{
    "id": 11,
    "name": "testroom",
    "description": "Test room",
    "createdAt": "1674456349778",
    "updatedAt": "1674456349778",
    "creatorId": "26",
    "tenantId": "26",
    ...
    "owners": [
        {
            "id": 26,
            "tenantId": "26",
            "email": "test@test.com",
            "password": "$2a$10$qw1jal6sxx6jvZO8LTgR/udtgYthA4ShCIq2hUbOWpuzsdZkEjHbO",
            "auth0Id": null,
            "name": null,
            "avatar": null
        }
    ]
}
```

Note, the password field is visible on the `user` property nested inside the `owners` property in the `rooms` service.

The resolver in roomOwners is here:
```
user: virtual(async (roomOwner, context) => {
    return context.app.service('users').get(roomOwner.userId);
}),
```

The resolver for the rooms is here:
```
owners: virtual(async (room, context) => {
    const { data } = await context.app.service('roomOwners').find({
        query: {
            roomId: room.id
        }
    });

    return data?.map((owner) => owner.user);
}),
```
Simple repository that reproduces this can be found here: https://github.com/edumeet/edumeet-management-server

### Expected behavior

For the external resolvers to run on nested properties.

### Actual behavior

The resolvers do not run.

`^5.0.0-pre.35`

**NodeJS version**:
`18.13.0`

daffl:
Closes #3003

### DIFF
--- a/packages/schema/test/fixture.ts
+++ b/packages/schema/test/fixture.ts
@@ -121,14 +121,24 @@ export const messageResolver = resolve<Message, HookContext<Application>>({
       throw new GeneralError('This is an error')
     }
 
-    return context.app.service('users').get(userId, context.params) as Promise<Message['user']>
+    const {
+      data: [user]
+    } = (await context.app.service('users').find({
+      ...context.params,
+      paginate: { default: 2 },
+      query: {
+        id: userId
+      }
+    })) as any
+
+    return user as Message['user']
   }),
   userList: virtual(async (_message, context) => {
     const users = await context.app.service('users').find({
       paginate: false
     })
 
-    return users as any
+    return users.map((user) => user) as any
   }),
   userPage: virtual(async (_message, context) => {
     const users = await context.app.service('users').find({
--- a/packages/schema/src/hooks/resolve.ts
+++ b/packages/schema/src/hooks/resolve.ts
@@ -121,7 +121,7 @@ export const resolveResult = <H extends HookContext>(...resolvers: Resolver<any,
 
 export const DISPATCH = Symbol('@feathersjs/schema/dispatch')
 
-export const getDispatchValue = (value: any, fallback = value): any => {
+export const getDispatchValue = (value: any): any => {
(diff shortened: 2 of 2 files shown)


packages/schema/test/fixture.ts
```diff
@@ -121,14 +121,24 @@ export const messageResolver = resolve<Message, HookContext<Application>>({
...
+
+    return user as Message['user']
   }),
   userList: virtual(async (_message, context) => {
     const users = await context.app.service('users').find({
       paginate: false
     })
 
-    return users as any
+    return users.map((user) => user) as any
```
marshallswain, line 141: What's going on here?  Seems this line is not needed.

marshallswain (approved):
Looks good after that one line in the test is fixed.

packages/schema/test/fixture.ts
daffl: It was part of the test. Basically the `users` array did have the dispatch information but not the individual user entries so you ended up getting the unsafe data if you e.g. grabbed a user from the array or mapped it like in this case.

marshallswain: Oh I see.

marshallswain:
:shipit:

PR #3008, opened 2023-01-24
feat(generators): Add service file for shared information

### LINKED ISSUE #2951: serviceName in cli to avoid hickups
I was breaking my head over some strange errors, 
`` "params": { "additionalProperty": "email"  }, "message": "must NOT have additional properties"}``

I was tarting to blame ... until, on Discord @marshallswain gave me the tip, in my "defense", just 5 minutes after I found it myself ;))):  I was applying hooks from one service to another service... oops.

The suggestion is now to introduce in the cli a constant ``servicePath``, one place change if you are cut and pasting the code to create a new service manually:

```ts
const servicePath = 'users' // the suggestion

export const user = async (app: Application) => {
  app.use(servicePath, new UserService(getOptions(app)), {
    methods: ['find', 'get', 'create', 'update', 'patch', 'remove'],
    events: []
  })

  app.service(servicePath).hooks(hooks)

```
The disadvantage is that it is slightly less nice to the eye... but well...just an idea ;)

daffl:
This pull request adds a `<service>.shared` file to a generated application which contains information that is shared between the client and the server:

- A `<service>Path` which contains the actual service path. This allows to change the path in a single place (as noted in #2951)
- A `<service>Methods` which contains a list of externally available methods. This allows to easily add or remove an external service method without having to update it in multiple places
- Exports of all schema types
- The client service registration (in line with the normal app service registration)

A shared service file looks like this:

```ts
// For more information about this file see https://dove.feathersjs.com/guides/cli/service.shared.html
import type { Params } from '@feathersjs/feathers'
import type { ClientApplication } from '../../client'
import type { Message, MessageData, MessagePatch, MessageQuery, MessageService } from './messages.class'

export type { Message, MessageData, MessagePatch, MessageQuery }

export const messagePath = 'messages'

export const messageMethods = ['find', 'get', 'create', 'patch', 'remove'] as const

export type MessageClientService = Pick<MessageService<Params<MessageQuery>>, (typeof messageMethods)[number]>

export const messageClient = (client: ClientApplication) => {
  const connection = client.get('connection')

  client.use(messagePath, connection.service(messagePath), {
    methods: messageMethods
  })
}

// Add this service to the client service type index
declare module '../../client' {
  interface ServiceTypes {
    [messagePath]: MessageClientService
  }
}
```

This file can also be used to e.g. define common helpers and schemas that can be used on both sides to e.g. do client side validation.

Closes https://github.com/feathersjs/feathers/issues/2951

### DIFF
--- a/docs/.vitepress/config.sidebar.ts
+++ b/docs/.vitepress/config.sidebar.ts
@@ -143,6 +143,10 @@ export default {
                     {
                       text: '📄 &lt;service&gt;.schemas',
                       link: '/guides/cli/service.schemas.md'
+                    },
+                    {
+                      text: '📄 &lt;service&gt;.shared',
+                      link: '/guides/cli/service.shared.md'
                     }
                   ]
                 },
--- a/docs/guides/cli/service.shared.md
+++ b/docs/guides/cli/service.shared.md
@@ -1,0 +1,18 @@
+---
+outline: deep
+---
+
+# Service Shared
+
+The `<service>.shared` file contains variables and type declarations that are shared between the [client](./client.md) and the [server application](./app.md). It can also be used for shared utility functions or schemas (e.g. for client side validation).
+
+## Variables
+
+By default two shared variables are exported:
+
+- `<name>Path` - The path of the service. Changing this will change the path for the service in all places like the application, the client and types
+- `<name>Methods` - The list of service methods available to the client. This can be updated with service and custom methods a client should be able to use.
+
+## Client setup
+
+This file also includes the client side service registration which will be included in the [client](./client.md). It will register a client side service based on the shared path and methods.
--- a/packages/generators/package.json
+++ b/packages/generators/package.json
@@ -43,7 +43,7 @@
   "scripts": {
     "prepublish": "npm run compile",
     "compile": "shx rm -rf lib/ && tsc && shx cp -r src/. lib/",
-    "test": "mocha --config ../../.mocharc.json --recursive test/**.test.ts test/**/*.test.ts"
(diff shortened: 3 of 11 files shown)


marshallswain:
I definitely prefer to group the types and variables, like this:

```ts
// For more information about this file see https://dove.feathersjs.com/guides/cli/service.shared.html
import type { Params } from '@feathersjs/feathers'
import type { ClientApplication } from '../../client'
import type { Message, MessageData, MessagePatch, MessageQuery, MessageService } from './messages.class'

export type { Message, MessageData, MessagePatch, MessageQuery }
export type MessageClientService = Pick<MessageService<Params<MessageQuery>>, (typeof messageMethods)[number]>

export const messagePath = 'messages'
export const messageMethods = ['find', 'get', 'create', 'patch', 'remove'] as const

export const messageClient = (client: ClientApplication) => {
  const connection = client.get('connection')

  client.use(messagePath, connection.service(messagePath), {
    methods: messageMethods
  })
}

// Add this service to the client service type index
declare module '../../client' {
  interface ServiceTypes {
    [messagePath]: MessageClientService
  }
}
```

Also, it's worth noting that the JS version will look like this:

```js
export const messagePath = 'messages'
export const messageMethods = ['find', 'get', 'create', 'patch', 'remove']

export const messageClient = (client) => {
  const connection = client.get('connection')

  client.use(messagePath, connection.service(messagePath), {
    methods: messageMethods
  })
}
```

Filenames we talked about:

- `service.shared.ts`
- `service.client.ts`
- `service.common.ts`

After reviewing it further, I think `service.shared.ts` makes the most sense.

marshallswain (approved):

PR #3010, opened 2023-01-25
fix(schema): Fix TypeBox extension value query syntax inference

daffl:
We noticed the error because it was just giving a `{ [string]: unknown }` type for nested query properties.


marshallswain:
:shipit:

marshallswain (approved):

PR #3015, opened 2023-01-27
docs(guides): Update migration to mention removal of req.authentication

KidkArolis:


daffl (approved):

daffl:
I wonder when this changed but if it is available in `req.feathers` already then that seems like the right place to mention that. Thank you!

PR #3019, opened 2023-01-27
fix(databases): Improve documentation for adapters and allow dynamic Knex adapter options

### LINKED ISSUE #2718: [feathers-dove] - How to perform transactions?
I'm just beginning with Feathers 5.x pre-release using Postgres, and although I can find support for transactions inside the feathersjs code (inside the knex/hooks.ts file), it's not clear to me from the documentation how I can approach this as an end user.

I couldn't find anything in the documentation either. Is there an example somewhere?

Thank you for your hard work on this!

### LINKED ISSUE #2791: How to provide unique index property in feathers dove schema just like feathers dove mongoose models
### Steps to reproduce

Give unique:true to any property in any schema. 

### Expected behavior
It should work like it used to in previous feathers verion with 
### Actual behavior
Gives error
 Error: strict mode: unknown keyword: "unique"

### System configuration

**NodeJS version**:

**Operating System**:

**Browser Version**:

**React Native Version**:

**Module Loader**:

### LINKED ISSUE #2989: .returning() is not supported by mysql and will not have any effect.
Randomly came across this warning.
Seems [the](https://github.com/feathersjs-ecosystem/feathers-knex/issues/209) issue that was reported and fixed before Knex was moved over is occurring again.

daffl:
This pull request reviews and improves the MongoDB and Knex adapter documentation.

For the MongoDB adapter:

- Fixes documentation for transactions
- Adds documentation for adding indexes
- Updates the data handling section to focus on ObjectIds with examples that can be dropped directly into a generated application
- Changes date handling suggestion to use timestamps instead of date strings and objects
- Removes the "common mistakes" section (this will eventually be addressed in #2820)

For the SQL Knex adapter:

- Updates all the documentation to be in line with the current implementation
- Adds documentation for missing methods
- Adds documentation for associations and related entries
- Fixes the setting of dynamic options

Addresses the following issues:

- Closes https://github.com/feathersjs/feathers/issues/2791
- Closes https://github.com/feathersjs/feathers/issues/2718
- Closes https://github.com/feathersjs/feathers/issues/2989


packages/adapter-tests/src/methods.ts
```diff
@@ -591,6 +591,22 @@ export default (test: AdapterMethodsTest, app: any, _errors: any, serviceName: s
         await service.remove(data[idProp])
       })
 
+      test('.create + query', async () => {
```
marshallswain: I can't figure out what we're testing, here?  Is it that create ignores the query? If so, maybe we can update the description to `.create + query ignores query`.

packages/knex/src/hooks.ts
```diff
@@ -8,7 +8,10 @@ const debug = createDebug('feathers-knex-transaction')
 const ROLLBACK = { rollback: true }
 
 export const getKnex = (context: HookContext): Knex => {
-  const knex = context.service.Model
+  const knex =
+    typeof context.service.getModel === 'function'
```
marshallswain: won't this conditional always be truthy since the`getModel` method is defined in `knex/src/adapter.ts`?

marshallswain (approved):

packages/knex/src/hooks.ts
daffl: You're right. It was a on of the dummy test services that still had a `.Model` which is not a good reason to keep the fallback.

PR #3021, opened 2023-01-28
feat(databases): Add $and to the query syntax

### LINKED ISSUE #2988: `$and` needs to be manually added and needs an ability to nest an `$or` (and vice versa)
### Steps to reproduce

I'm using `feathers-casl` which injects an `$and` filter in the query. If it gets down to the `filterQuery()` in `adapter-commons` the `$and` throws an error: "Invalid filter value $and". 

I have to manually add it in the `.class.js` file to make it work.

```
export class MyCustomService extends MongoDBService {}
export const getOptions = (app) => {
  return {
    paginate: app.get('paginate'),
    Model: app.get('mongodbClient').then((db) => db.collection('my-custom-service')),
    filters: { $and: true } <- HERE
  }
}
```

> If validated against a query schema this additional configuration should no longer be necessary. I'll have to double check if that is working. @daffl

--- 

Also, for this to work, `$and` and `$or` need to be infinitely nest-able within the operator schemas. `$and` needs to be a valid property of `$or` and vice versa. It is possible to have a query like:
```
$and: [
    $or: [
        $and: []
     ]
 ]
```

Per @fratzinger 

_$and also works for all adapters. It's highly needed for permissions. For something like:_
```
{
  $and: [
    { id: $in: [1, 2, 3] }, // <-- added from the client
    { id: $nin: [1, 3] } // <-- added from the server side permission system
  ]
```
_For simple queries you can always work around that with comparing values, exclude values from $in or overwrite a value. But there are always use cases where only a simple $and can help._

daffl:
It is already supported by most adapters so just needed to be added to the documentation and filters.

Closes https://github.com/feathersjs/feathers/issues/2988


marshallswain (approved):
:shipit:

PR #3022, opened 2023-01-28
fix(databases): Ensure that query sanitization is not necessary when using query schemas

daffl:
This pull request makes sure that the legacy query sanitization is not used when the query has already been validated against a schema.

Related to https://github.com/feathersjs/feathers/issues/2988


marshallswain (approved):
:shipit:

PR #3023, opened 2023-01-28
fix(configuration): Add pool and connection object to SQL database default configuration

### LINKED ISSUE #2784: Documentation for knex and connections
If you are using a free postgress, or a low connection database, make sure to change your pool settings.

Actually, you should change the min either way.

This is my `postgresql.js` file. I will be moving these connection variable into my config files, this is only used for dev (this is why the connections are low), but you should adjust the max to your max connections as this will speed stuff up.

```javascript
import knex from 'knex'
import _ from 'lodash'

export const postgresql = (app) => {
  let config = app.get('postgresql')
  config = _.assign(config, {
    pool: {
      min: 0,
      max: 7
    }
  })
  const db = knex(config)

  app.set('postgresqlClient', db)
}
```

daffl:
Closes https://github.com/feathersjs/feathers/issues/2784

### DIFF
--- a/docs/guides/cli/databases.md
+++ b/docs/guides/cli/databases.md
@@ -17,6 +17,34 @@ Depending on the SQL database you selected, a `src/<database>.ts` file will be c
 const knex = app.get('<database>Client')
 ```
 
+The database pool size can be set in the [configuration](./default.json.md) like this:
+
+```json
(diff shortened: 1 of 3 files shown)


marshallswain (approved):
:shipit:

PR #3034, opened 2023-01-31
fix: generated app does not start when choosing JSON schema

w3cj:
### Summary

  - After generating an app with "JSON schema" definition format
  - Error `Property '$id' is missing in type` is thrown for the file `src/configuration.ts`
  - Error `Cannot redeclare block-scoped variable 'userDataSchema'.` is thrown for `src/services/users/users.schema.ts`
  - No
  - No


daffl (approved):

daffl:
Oof, sorry about that! Once again learned the lesson that if it isn't covered with tests it will break at some point 😆 I now added full JSON schema tests to the CI test matrix in https://github.com/feathersjs/feathers/pull/3035 and with your fix they all pass now, thank you!

PR #3041, opened 2023-02-05
feat(mongodb): Add Object ID keyword converter and update MongoDB CLI & docs

marshallswain:
Fixes two issues:

- [x] Make CLI install correct versions of database-client packages (like `mongodb` client v5)
- [x] Adds the `objectid` keyword as an export from the `mongodb` package. So now we can use it in our schemas to automatically convert ObjectIds with AJV:

```ts
Type.String({ objectid: true })
```


marshallswain:
:shipit:  It's my PR, so you'll have to review and approve.  I've reviewed everything you added.  Looks good.

daffl (approved):

PR #3060, opened 2023-02-15
feat(generators): Final tweaks to the generators

daffl:
This pull request fixes some last remaining things with the generated app and tries to make the steps and everything the generator does more explicit and adds some more descriptive hints in the prompts. 

- `npm create feathers` will now create a minimalist Feathers application and allows to pick if you want to generate a client or not. You are able to pick if a client should be generated and if you want to use a schema at all (though not recommended)
- `npm create authentication` needs to be called explicitly after generating the application
- Separate user and service schema templates were consolidated into a single template. It makes the template a little more complex but easier to just change everything in one spot.


marshallswain (approved):
:shipit:

PR #3061, opened 2023-02-16
fix(koa): Fix missing dependency on feathers

vonagam:
`@feathersjs/feathers` should be in `dependencies`, not in `devDependencies`, because the koa package uses feathers types in its own exported types: in `Application` type and in augmentation of koa's `ExtendableContext`.

### DIFF
--- a/packages/koa/package.json
+++ b/packages/koa/package.json
@@ -52,6 +52,7 @@
     "@feathersjs/authentication": "^5.0.0-pre.37",
     "@feathersjs/commons": "^5.0.0-pre.37",
     "@feathersjs/errors": "^5.0.0-pre.37",
+    "@feathersjs/feathers": "^5.0.0-pre.37",
     "@feathersjs/transport-commons": "^5.0.0-pre.37",
     "@koa/cors": "^4.0.0",
     "@types/koa": "^2.13.5",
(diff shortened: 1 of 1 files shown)


daffl (approved):

daffl:
Thank you. I wish there was a way to easily test for this but I don't think there is.

vonagam:
Yeah, nothing comes to mind unfortunately.

Does not help that npm is the most popular package manager and with `node_modules` even if a package is not listed as a dependency but actually present in that folder it will be resolved. So for most users it will not be a problem and will not get reported.

vonagam:
Hm... actually there should be a way to detect it. After lib folder is built a script can look for `require("$x")` and `from '$x'` strings with `$x` being a name from `devDependencies` field. That's it.

Just looked and found [depcheck](https://github.com/depcheck/depcheck). By default it includes `devDependencies`, so basic cli invocation will not do, but its default export `depcheck(rootDir, options, callback)` has `package` field in `options`. So it is possible to parse `package.json` then clear `devDependencies` and pass it to that function. But I don't see an appropriate place where such utility script will be put in feathers repo.

Oh, there is also [npm-check](https://github.com/dylang/npm-check) which is based on `depcheck` but does have `--production` flag to skip `devDependencies`, so maybe this one is a way to go.

PR #3075, opened 2023-02-24
fix(transport-commons): Fix dispatching of arrays

daffl:
Equivalent change from #3073


packages/transport-commons/test/socket/utils.test.ts
```diff
@@ -163,6 +163,25 @@ describe('socket commons utils', () => {
         dispatcher('testing', dummyChannel, dummyHook, data2)
       })
 
+      it('dispatches dispatch arrays properly', (done) => {
```
marshallswain, line 166: take out extra "dispatch"?

marshallswain (approved):
One optional, non-issue in the test name.  :shipit:

PR #3087, opened 2023-03-01
fix(core): Use Symbol.for to instantiate shared symbols

### LINKED ISSUE #3071: @feathersjs/express^5 has a reused Symbol() across pkgs, but not actually equal
### Steps to reproduce

1. Build a small Express based app with v5.
2. Modify the package.json to leverage aliases for v5 express and v5 core.
  ```json
   "dependencies": {
      "f5": "npm:@feathersjs/feathers@^5.0.0-pre.38",
      "f5_exp": "npm:@feathersjs/express@^5.0.0-pre.38"
    }
  ```
3. Modify your app imports to use the aliases:
  ```javascript
  const feathers = require('f5');
  const express = require('f5_exp');
  ``` 
4. Send a request to your sample service.
5. See 500 error:
```
TypeError: Cannot read properties of undefined (reading 'express')
    at /Users/neptune/Sites/personal/tester/node_modules/f5_exp/lib/rest.js:51:36
    at /Users/neptune/Sites/personal/tester/node_modules/f5_exp/lib/rest.js:12:32
    at Layer.handle [as handle_request] (/Users/neptune/Sites/personal/tester/node_modules/f5_exp/node_modules/express/lib/router/layer.js:95:5)
    at trim_prefix (/Users/neptune/Sites/personal/tester/node_modules/f5_exp/node_modules/express/lib/router/index.js:328:13)
    at /Users/neptune/Sites/personal/tester/node_modules/f5_exp/node_modules/express/lib/router/index.js:286:9
    at Function.process_params (/Users/neptune/Sites/personal/tester/node_modules/f5_exp/node_modules/express/lib/router/index.js:346:12)
    at next (/Users/neptune/Sites/personal/tester/node_modules/f5_exp/node_modules/express/lib/router/index.js:280:10)
    at /Users/neptune/Sites/personal/tester/node_modules/f5_exp/lib/authentication.js:16:20
    at /Users/neptune/Sites/personal/tester/node_modules/f5_exp/lib/authentication.js:8:32
    at Layer.handle [as handle_request] (/Users/neptune/Sites/personal/tester/node_modules/f5_exp/node_modules/express/lib/router/layer.js:95:5)
```

### Possible Root Cause

The error is caused by an empty `options` object here: 
https://github.com/feathersjs/feathers/blob/4fbbfff2a3c625f8e6929e5a09e2cf7b739ffe11/packages/express/src/rest.ts#L65-L66
This is returned by the `getServiceOptions` function, which is pretty straightforward:
https://github.com/feathersjs/feathers/blob/4fbbfff2a3c625f8e6929e5a09e2cf7b739ffe11/packages/feathers/src/service.ts#L38-L40

It does a property check on the service for a SYMBOL('@feathersjs/service').  

https://github.com/feathersjs/feathers/blob/4fbbfff2a3c625f8e6929e5a09e2cf7b739ffe11/packages/feathers/src/service.ts#L5

When I inspect in debugger, the data is there, but the SYMBOL is different from what was set as a property and what is being checked.

I've tried legacy bundling, but that didn't work either.  I suspect that the SYMBOL is being created in different contexts, and therefore doesn't reference the same object.  Seems like there's a dependency injection/inversion here from the express lib that could resolve this, but I don't fully understand the benefit of the SYMBOL usage.

Any guidance?

### How this was discovered.

I'm trying to setup automated testing across multiple Feathers libraries that currently work in 4 and need to be extended to 5.  Creating multiple branches for testing felt like an ugly solution since it would require a lot of pipeline revisions and tie the semVer of the library to the Feathers versions.  I explored `npm overrides`, `npm aliases`, and `legacy-bundling=true` to see if I could deduce the reason the SYMBOL was different in the express function.

### Expected behavior

Dove Express should return a normal response even when aliasing the libraries.

### Actual behavior

Dove Express throws an error related to unreferenced SYMBOL

### System configuration

**Module versions** 
  @feathersjs/feathers@^5.0.0-pre.38
  @feathersjs/express@^5.0.0-pre.38

**NodeJS version**:
- 16 & 18

**Operating System**:
- OSX

**Browser Version**:
- n/a

**React Native Version**:
- n/a

**Module Loader**:
- CommonJS

daffl:
This pull request ensures that all Symbols are shared from the global symbol registry so that it is possible to use different package versions.

Closes #3071

### DIFF
--- a/packages/adapter-commons/src/service.ts
+++ b/packages/adapter-commons/src/service.ts
@@ -7,7 +7,7 @@ import {
 } from './declarations'
 import { filterQuery } from './query'
 
-export const VALIDATED = Symbol('@feathersjs/adapter/sanitized')
+export const VALIDATED = Symbol.for('@feathersjs/adapter/sanitized')
 
 const alwaysMulti: { [key: string]: boolean } = {
   find: true,
--- a/packages/commons/src/index.ts
+++ b/packages/commons/src/index.ts
@@ -96,7 +96,7 @@ export function isPromise(result: any) {
 }
 
 export function createSymbol(name: string) {
-  return typeof Symbol !== 'undefined' ? Symbol(name) : name
+  return typeof Symbol !== 'undefined' ? Symbol.for(name) : name
 }
 
 export * from './debug'
--- a/packages/knex/src/error-handler.ts
+++ b/packages/knex/src/error-handler.ts
@@ -1,6 +1,6 @@
 import { errors } from '@feathersjs/errors'
 
-export const ERROR = Symbol('@feathersjs/knex/error')
+export const ERROR = Symbol.for('@feathersjs/knex/error')
 
 export function errorHandler(error: any) {
   const { message } = error
--- a/packages/schema/src/hooks/resolve.ts
+++ b/packages/schema/src/hooks/resolve.ts
@@ -110,7 +110,7 @@ export const resolveResult = <H extends HookContext>(...resolvers: Resolver<any,
   }
 }
 
-export const DISPATCH = Symbol('@feathersjs/schema/dispatch')
+export const DISPATCH = Symbol.for('@feathersjs/schema/dispatch')
 
 export const getDispatchValue = (value: any): any => {
   if (typeof value === 'object' && value !== null) {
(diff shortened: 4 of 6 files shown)


marshallswain (approved):
:shipit:

PR #3093, opened 2023-03-08
fix(koa): Replace koa-bodyparser with koa-body

8BallBomBom:
Been looking at possible ways to implement an upload service.

After looking over the current options it would appear either we use `koa-body` to allow handling of `multipart/form-data` or we roll the dice and find another package as `koa-bodyparser` doesn't support `multipart/form-data`.

There is also the option where each developer could import and use `koa-body` themselves.
But there is currently a type compatibility issue with the older `koa-bodyparser` package.
More info over [here](https://github.com/koajs/bodyparser/issues/150).

What i can say though is it seems `koa-body` functions as a more up to date feature full drop in replacement.
Might need some extra testing though.

### DIFF
--- a/docs/api/koa.md
+++ b/docs/api/koa.md
@@ -253,7 +253,7 @@ app.use(
 
 ### bodyParser
 
-A reference to the [koa-bodyparser](https://github.com/koajs/bodyparser) module.
+A reference to the [koa-body](https://github.com/koajs/koa-body) module.
 
 ### cors
 
--- a/packages/koa/package.json
+++ b/packages/koa/package.json
@@ -56,12 +56,11 @@
     "@feathersjs/transport-commons": "^5.0.0",
     "@koa/cors": "^4.0.0",
     "@types/koa": "^2.13.5",
-    "@types/koa-bodyparser": "^4.3.10",
     "@types/koa-qs": "^2.0.0",
     "@types/koa-static": "^4.0.2",
     "@types/koa__cors": "^3.3.1",
     "koa": "^2.14.1",
-    "koa-bodyparser": "^4.3.0",
+    "koa-body": "^6.0.1",
     "koa-compose": "^4.1.0",
     "koa-qs": "^3.0.0",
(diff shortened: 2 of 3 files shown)


daffl (approved):

caiqueacct:
Is possible to create a sample how to upload a file using koa-body? I can see the file being saved on the OS but I don't know how to get the filename saved using the context. This is how I configured:

```
app.use(
  bodyParser({
    multipart: true,
    jsonLimit: '10mb',
    formidable: {
      uploadDir: bucketUpload.upload
    }
  })
)

```

Thanks

8BallBomBom:
The way i'm handling this currently is by using the koa before hook in a service, check [this](https://feathersjs.com/api/koa.html#service-middleware).
`ctx.request.files` will return an array of files you're uploading.

As for the right way to implement things, that is a good question, currently i'm rocking a janky implementation of `@koa/router` to make sure only set routes allow uploads... otherwise every route will accept an upload.
Not really sure about the best method of implementation as of yet.

caiqueacct:
Thanks. That's what I ended up using too. This is the code that I'm using although I also don't know if it's the best implementation but it's working like charm.

```
async (ctx, next) => {
  const readableStream = await getObjectStorage(
    objectStorage,
    result.objectName
  );
  const res = ctx.res;
  ctx.response.attachment(result.originalFilename, { type: "inline" });

  await new Promise((resolve, reject) => {
    const writableStream = new WritableStream({
      write(chunk) {
        try {
          res.write(chunk);
        } catch (e) {
          console.error("Error writing ", e);
        }
      },
    });

    readableStream
      .pipeTo(writableStream)
      .then(() => {
        res.end();
        resolve("SUCCESS");
      })
      .catch((e) => {
        console.error("Erro", e);
        reject(e.message);
      });
  });
};
```

8BallBomBom:
Thanks for sharing, if i find a better way to handle things then i'll probably contribute towards the docs 👍🏻

PR #3098, opened 2023-03-09
docs: update schema/resolveResult to specify it only works as an around hook

joezappie:
### Summary
`resolveResult` no longer can be an `after` hook. 

Updated the docs to state it only works as an `around` hook.


daffl (approved):

daffl:
Thank you!

PR #3121, opened 2023-03-22
Update docs for those without feathers cli in dev/globals

FossPrime:
### Summary

This could work 100% of the time... but right now it only works IF you have feathersjs/cli installed globally or in the devDependencies. If you don't you'll have to know it's in `@feathersjs/cli` somehow. That's specially not obvious as the Getting Started guide uses `create-feathers` not the CLI explicitly.

Daffl could replace  https://www.npmjs.com/package/feathers with the CLI to make it work... Perhaps it's worth cross-publishing.

### DIFF
--- a/docs/guides/basics/services.md
+++ b/docs/guides/basics/services.md
@@ -124,6 +124,7 @@ If you went with the default selection, we will use **SQLite** which writes the 
 In our new `feathers-chat` application, we can create database backed services with the following command:
 
 ```sh
+# Provided by `@feathersjs/cli`
 npx feathers generate service
 ```
 


daffl:
If you follow the guide as intended that command should always work since a generated v5 app will always install the CLI as a `devDependency`.

daffl:
I'd still be curious if you ran into a problem with it? I know a few people had issues with the old CLI still being installed globally.

FossPrime:
I was guiding a newbie on Feathers CLI and he managed to get this while generating a service... not sure if it was because originally the app was created with custom service, or because of some SIGKILL interruption. I don't use the CLI much as 99% of the time, I'm making a prototype and don't want to deal with binary DB's, but do want a real adapter like feathers-memory or what feathers-NeDB used to be. If I see the issue again I'll report.

Screenshot 2023-03-21 14 51 27

PR #3127, opened 2023-03-25
Fix error on first line of NPM page

FossPrime:
Screenshot from 2023-03-24 23-05-02


fratzinger (approved):

daffl:
Good catch. I always keep forgetting to review the individual packages. Should we bring back the logo instead? I believe it's used in other places as well.

PR #3128, opened 2023-03-25
Sort all front-end clients list to remain client framework agnostic

FossPrime:
Minor, I know. But hear me out.

We are implying that we support or prefer angular over Vue in some way by not having an alphabetical list, or one that's not even sorted by github stars... Vue has [203k](https://github.com/vuejs/vue) + [36k](https://github.com/vuejs/core) stars, over angular's  [87k](https://github.com/angular/angular), which is technically lower than Node.js, but that's not a front-end framework (Unless you count webcontainers).

We can cutout Lit, Svelte and Astro as they did not exist when Feathers was released, and having 4 would violate the literary convention known as "rule of three's".

This PR also adds Angular and Vue where it's missing from a list mentioning other compatible Feathers clients.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@
 
 Feathers is a full-stack framework for creating web APIs and real-time applications with TypeScript or JavaScript.
 
-Feathers can interact with any backend technology, supports many databases out of the box and works with any frontend like React, VueJS, Angular, React Native, Android or iOS.
+Feathers can interact with any backend technology, supports many databases out of the box and works with any frontend like Angular/React/Vue, Android or iOS.
 
 # Getting started
 
--- a/docs/feathers-vs-meteor.md
+++ b/docs/feathers-vs-meteor.md
@@ -4,7 +4,7 @@ Both Feathers and Meteor are open source real-time JavaScript platforms that pro
 
 Feathers is community supported, whereas Meteor is venture backed and has raised 31.2 million dollars to date.
 
-Meteor only has official support for MongoDB but there are some community modules of various levels of quality that support other databases. Meteor has it's own package manager and package ecosystem. They have their own template engine called Blaze which is based off of Mustache along with their own build system, but also have guides for Angular and React.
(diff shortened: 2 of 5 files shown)


mkadirtan:
Why use / instead of , though?

FossPrime:
> Why use / instead of , though?

These two have similar meaning:

- For Angular/React/Vue clients.
- For Angular, React, Vue or other clients.

The less traditional grammar encourages that the enumeration be more loosely interpreted, than a binding comma separated list. By no means do we discourage TS clients exports use with Svelte, Nuxt, Observers and Vite clients… but we have to draw the line somewhere. And, TS supporting frontend clients is needlessly generic and take too long to process when reading.

The biggest benefit is probably how easy it is to go back and change it. It's more semantic and easier to spot and remember than a plain English comma separated sentence that could easily get misremembered. As long as we're consistent, that's all that matters.

I'm also open to using the Web Component frameworks terminology, bit again, it makes the docs much harder to read. Kind of like renaming JavaScript to ECMAScript, or calling Linux, GNU/Linux, which means GNU plus Linux.

mkadirtan:
> > Why use / instead of , though?
> 
> These two have similar meaning:
> 
> * For Angular/React/Vue clients.
> * For Angular, React, Vue or other clients.
> 
> The less traditional grammar encourages that the enumeration be more loosely interpreted, than a binding comma separated list. By no means do we discourage TS clients exports use with Svelte, Nuxt, Observers and Vite clients… but we have to draw the line somewhere. And, TS supporting frontend clients is needlessly generic and take too long to process when reading.
> 
> The biggest benefit is probably how easy it is to go back and change it. It's more semantic and easier to spot and remember than a plain English comma separated sentence that could easily get misremembered. As long as we're consistent, that's all that matters.
> 
> I'm also open to using the Web Component frameworks terminology, bit again, it makes the docs much harder to read. Kind of like renaming JavaScript to ECMAScript, or calling Linux, GNU/Linux, which means GNU plus Linux.

Thanks for the explanation. It makes sense, and I agree with you.

PR #3136, opened 2023-03-31
fix(authentication): Ensure authentication.entity configuration can be null

psi-4ward:
If the authentication does not need a (user)-service one can set `authentication.entity = null` which [disables service validation](https://github.com/feathersjs/feathers/blob/dove/packages/authentication/src/service.ts#L181) but the base authenticationSchema does not allow using `null` in a config-json.


daffl (approved):

daffl:
Great, thank you for the quick fix!

PR #3137, opened 2023-04-01
Add cookbook section to top-nav

psi-4ward:
image


psi-4ward:
Ping 😬

daffl:
Sorry about that. The reason why we haven't added it before is because most Cookbook entries haven't been updated to the latest version. I'm not sure what to do about that to be honest.

PR #3145, opened 2023-04-05
fix(knex): Only apply default order for MSSQL

### LINKED ISSUE #3047: Knex: Unnecessary default ORDER BY clause added in find()
### Steps to reproduce

1. Run app with knex debugging `DEBUG=knex:query npm run dev`
2. Call `find` with `paginate:false` for a KnexService with a potentially large database table. eg:
```ts
   const events = await this.app.service("event-log").find({
      paginate: false,
      query: {
        company_id: someCompanyId,
        created_at: { $gt: someDate },
      },
    });
```
3. Check resulting sql in the log, it contains an unexpected `ORDER BY` clause which could potentially make the sql query heavier to execute or affect query execution planning
`knex:query select "event_log".* from "event_log" where "company_id" = $1 and "created_at" > $2 order by "event_log"."id" asc`

### Expected behavior

User can control whether queries should be sorted in DB request or not for non-paginated queries. For paginated queries it makes sense to force a default sorting but surely it is not needed for non-paginated requests or am I missing something? At least there could be some way of opting out of DB sorting like if setting `$sort: null` or such.

### Actual behavior

`ORDER BY` clause is added to all select statements from `find` and there is no way to opt out of the sorting.
https://github.com/feathersjs/feathers/blob/dove/packages/knex/src/adapter.ts#L184

daffl:
Only MSSQL requires a default sort order. All other database should use the natural or an explicit sort order.

Closes https://github.com/feathersjs/feathers/issues/3047

### DIFF
--- a/packages/knex/src/adapter.ts
+++ b/packages/knex/src/adapter.ts
@@ -180,7 +180,7 @@ export class KnexAdapter<
     }
 
     // provide default sorting if its not set
-    if (!filters.$sort) {
+    if (!filters.$sort && builder.client.driverName === 'mssql') {
       builder.orderBy(`${name}.${id}`, 'asc')
     }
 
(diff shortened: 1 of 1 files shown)


marshallswain (approved):
:shipit:

PR #3146, opened 2023-04-05
fix(knex): Get by id and transactions should work with params.knex

### LINKED ISSUE #3084: service.get(id) looses the where clause if query created in hook
Ran into a weird issue, whenever I add a before get hook like this:

```
  before: {
    get: [
      async (context: HookContext<TeamMembershipsService>) => {
        const query = context.service.createQuery(context.params)
        context.params.knex = query
      }
    ],
   ...
  }
```
which basically should do nothing. I get error: `NotFound: No record found for id 'cb36a3b0-007c-4719-9298-dcdea66705f8'.`
That's because the query executed by knex looses the `where` clause and returns more than 1 entry:
`knex:query select "team_memberships".* from "team_memberships" order by "team_memberships"."id" asc`

When `get(id)` is invoked it trickles down to this line in `_find()` in `KnexAdapter`:
```
const builder = params.knex ? params.knex.clone() : this.createQuery(params)
```

Since I've already created the query, `createQuery` is not invoked, and the `id` filter doesn't get added.
The `id` query parameter is not added in my hook either when I call `createQuery(context.params)` because `id` is not present in params it's in `context.id`

Adding the id filter to the query in the hook helps:
```
async (context: HookContext<TeamMembershipsService>) => {
  const query = context.service.createQuery(context.params)
  query.where('id', context.id)
  context.params.knex = query
}
```

### Expected behavior
I expected the id filter to be appended to the query i create in the hook automatically

### Actual behavior
where id=abc filter doesn't get added, query fails if more than 1 row present in the table.

### System configuration
Dove 5.0.0

daffl:
This pull request makes sure that queries by id and transactions work when passing `params.knex`.

Closes https://github.com/feathersjs/feathers/issues/3084

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -208,12 +208,12 @@
       "optional": true
     },
     "node_modules/@aws-sdk/abort-controller": {
-      "version": "3.303.0",
-      "resolved": "https://registry.npmjs.org/@aws-sdk/abort-controller/-/abort-controller-3.303.0.tgz",
-      "integrity": "sha512-LzNzpeyTppcmV/6SAQI3T/huOkMrUnFveplgVNwJxw+rVqmqmGV6z6vpg+oRICRDcjXWYiSiaClxxSVvOy0sDQ==",
+      "version": "3.306.0",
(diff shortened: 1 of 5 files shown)


marshallswain (approved):
:shipit:

PR #3169, opened 2023-04-19
docs: Update resolvers.md

wlukorito:
### Summary

Docs for `@feathersjs/schema` show imports for `schemaHooks` but there's no such export. The package exports `hooks` instead. Following the docs would give the below error:

`"@feathersjs/schema" has no exported member 'schemaHooks`

Use named export `hooks` and alias it as `schemaHooks`


daffl (approved):

daffl:
Good catch, thank you!

PR #3186, opened 2023-05-03
fix(authentication-oauth): Update OAuth redirect to handle user requested redirect paths

engineertdog:
This PR fixes an issue with the redirect functionality for `authentication-oauth`. If a user had set `origins` in their configuration, and attempted to redirect to a different page on the client side by using a link in the format of `oauth/<name>/?redirect=my/path?`, the redirect would only send them to the path of the origin `myorigin/#access_token`.

The problem was the variable the `separator` variable used to determine how to build the resulting URL. An addition test was added to verify the issue. With the change, if the user has an origin of `http://localhost:8080` and a frontend application that uses OAuth with `oauth/<name>/?redirect=my/path`, they will be redirected properly to `http://localhost:8080/my/path?access_token`.

### DIFF
--- a/packages/authentication-oauth/src/strategy.ts
+++ b/packages/authentication-oauth/src/strategy.ts
@@ -96,7 +96,7 @@ export class OAuthStrategy extends AuthenticationBaseStrategy {
     }
 
     const redirectUrl = `${redirect}${queryRedirect}`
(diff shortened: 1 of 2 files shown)


daffl (approved):

daffl:
Great catch! Will go out with the next patch release.

PR #3195, opened 2023-05-21
fix(authentication-client): removeAccessToken throws error if storage not defined

foxhound87:
fix: removeAccessToken throws error if storage not defined

### DIFF
--- a/packages/authentication-client/src/core.ts
+++ b/packages/authentication-client/src/core.ts
@@ -152,7 +152,9 @@ export class AuthenticationClient {
     // For NotAuthenticated, PaymentError, Forbidden, NotFound, MethodNotAllowed, NotAcceptable
     // errors, remove the access token
     if (error.code > 400 && error.code < 408) {
-      const promise = this.removeAccessToken().then(() => this.reset())
+      const promise = (this.storage as StorageWrapper).storage
+        ? this.removeAccessToken().then(() => this.reset())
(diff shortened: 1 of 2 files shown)


packages/authentication-client/src/core.ts
```diff
@@ -152,7 +152,9 @@ export class AuthenticationClient {
     // For NotAuthenticated, PaymentError, Forbidden, NotFound, MethodNotAllowed, NotAcceptable
     // errors, remove the access token
     if (error.code > 400 && error.code < 408) {
-      const promise = this.removeAccessToken().then(() => this.reset())
+      const promise = (this.storage as StorageWrapper).storage
```
marshallswain: My vote would be to leave this line alone, and let it make the request to `removeAccessToken`.  Then update line 136 in `removeAccessToken` to read

```ts
    return this.storage?.removeItem(this.options.storageKey)
```

In other words, let's let `removeAccessToken` check for `storage` instead of the error handler.

marshallswain (commented):
Thanks for finding the problem and making the effort to open a PR. I think if we move the check for `storage` into `removeAccessToken` it might help us avoid this same scenario if `removeAccessToken` is ever called for a different reason/angle.

marshallswain (approved):

daffl:
This is a good catch, thank you!

PR #3205, opened 2023-06-03
Make service optional in App.use

FossPrime:
I'm not sure why this isn't being caught automatically... but we should definitely not skipLibCheck on things we control. If we depend on something with bad TS... perhaps we shouldn't. 

skipLibCheck should remain the default for userland of course.

Screenshot 2023-06-03 11 35 28

https://github.com/feathersjs/feathers-chat/blob/dove/feathers-chat-ts/src/app.ts#LL22C40-L22C40
https://stackblitz.com/edit/node-tnne2i?file=app.ts%3AL46

### DIFF
--- a/packages/feathers/src/declarations.ts
+++ b/packages/feathers/src/declarations.ts
@@ -265,7 +265,7 @@ export interface FeathersApplication<Services = any, Settings = any> {
    */
   use<L extends keyof Services & string>(
     path: L,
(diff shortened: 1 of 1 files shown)


daffl:
The middleware extensions are added by the Koa module but looking at the warning, it looks like it'll require the `esModuleInterop: true` TypeScript setting.

PR #3206, opened 2023-06-03
fix[authentication]: Support StackBlitz

FossPrime:
- On AuthenticationService.setup, The error suppression is necessary as without publish, you'll also get errors when login out.
- disables validation and signing of JWT. This works surprisingly well... nothing is lost feature wise, aside from allowing others to spoof login, but that's not a concern in StackBlitz as WebContainers can't be hosted publicly. WASI runtimes run on bare metal and would not have an issue or be affected by this code.
- performance impact is undetectable, as jwt.sign is a usertime function, and this boolean check is a computer time check of the fastest order.

---

P.S. I was working on a fine example for far too long, but the lack of ESM exports has ruined my day. 
https://stackblitz.com/edit/feathersjs-playground-veesb9?file=vite.config.ts

### DIFF
--- a/packages/authentication/src/core.ts
+++ b/packages/authentication/src/core.ts
@@ -199,7 +199,7 @@ export class AuthenticationBase {
   ) {
     const { secret, jwtOptions } = this.configuration
     // Use configuration by default but allow overriding the secret
-    const jwtSecret = secretOverride || secret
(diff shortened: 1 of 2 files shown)


FossPrime:
Mitigated upstream

PR #3214, opened 2023-06-14
fix(authentication): Export JwtVerifyOptions

Gabetd:
### Summary
Added JwtVerifyOptions to the list exports

- [ ] allows JwtVerifyOptions to be accessed without importing from auth/lib/core. 

### Other Information

very light bug fix


FossPrime (approved):
Looks good. Will be important if we reduce the API surface area in Feathers 6 with subpath exports

This used to be exported in feathers core v4... so should help upgraders.

daffl (approved):

PR #3223, opened 2023-06-22
fix(feathers): Paginated<Result> for ServiceMethods.find

fratzinger:
The ReturnType of `ServiceMethods['find']` was `Promise<Result | Result[]>`. The correct way is `Promise<Paginated<Result> | Result[]>`.

I also added it to `ServiceOverloads`.


fratzinger:
@daffl: I'm stuck with the error. The service `OAuthService` in `@feathersjs/authentication-oauth` does not follow the common service definition. `OAuthService['find']` returns `Promise<GrantResponse>`.

It has to follow the common service definition, which should be `Promise<Paginated<R> | R[]>`. That is what the current error is about.

see
Pipeline:
image
and Code:
image


How to continue?
1. use @ts-ignore to move on?
2. make `OAuthService['find']` to return an array or a paginated result?
3. rename `find` of `OAuthService` to anything else?

daffl:
I'm not sure if we can add the pagination options to the main interface - I believe that's the error that is happening since any service interface implementor now expects the pagination overloads to be implement.

PR #3230, opened 2023-07-07
fix(generators): Fix channel/service configuration order for Koa based apps

myriads-dev:
When creating a new app using Koa, the generated `app.ts` configured `app.configure(channels)` before `app.configure(services)`. This resulted in service references being unavailable in `channels.ts`.

This update to the template places the services configuration before channels, allowing services to be referenced in channels as expected.


daffl (approved):

daffl:
Thank you!

PR #3234, opened 2023-07-14
fix(koa): Ensure .teardown works without a server

### LINKED ISSUE #3224: @feathersjs/koa: teardown() fails if setup() was called but listen() was not called
### Steps to reproduce

The overridden `Application#teardown()` method in @feathersjs/koa assumes there is a server started and throws a TypeError if there is not one.

If I'm using the app in a migration, a CLI tool, etc. I might not need to start the HTTP server. I'd like to be able to call `setup()` and then `teardown()` when I'm done.

Seems to me this worked in v4 with express. Any reason not to add an `if (this.server) {}` to enable this usage?

### Expected behavior

- Call `app.setup()` to initialize the app
- Use the app
- Call `app.teardown()` to close out the app

### Actual behavior

- Calling `app.teardown()` without first calling `app.listen()` throws because `this.server` is undefined

### System configuration

@feathersjs/koa@5.0.4

**NodeJS version**:

v18.16.0

**Operating System**:

macOS 12.6.6

**Browser Version**:

**React Native Version**:

**Module Loader**:

daffl:
Closes https://github.com/feathersjs/feathers/issues/3224

### DIFF
--- a/packages/koa/src/index.ts
+++ b/packages/koa/src/index.ts
@@ -53,11 +53,16 @@ export function koa<S = any, C = any>(
     },
 
     async teardown(server?: any) {
-      return feathersTeardown
-        .call(this, server)
-        .then(
-          () => new Promise((resolve, reject) => this.server.close((e) => (e ? reject(e) : resolve(this))))
-        )
+      return feathersTeardown.call(this, server).then(
+        () =>
+          new Promise((resolve, reject) => {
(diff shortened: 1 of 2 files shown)


marshallswain (approved):

marshallswain (approved):

marshallswain:
:shipit:

PR #3246, opened 2023-07-21
fix(generators): use `export type` vs `export`

b-steel:
### Summary

Change `export { NextFunction } ` to `export type { NextFunction }` and change authentication generator to look for either version when inserting new code. 

   I get the following error in `src/declarations.ts` on the line with `export { NextFunction }`: 
  `Re-exporting a type when 'isolatedModules' is enabled requires using 'export type'.ts(1205)`.  

   If that suggestion is followed, then running `feathers generate authentication` fails since it is looking for the line `export { NextFunction }` to insert new code after/before.  

   Not that I can find
  No


daffl:
I believe this should work and I'm not entirely sure why the tests are failing. Could you merge with latest and see if that works?

b-steel:
@daffl Yeah I was a bit confused as well, the error from the tests was rather cryptic.  I merged the latest and it seems to be passing now.

daffl (approved):

daffl:
Thank you! Will go out with the next release.

PR #3253, opened 2023-07-28
remove lodash/emit and flatten

marshallswain:
Replace `lodash/flatten` with `Array.prototype.flat()`
Replaces `lodash/omit` with `_.omit` from the `@feathersjs/commons` package.


daffl:
It looks like something is failing. May just need to be updated or it's something else. Also, can we remove the Lodash dependency completely then?

daffl:
Ok, so I verified that this is legitimately failing because Lodash's `_.omit` supports `.` separated field name. I'm inclined to close this unless we can remove the dependency completely without loosing any functionality (which doesn't seem to be the case here).

PR #3260, opened 2023-08-09
docs: Fix example in Discord oauth cookbook

gustojs:
Both Discord and `@feathers/authentication-oauth` expect "scope" key in config file instead of "scopes".


gustojs:
There's a mistake in my commit title but not in the commit itself.
I confirm it has to be "scope", singular. It's plural now.

gustojs:
I've added another fix in second commit - we need to import `oauth`, not `expressOauth` from `@feathersjs/authentication-oauth`.

daffl (approved):

daffl:
Thank you!

PR #3264, opened 2023-08-21
fix(generators): Harden mongodb.js to reliably extract database from any connection string

jermsam:
The default connection for mongodb in mongodb.js fails when you try to connect with a connection string that uses mongodb-replicas sets  or clusters ... (eg: mongodb://localhost:27017,127.0.0.1:27018/mo-db?replicaSet=rs0) ... This is because 
new URL(connection).pathname.substring(1)
 fails with an is not a valid URL error .... I propose the following replacement to  mongodb.js as it will satisfy all most all formats of connection strings


daffl:
Thank you for putting this PR up (and sorry for the late reply). I wonder if we should make this a helper since we'd want the generated code to be as short as possible.

jermsam:
Let me optimize it to be as short as possible

jermsam:
it's now just one additional line of code from what it was originally

daffl (approved):

daffl:
Thank you for updating and sorry for the delay.

daffl:
Apologies again but I had to revert the merge. I thought the tests were failing because things weren't up to date with latest but the error was actually an invalid regular expression that I wasn't able to fix (see errors in https://github.com/feathersjs/feathers/actions/runs/7427523402/job/20213354620#step:8:3307). You can run all test locally by running `npm test` in the main repo.

Issue #3266, opened 2023-08-26
TypeError: Cannot read properties of undefined (reading 'sub')


MarcGodard:
### Steps to reproduce

- Create new project (then add authentication as per docs), setup google or github (probably others too) for oauth
- visit correct /oauth/google folder
- authorize the app
- app redirects
- get the following error:

```
error: TypeError: Cannot read properties of undefined (reading 'sub')
    at GoogleStrategy.getEntityQuery (/home/marcgodard/Documents/Github/gozamm-api/node_modules/@feathersjs/authentication-oauth/lib/strategy.js:29:41)
    at GoogleStrategy.findEntity (/home/marcgodard/Documents/Github/gozamm-api/node_modules/@feathersjs/authentication-oauth/lib/strategy.js:79:34)
    at GoogleStrategy.authenticate (/home/marcgodard/Documents/Github/gozamm-api/node_modules/@feathersjs/authentication-oauth/lib/strategy.js:118:44)
```

### Expected behavior

No error and user created

### Actual behavior

Above Error

### System configuration

Brand new feathers 5.0.8 project (javascript, npm, json schemas)

**Module versions** (especially the part that's not working): 5.0.8

**NodeJS version**: v16

**Operating System**: linux

MarcGodard:
In `Strategy.js` Changing this:

```
    async getEntityQuery(profile, _params) {
        console.log(profile)
        return {
            [`${this.name}Id`]: profile.sub || profile.id
        };
    }
    async getEntityData(profile, _existingEntity, _params) {
        console.log(profile)
        return {
            [`${this.name}Id`]: profile.sub || profile.id
        };
    }
```

To this (in my node_module files):

```
    async getEntityQuery(profile, _params) {
        console.log(profile)
        return {
            [`${this.name}Id`]: profile?.sub ?? String(profile.id)
        };
    }
    async getEntityData(profile, _existingEntity, _params) {
        console.log(profile)
        return {
            [`${this.name}Id`]: profile?.sub ?? String(profile.id)
        };
    }
```

Made it work for both Google and Github.

JonasTheGamer:
Hm, after applying your fix I'm now getting this error:
```
info: Feathers app listening on http://FlexibeleJonas2:3030
undefined
error: TypeError: Cannot read properties of undefined (reading 'id')
[...]

PR #3267, opened 2023-08-26
Update strategy.ts

MarcGodard:
### Summary

See https://github.com/feathersjs/feathers/issues/3266


daffl:
Thank you for the pull request. I wonder why the profile ends up being undefined - I don't think it should.

The problem with this solution is that if the profile is undefined, it would set `<provider>Id` to `'undefined'` for every user.

MarcGodard:
Not sure either. But either way this can be managed in the authentication code. If anything that might be the best solution and to remove this from the strategy?

PR #3271, opened 2023-09-06
fix(generators): Fix configure channels when not real-time app

DeplanqueMathis:
### Summary

When I create a new project with feathers cli the app configure "channels" even if I answer the app is not a real-time app.
The presence of this line make an error when I run the project wich said "channels is undefined".
I see the import of "channels" is conditioned by the presence of "websockets" but not the "app.configure".


daffl (approved):

daffl:
Great, thank you!

PR #3287, opened 2023-09-30
fix(schema): add typescript >=4.7 as peerDependency

Artaud:
Schema package doesn't compile when installed TS is below 4.7, giving cryptic errors that are hard to track down. See https://discord.com/channels/509848480760725514/1079914294323445975 for context

### DIFF
--- a/packages/schema/package.json
+++ b/packages/schema/package.json
@@ -73,5 +73,8 @@
     "shx": "^0.3.4",
     "typescript": "^5.2.2"
   },
+  "peerDependencies": {
+    "typescript": ">=4.7"
+  },
   "gitHead": "90caf635aec850550b9d37bea2762af959d9e8d5"
 }


Artaud:
This probably doesn't need any tests, should not be a breaking change (as projects without correct version TS do not compile anyway - AFAIK it doesn't compile even with `"skipLibCheck": true` in tsconfig.json).

Not sure if it needs any documentation changes, but I guess not.

The main reason for this change is (at least for me) that for Feathers projects that use the client lib on the frontend with Angular, the least version of Angular needed is 14, as 13 **needs** typescript < 4.7.
It is definitely not rare to be on a lower version currently, and projects that are cannot use Dove client lib.
However, when they try to update, the project doesn't compile, errors are cryptic, and there's no clear indication on what's going on. This change should make it clear.

daffl (approved):

daffl:
That makes sense. Thank you!

PR #3303, opened 2023-10-06
fix(core): context.path is now typed correctly

AshotN:
### Summary

Currently `app.service()` is typed to all the registered services, and since `context.path` is just a string doing `app.service(context.path)` in a hook leads to an error that string is not assignable to `<list of schemas>`

This issue resolve that by allow `context.path` to check against the service provided in the `HookContext`. If `HookContext` is not typed with a service, it will default back to all services. 

If you create a hook in your user service. Where the `context.path` should always be `user`. It should not be typed as such

If you create a global hook, `context.path` will be a intersection of all your registered services.


daffl (approved):

daffl:
That's a great change, thank you!

PR #3307, opened 2023-10-09
fix(memory): Ensure correct pagination totals

DaddyWarbucks:
This PR addresses the comments made at https://github.com/feathersjs/feathers/pull/2844.

When pagination is enabled, the total is now correct and essentially works as things used to. There are two changes from the original implementation. 

1 - We check `hasQuery` before running the matcher. This is something I mentioned in my first PR. It just means that for our most basic calls like `service.find()` that we don't match the whole array for no reason.

2 - We now also add `this.id` in `_select(value, params, this.id))`. This ensures that the record ID is always selected, and is more in line with how most of the the DB adapters work. We may or may not want this, but its something to note as you guys are moving this over to Wings.

### DIFF
--- a/packages/memory/src/index.ts
+++ b/packages/memory/src/index.ts
@@ -76,7 +76,6 @@ export class MemoryAdapter<
     const { query, filters } = this.getQuery(params)
 
     let values = _.values(this.store)
-    const total = values.length
     const hasSkip = filters.$skip !== undefined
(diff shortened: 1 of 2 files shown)


daffl (approved):

daffl:
Thank you for fixing that! Will go out shortly.

PR #3344, opened 2023-11-21
chore: vitest

fratzinger:
migrate tests to vitest 

- add vitest to all packages
- rm mocha & ts-node from all packages
- change scripts accordingly
- rename `before`/`after` mocha-hooks to `beforeAll` & `afterAll`
- migrate from callbacks (deprecated in vitest) to Promises
- do not use static ports, use pkg 'get-port' instead


noor-tg (commented):
Will you also change the building & dev tools to vite ?

For reference, I am using vite & vitest now with feathers .

So if you need any help . I am here

daffl:
I'd be open to this if the tests all pass, it run faster and supports a watch mode where it runs a package test suite on code changes.

Personally I was debating of moving everything to the new built-in Node test runner to reduce dependencies and - now that it is stable - also have to worry less about version upgrades (which was finicky sometimes even with Mocha which hasn't really changed much). But it's still clunky to use in some cases. For example, it doesn't support running TypeScript tests recursively (except with a programmatic hack).

daffl:
I started using Vitest in another project and it is indeed working quite nicely and was the only test runner that worked with ES modules out of the box so this might also help us with the pains from #3375

fratzinger:
CI is broken for some reason. Only package that fails is '@feathersjs/knex' I think. But I don't know why. Please @daffl have a look.

PR #3353, opened 2023-11-27
fix(transport-commons): Allow case insensitive route lookups

### LINKED ISSUE #3330: Case sensitivity of path matching
### Steps to reproduce

- Launch a feathers app with at least one rest path, e.g. `/hello`
- Hit that path with some letters capitalized differently, e.g. `/HELLO`

### Expected behavior

Any capitalization of the path should hit the same feathers service.

It appears that paths were matched in a case-insensitive manner prior to feathers 5.

### Actual behavior

Exact capitalization is required. It does not appear to be configurable.

### System configuration

- We recently upgraded from  feathers 4.5.11 to 5.0.8.
- We found relevant code in `packages/transport-commons/src/routing/router.ts`, which is on 5.0.8.
- Node 16.14.0, although I can't imagine that's relevant.

### Code Sample

We're using express, and we tried playing with express properties to control case-sensitivity, e.g.
```
// app is the express app in this snippet, not a feathers app
app.set('case sensitive routing', true);
```
We played with several variants of constructing an express app and configuring it's path matching, then associating it with a feathers app. After we found `packages/transport-commons/src/routing/router.ts` it became clear that the path matching isn't using express' built-in behavior at all.

We found some text in this document https://feathersjs.com/guides/whats-new.html that references new path matching behavior, but no mention of case sensitivity. That links to https://feathersjs.com/api/application.html#lookup-path, but ditto.

We've worked around this by changing the path matching behavior in `packages/transport-commons/src/routing/router.ts`. Our snippet is slightly more complicated, but the relevant change is:
```
$ git diff
diff --git a/packages/transport-commons/src/routing/router.ts b/packages/transport-commons/src/routing/router.ts
index ad76319b..c46fe0b7 100644
--- a/packages/transport-commons/src/routing/router.ts
+++ b/packages/transport-commons/src/routing/router.ts
@@ -90,7 +90,10 @@ export class RouteNode<T = any> {
     }
 
     const current = path[this.depth]
-    const child = this.children[current]
+    const currentLowerCase = current.toLowerCase()
+    const child = Object.entries(this.children).find(
+      ([child]) => child.toLowerCase() === currentLowerCase
+    )?.[1]
 
     if (child) {
       const lookup = child.lookup(path, info)

```

daffl:
This pull request allows case insensitive route lookups by setting `app.routes.caseSensitive = false`.

Closes https://github.com/feathersjs/feathers/issues/3330

### DIFF
--- a/docs/api/application.md
+++ b/docs/api/application.md
@@ -155,6 +155,12 @@ lookup.service.find({
 })
 ```
 
+Case insensitive lookups can be enabled in the `app` file like this:
+
+```ts
+app.routes.caseSensitive = false
+```
+
 ## .hooks(hooks)
 
 `app.hooks(hooks) -> app` allows registration of application-level hooks. For more information see the [application hooks section in the hooks chapter](./hooks.md#application-hooks).
--- a/packages/transport-commons/src/routing/router.ts
+++ b/packages/transport-commons/src/routing/router.ts
@@ -115,10 +115,18 @@ export class RouteNode<T = any> {
 }
 
 export class Router<T = any> {
+  public caseSensitive = true
+
   constructor(public root: RouteNode<T> = new RouteNode<T>('', 0)) {}
 
   getPath(path: string) {
-    return stripSlashes(path).split('/')
+    const result = stripSlashes(path).split('/')
+
+    if (!this.caseSensitive) {
+      return result.map((p) => (p.startsWith(':') ? p : p.toLowerCase()))
+    }
+
+    return result
   }
 
   insert(path: string, data: T) {
--- a/packages/transport-commons/test/routing/index.test.ts
+++ b/packages/transport-commons/test/routing/index.test.ts
@@ -43,6 +43,14 @@ describe('app.routes', () => {
     assert.strictEqual(result.service, app.service('/my/service/'))
   })
 
+  it('can look up case insensitive', () => {
+    app.routes.caseSensitive = false
+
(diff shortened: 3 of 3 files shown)


marshallswain (approved):

jeffreythomasprice:
Is it intended that when case-insensitive mode is on, all path parameters are to be lower cased as well? I suspect that it is not, given the check for the `:` prefix in the new code.

The behavior I see is that incoming path parameters are lower cased before they make it to our code. Our path parameters happen to be case sensitive, although the rest of the path is not. (Now that I'm seeing this, that's maybe a strange design decision we should revisit...)

I have a service like:
```
app.use('/foo/:bar/:baz')
```

The new code in `packages/transport-commons/src/routing/router.ts` does this:
```
getPath(path: string) {
  return stripSlashes(path).split('/')
  const result = stripSlashes(path).split('/')

  if (!this.caseSensitive) {
    return result.map((p) => (p.startsWith(':') ? p : p.toLowerCase()))
  }

  return result
}
```

If I invoke this endpoint with:
```
/foo/Something/SomethingElse
```

`result` will then be:
```
['foo', 'Something', 'SomethingElse']
```

None of those elements starts with `:`, so they will all get lowered.

daffl:
Ah, the parameter name will be properly cased but not the value. That should indeed not be the case. The nodes themselves probably need to check it so this should be a follow-up issue to fix.

PR #3376, opened 2023-12-25
docs: validation for custom methods

AshotN:
### Summary

This PR expands the documentation on custom methods. Mainly an example on how to use validators with custom methods.

### DIFF
--- a/docs/api/schema/validators.md
+++ b/docs/api/schema/validators.md
@@ -115,3 +115,46 @@ app.service('messages').hooks({
   }
 })
 ```
+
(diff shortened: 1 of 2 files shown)


daffl:
Thank you! I think we should move this to the validators documentation at https://feathersjs.com/api/schema/validators.html but then we can merge.

marshallswain:
Thanks @AshotN for enabling PR branch editing.  I've pushed up the change that @daffl requested, so this should be ready to merge.

daffl (approved):

PR #3379, opened 2024-01-04
docs: Update anonymous hook to support around hooks

howlettt:
The current anonymous hook in the documentation doesn't work with around hooks since it doesn't have the `next` param
https://feathersjs.com/cookbook/authentication/anonymous.html
```
return async (context: HookContext) => {
```

This PR adds a optional `next` param so it can be used in both before & around hooks, making it consistent with the authenticate hook
https://feathersjs.com/api/authentication/hook
```
It can be used as a before or around hook.
```


daffl (approved):

daffl:
Thank you!

PR #3386, opened 2024-01-19
fix(generators): Move generators and CLI to featherscloud/pinion

daffl:
Moving to the official release of https://feathers.cloud/pinion, also moves at least the generators and CLI to ES modules.


marshallswain (approved):

marshallswain:
:shipit:

PR #3389, opened 2024-01-19
fix(knex): Add sqlite to returning clients

### LINKED ISSUE #3388: Add 'sqlite3' to the RETURNING_CLIENTS constant in KnexAdapter
> ## Summary:
> Add 'sqlite3' to the [@feathersjs/knex/src/adapter.ts:27](https://github.com/feathersjs/feathers/blob/4c65efcb668abf0879426a69200b647cdd1ea9f5/packages/knex/src/adapter.ts#L27)'s `RETURNING_CLIENTS` const list, in order to return the correct `id` column from INSERT statements

I'm a new user of feathers, and recently created my first project, and chose `sqlite` as the database since I was just toying around

I tried to change the primary key on the `user` and other generated service's tables to use a `uuid` instead of the incremental `integer`, but no matter what I tried to change in the migrations and schemas, the `create` route was always throwing a 404 NotFound, even if the row was correctly inserted:
```JSON
{
    "name": "NotFound",
    "message": "No record found for id '1'",
    "code": 404,
    "className": "not-found"
}
```

So my migration's `up()` was pretty much:
```typescript

export async function up(knex: Knex): Promise<void> {
  await knex.schema.createTable('users', (table) => {
    // changed from .increments() to .uuid()    
    table.uuid('id').primary().defaultTo(knex.fn.uuid())

    // changed usernameField from "email" to "username" at the correct places in local auth strategy and schemas
    table.string('username').unique()
    table.string('password')
  })
}
```

So I decided to look into the `_create(data, params)` method at the [KnexAdapter](https://github.com/feathersjs/feathers/blob/dove/packages/knex/src/adapter.ts) at `@feathersjs/knex` , and I realized the issue was probably related to this:

At [@feathersjs/knex/src/adapter.ts, lines 243 to 256](https://github.com/feathersjs/feathers/blob/4c65efcb668abf0879426a69200b647cdd1ea9f5/packages/knex/src/adapter.ts#L247):
```typescript
// This probably tells knex to add a "RETURNING <primaryKeyColumn>"
// clause (available at only a few db clients) at the end of the INSERT statement
const returning = RETURNING_CLIENTS.includes(client as string) ? [this.id] : []
const rows: any = await this.db(params)
  .insert(data, returning, { includeTriggerModifications: true })
  .catch(errorHandler)

// This one seems to try and pick the primary key from results...
const id = data[this.id] || rows[0][this.id] || rows[0]

if (!id) {
  return rows as Result[]
}

// ...to issue a SELECT statement and return the complete row
return this._get(id, {
  ...params,
  query: _.pick(params?.query || {}, '$select')
})
```

And at [line 27](https://github.com/feathersjs/feathers/blob/4c65efcb668abf0879426a69200b647cdd1ea9f5/packages/knex/src/adapter.ts#L27):
```typescript
const RETURNING_CLIENTS = ['postgresql', 'pg', 'oracledb', 'mssql']
```
So, the `sqlite` or `sqlite3` is missing from this list.

Then I looked into SQLite's official documentation and found out that:
1. SQLite [adds a ROWID column](https://www.sqlite.org/lang_createtable.html#rowid) containing an uniquely indexed `INTEGER` for all tables, by default
2. [the RETURNING clause is already implemented](https://www.sqlite.org/lang_returning.html) on SQLite, in the likes of PostgreSQL

It seems that `sqlite3` returns this `ROWID` like `[1]` by default when inserting, which is then used as the primary key for the subsequent `_get()` call, probably causing the NotFound error I've mentioned.

Then I tried altering this line at adapter.js to include `sqlite3` on the RETURNING_CLIENTS list:
```typescript
const RETURNING_CLIENTS = ['postgresql', 'pg', 'oracledb', 'mssql', 'sqlite3']
```

and _voilà_, got the expected response from calling the `create` method on the API:
```JSON
{
    "id": "124bfd37-xxxx-xxxx-xxxx-b78x0xx06cd6",
    "username": "some_random@email.com"
}
```
Therefore, I'd like to suggest simply including `sqlite` to the list of clients that implement the RETURNING clause, to correctly return created items at insert calls

Muliro1:
### Summary

here](https://github.com/feathersjs/feathers/blob/dove/.github/contributing.md#pull-requests))

### Other Information

If there's anything else that's important and relevant to your pull
request, mention that information here. This could include
benchmarks, or other information.

Your PR will be reviewed by a core team member and they will work with you to get your changes merged in a timely manner. If merged your PR will automatically be added to the changelog in the next release.

If this is a new feature, please remember to add the appropriate documentation in their respective pages in the `docs` folder.

Closes https://github.com/feathersjs/feathers/issues/3388

### DIFF
--- a/packages/knex/src/adapter.ts
+++ b/packages/knex/src/adapter.ts
@@ -24,7 +24,7 @@ const OPERATORS = {
   $ilike: 'ilike'
 }
 
-const RETURNING_CLIENTS = ['postgresql', 'pg', 'oracledb', 'mssql']
+const RETURNING_CLIENTS = ['postgresql', 'pg', 'oracledb', 'mssql', 'sqlite3']
 
 export class KnexAdapter<
   Result,


daffl (approved):

daffl:
Thank you, this will go out with the next release!

PR #3402, opened 2024-01-24
fix(generators): Use cross-platform ES module __dirname

daffl:
More ES module shenanigans because they actually don't work well 😞


fratzinger:
FYI: This is better since node v20.11.0 / v21.2.0. Once we deprecate lower versions in the future, we can switch to
```ts
import.meta.dirname
```
see https://twitter.com/wesbos/status/1745882792120955268
see https://github.com/nodejs/node/pull/48740

PR #3455, opened 2024-04-03
fix: reduce usage of lodash

fratzinger:
- import from lodash submodules directly where possible
- use `const { provider, ...rest } = params` instead of `omit`
- use `{ ...params, query: { ...params.query, provider } }` instead of `merge`
- rm useless `flatten`
- use `.filter(Boolean)` instead of `compact()`. _could be improved with a typesafe `IsNonNullable` function_
- use `() => {}` instead of `noop`


daffl (approved):

daffl:
Great, I've been meaning to look into this so thank you for doing that.

PR #3487, opened 2024-05-15
fix(schema): Allow regular functions in resolvers

ramijarrar:
Alters the expected return type of property resolvers and converters from `Promise<T>` to `Promise<T> | T`.

This ensures users are not forced to unnecessarily provide async functions with no await expression, which is often [considered bad practice](https://github.com/typescript-eslint/typescript-eslint/issues/7536#issuecomment-1719515791) and disallowed by linting rules.

### DIFF
--- a/docs/api/schema/resolvers.md
+++ b/docs/api/schema/resolvers.md
@@ -1,6 +1,7 @@
 ---
 outline: deep
 ---
+
 # Resolvers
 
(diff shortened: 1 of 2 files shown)


marshallswain (approved):
Thank you!

daffl:
This is a really good change. Will go out with the next release!

PR #3495, opened 2024-06-02
fix(adapter-commons): faster sorter

fratzinger:
This is a small refactoring.

- check for 'null && undefined' in one if
- detect typeof only once
- move get fn outside
- check key.includes('.') before key.split('.')
- type compareNSB & compareString

### DIFF
--- a/packages/adapter-commons/src/sort.ts
+++ b/packages/adapter-commons/src/sort.ts
@@ -1,17 +1,15 @@
 // Sorting algorithm taken from NeDB (https://github.com/louischatriot/nedb)
 // See https://github.com/louischatriot/nedb/blob/e3f0078499aa1005a59d0c2372e425ab789145c1/lib/model.js#L189
 
-export function compareNSB(a: any, b: any) {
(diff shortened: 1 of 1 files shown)


daffl (approved):

daffl:
This looks good, thank you. I wonder if there is a more generic package for this kind of thing. Doesn't seem to be something we should have to maintain ourselves - I took this from NeDB when it stopped being maintained.

fratzinger:
https://github.com/kofrasa/mingo could help. But I think it's fine. It's a small utility which is not a subject of change often and most adapters use `@feathersjs/adapter-commons` for other reasons anyways. It's good to have it in this package. 

I think we should not overcomplicate this. Or we end up importing and exporting from another library and their versioning has to align with our versioning. We have enough of these problems.

daffl:
True. It might be worth getting more test coverage for it eventually if we do want to keep it around.

fratzinger:
💯. I also thought about that.

Issue #3492, opened 2024-05-30
Resolver Performance


DaddyWarbucks:
I was clicking around the repo today and noticed this issue: https://github.com/feathersjs/feathers/pull/3487. It reminded me that awhile back I completed an issue for yup.js that I believe resolvers could also benefit from.

See: https://github.com/jquense/yup/issues/2052

The problem is that the resolvers treat ALL resolver functions as async and pushes/pops them onto the event loop. This means it schedules potentially sync work on the event loop. This can be solved by creating a more callback oriented flow that allows sync functions and async functions to be scheduled differently.

I don't think the performance benefits will be as great here as they were in yup.js. Yup generally has more "resolvers" and are more regularly sync. Feathers resolvers generally have fewer "resolvers" and are more commonly async. But, for large datasets I bet there would be a considerable performance gain.

DaddyWarbucks:
This would not require a change to resolver function parameters and they would continue to work as expected. 

Under the hood the resolvers would not be forced into a `Promise.all`. Instead the resolvers would be looped, synchronous resolvers would be called and then call a simple callback, and async functions would be called in a `Promise.all` where each function resolves with the callback. A counter of all the callbacks would indicate when the loop is done.

Long story short, overloading the event queue with synchronous work has a measurable impact. Not only on the the current request but anything else sharing the queue at that time. There were very impressive performance gains when I added this to yup. Resolvers could certainly benefit from this.

I am also curious if hooks could benefit from this approach. I haven't really poked around at hooks since async/koa style hooks were implemented. But that is another place where there is likely a bunch of hooks that are really synchronous but being treated as promises.

PR #3504, opened 2024-06-22
Feature: sync resolvers

DaddyWarbucks:
See: https://github.com/feathersjs/feathers/issues/3492

This PR is a WIP and a proof of concept. I have added a new `_resolve` method and added a test to compare the speed of this method compared to the current `resolve` method. The new `_resolve` method does not use `Promise.all` to resolve properties. Instead, it uses a basic for loop and done callback. This ensures resolvers that are not actually promises are not scheduled on the event loop. 

When running the tests for the package you will see some `console.time` results. I purposefully left the test configure in a manner that favored `syncResolvers`. This should accurately illustrate the problem with assuming all resolvers are promises. But you can configure these settings however you would like.

```js
const count = 10000
const asyncResolvers = 1
const syncResolvers = 10
const runs = 10

sync0: 86.879ms
sync1: 72.509ms
sync2: 171.684ms
sync3: 69.843ms
sync4: 67.955ms
sync5: 77.137ms
sync6: 75.034ms
sync7: 79.12ms
sync8: 77.653ms
sync9: 78.587ms

async0: 180.141ms
async1: 173.855ms
async2: 169.611ms
async3: 168.678ms
async4: 162.778ms
async5: 177.389ms
async6: 193.163ms
async7: 316.745ms
async8: 157.414ms
async9: 163.333ms
``` 

Interestingly, even when using all `asyncResolvers` the `_resolve` method is faster. I have noticed this when writing similar updates for yupjs and feathers-fletching. I speculate its because the callback pattern can pop promises off the loop as it goes and `Promise.all` has to keep them all in the loop until they are done? Pure speculation.

```js
const count = 10000
const asyncResolvers = 10
const syncResolvers = 0
const runs = 10

sync0: 258.305ms
sync1: 124.654ms
sync2: 123.168ms
sync3: 128.291ms
sync4: 126.634ms
sync5: 126.913ms
sync6: 131.84ms
sync7: 134.369ms
sync8: 125.799ms
sync9: 128.844ms

async0: 159.942ms
async1: 167.286ms
async2: 156.082ms
async3: 156.891ms
async4: 321.241ms
async5: 177.017ms
async6: 160.917ms
async7: 155.687ms
async8: 149.171ms
async9: 146.085ms
```

The `_resolve` method passes all tests but one. You can rename it to `resolve` (comment/rename real resolve method). It fails one test around virtuals. I played with it and couldn't get it, but I am confident its solvable.

I am also curious if `feathersjs/hooks` could benefit from a similar treatment. I am not sure how it would work with the next-able hooks, but that may even make it easier. This would allow entire Feathers applications to get a performance boost if hooks like `disallow`, etc aren't consuming the event loop.

### DIFF
--- a/packages/schema/src/resolver.ts
+++ b/packages/schema/src/resolver.ts
@@ -26,7 +26,7 @@ export const IS_VIRTUAL = Symbol.for('@feathersjs/schema/virtual')
  * @returns The property resolver function
  */
 export const virtual = <T, V, C>(virtualResolver: VirtualResolver<T, V, C>) => {
-  const propertyResolver: PropertyResolver<T, V, C> = async (_value, obj, context, status) =>
+  const propertyResolver: PropertyResolver<T, V, C> = (_value, obj, context, status) =>
     virtualResolver(obj, context, status)
 
   propertyResolver[IS_VIRTUAL] = true
@@ -90,12 +90,12 @@ export class Resolver<T, C> {
    * @param status The current resolver status
    * @returns The resolver property
    */
-  async resolveProperty<D, K extends keyof T>(
+  resolveProperty<D, K extends keyof T>(
     name: K,
     data: D,
     context: C,
     status: Partial<ResolverStatus<T, C>> = {}
-  ): Promise<T[K]> {
+  ): PromiseOrLiteral<T[K]> {
     const resolver = this.options.properties[name]
     const value = (data as any)[name]
     const { path = [], stack = [] } = status || {}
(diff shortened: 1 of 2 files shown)


DaddyWarbucks:
I know this needs to fix one test before its actually ready. But I wanted to get @daffl and @marshallswain eyes on it for some feedback. Ya'll let me know any thoughts/questions/concerns.

daffl:
Thank you for putting this up. It's kind of interesting that things like this don't seem to get optimized by V8.

PR #3513, opened 2024-07-10
fix(generators): Fix JavaScript generators

daffl:
It looks like the behaviour in TypeScript transpiler changed for this.


marshallswain (approved):
:shipit:

PR #3532, opened 2024-09-09
Update typebox

alidplus:
### Summary
1. Tell us about the problem your pull request is solving.
[@sinclairzx81/typebox](https://github.com/sinclairzx81/typebox) is an awesome project and it is working well with feathers. As there are many new and usable utils like `Type.Composite` in newer versions, I usually use to install newer versions to my projects. However, using two different versions creates a lot of problems with type mismatch.
The first solution to fix it is to use `resolutions` in the package.json file. but it causes some other problems.
Finally, I decided to update it.

2. Are there any open issues that are related to this?
No!

3. Is this PR dependent on PRs in other repos?
Only two small changes in @feathers/generators

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -25059,7 +25059,7 @@
       "license": "MIT",
       "dependencies": {
         "@feathersjs/schema": "^5.0.30",
-        "@sinclair/typebox": "^0.25.0"
+        "@sinclair/typebox": "^0.33.9"
       },
       "devDependencies": {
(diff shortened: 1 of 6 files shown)


daffl:
Thank you for the pull request. I branched off of it at https://github.com/feathersjs/feathers/pull/3612 for a working implementation for the next release.

daffl:
The update to TypeBox version 1 has now been done in https://github.com/feathersjs/feathers/pull/3616

PR #3537, opened 2024-09-19
fix #3057

### LINKED ISSUE #3057: result.total does not account for params.pipeline
### Steps to reproduce

pre.37, db: MongoDB

1. Set custom pipeline in a service find before hook like 

(context) => {
      context.params.pipeline = [{ $match: { foo: 'bar' } }]
      return context
}

2. Call find() with empty query

The result.data array is correct
result.total should be equal to the number of db records where foo is equal to 'bar', but it's equal to the total number of records as if no custom pipeline specified

alidplus:
### Summary
fixes #3057

### DIFF
--- a/packages/mongodb/src/adapter.ts
+++ b/packages/mongodb/src/adapter.ts
@@ -149,7 +149,7 @@ export class MongoDbAdapter<
     const model = await this.getModel(params)
     const { before, after } = this.parsePipeline(params)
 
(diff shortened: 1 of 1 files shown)


daffl:
Thank you for the pull request. I merged this via https://github.com/feathersjs/feathers/pull/3541 which I believe solves the same problem and also included tests.

PR #3539, opened 2024-09-23
fix(knex): Add tableOptions parameter for inheritance on knex adapter options to pass on knex builder

fidelio314:
Add support on knex adapter to passe tableOptions on the Knex builder in order to support the option  ONLY that discard the inheriting tables - **For PostgreSQL only**

Today the feathesjs-knex package was not able to pass the option:
```
{
  only: true
}
```
Here is the knex doc: https://knexjs.org/guide/query-builder.html#common

I propose a new knex specific adapter option called: tableOptions to be used on the getOptions method of the service class.

this can be passed on the db method on the knex adapter and therfore to the 

```
Model(name, tableOptions)
```

### DIFF
--- a/packages/knex/src/adapter.ts
+++ b/packages/knex/src/adapter.ts
@@ -69,7 +69,7 @@ export class KnexAdapter<
   }
 
   db(params?: ServiceParams) {
-    const { Model, name, schema } = this.getOptions(params)
+    const { Model, name, schema, tableOptions } = this.getOptions(params)
 
     if (params && params.transaction && params.transaction.trx) {
       const { trx } = params.transaction
@@ -77,7 +77,7 @@ export class KnexAdapter<
       return schema ? (trx.withSchema(schema).table(name) as Knex.QueryBuilder) : trx(name)
     }
 
-    return schema ? (Model.withSchema(schema).table(name) as Knex.QueryBuilder) : Model(name)
+    return schema ? (Model.withSchema(schema).table(name) as Knex.QueryBuilder) : Model(name, tableOptions)
   }
 
   knexify(knexQuery: Knex.QueryBuilder, query: Query = {}, parentKey?: string): Knex.QueryBuilder {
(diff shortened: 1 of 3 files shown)


packages/knex/src/adapter.ts
```diff
@@ -69,15 +69,15 @@ export class KnexAdapter<
...
+    const { Model, name, schema, tableOptions } = this.getOptions(params)
 
     if (params && params.transaction && params.transaction.trx) {
       const { trx } = params.transaction
       // debug('ran %s with transaction %s', fullName, id)
       return schema ? (trx.withSchema(schema).table(name) as Knex.QueryBuilder) : trx(name)
     }
 
-    return schema ? (Model.withSchema(schema).table(name) as Knex.QueryBuilder) : Model(name)
+    return schema ? (Model.withSchema(schema).table(name) as Knex.QueryBuilder) : Model(name, tableOptions)
```
e-kibet, line 80: Will it work if the tableOptions has not been provided, I would prefer we check it before passing the tableOptions as parameter on the knex adapter options like this:-

```
let queryBuilder: Knex.QueryBuilder;
  if (params?.transaction?.trx) {
    const { trx } = params.transaction;
    queryBuilder = schema ? trx.withSchema(schema).table(name) : trx.table(name);
  } else {
    queryBuilder = schema ? Model.withSchema(schema).table(name) : Model.table(name);
  }

  // Apply tableOptions if provided
  if (tableOptions) {
    queryBuilder = queryBuilder.withOptions(tableOptions); // Adjust this if .withOptions() isn't the exact method for tableOptions
  }
  return queryBuilder;
```

fidelio314: the 
`Model(name, tableOptions)`

has already the tableOptions that can undefined. (its tha case now)
i am just proposign to pass them and then allow to have this property reach the knex mode. 
no the ONLY postgres query is impossible to have though feathers.

daffl (approved):

daffl:
Ooops, sorry for some reason I thought the tests were failing but they just didn't have approval to run. Thank you for putting this up, it will go out with the next release.

PR #3541, opened 2024-10-10
fix(mongodb): Fix mongo count

DaddyWarbucks:
This PR fixes a bug in the mongo package that caused a miscount when counting documents while using the aggregation pipeline. I recently installed this in a large production app and caught a couple of bugs that the tests had not. 

0f1deb642ad4a4585c17aa2bf4eaec8fd7152ca0 - Add `paginate: false` to ensure counting all documents.

e6595b8d3c2dc6fbc32084fb980f3f85805b75f5 - Use the pipeline's `$count` operator for better performance.

e126577a47081733d18adb11a96e0fce3a6d93e7 - Rearrange the pipeline so that it skips and limits before projecting (aka select). Mongo supposedly optimizes the pipeline steps, but its still make sense to me and reads better if we skip/limit before projection.

927f357232dfd8f7cb946db4e67f4725dbb27e4e - Add 1 missing test from the generic test suite.

Please let me know anything I can do to move this PR through quickly. It is critical bug in a production app and I cannot roll it back easily.

### DIFF
--- a/packages/mongodb/src/adapter.ts
+++ b/packages/mongodb/src/adapter.ts
@@ -213,6 +213,7 @@ export class MongoDbAdapter<
     if (params.pipeline) {
       const aggregateParams = {
         ...params,
+        paginate: false,
         query: {
           ...params.query,
           $select: [this.id],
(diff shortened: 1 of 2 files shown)


DaddyWarbucks:
I was actually able to solve this problem in my own app by just extending the `MongoService` class with the updates from the PR.

daffl:
Thank you! I'll look into why the tests aren't passing today.

PR #3554, opened 2024-12-06
fix(generators): remove ctx.cwd from prettier.resolveConfig()

buxit:
fix prettier formating in generators

prettier.resolveConfig() should either be called with a concrete config file or without any parameters. with just a directory it fails to resolve correctly


daffl (approved):

daffl:
Thank you for the pull request. It looks like the types require a parameter which is why the compile step fails. What does the current failure look like?

buxit:
> Thank you for the pull request. It looks like the types require a parameter which is why the compile step fails. What does the current failure look like?

it doesn't pick up `.prettierrc` in the base directory, its contents are ignored. i see two options, calling `resolveConfig` with an empty string or with `await prettier.resolveConfigFile(ctx.cwd)` as argument.

the empty string works because there's a check for `!configFile` in prettier (https://github.com/prettier/prettier/blob/f66d8ec2e02ce95d41fcd0109c9143b1dc3c9b62/src/config/resolve-config.js#L33C3-L36C4)

daffl:
`{ resolveConfig: await prettier.resolveConfigFile(ctx.cwd) }` sounds like the more robust solution to me?

buxit:
> `{ resolveConfig: await prettier.resolveConfigFile(ctx.cwd) }` sounds like the more robust solution to me?

not sure `await prettier.resolveConfig(await prettier.resolveConfigFile(ctx.cwd))` will actually compile, as `resolveConfigFile(ctx.cwd)` could return `null`

e-kibet (approved):
LGTM

PR #3574, opened 2025-03-11
fix(oauth): return updated object `value` key instead of complex patched object that causes crash

francescochiapello:
OAuth is not working since the object returned from patch function is a complex object composed by `value` item that contains the payload of the user account.

There are not other open issues that are related to this and this PR is not dependent on PRs in other repos.

### DIFF
--- a/packages/authentication-oauth/src/strategy.ts
+++ b/packages/authentication-oauth/src/strategy.ts
@@ -166,7 +166,7 @@ export class OAuthStrategy extends AuthenticationBaseStrategy {
 
     const authEntity = !existingEntity
       ? await this.createEntity(profile, params)
-      : await this.updateEntity(existingEntity, profile, params)
+      : (await this.updateEntity(existingEntity, profile, params)).value
 
     return {
       authentication: { strategy: this.name },


francescochiapello:
Hi guys, without this PR merged, is not possible to handle oauth login using Facebook, Google or other third party provider. Please have a look at it and if possible merge it to fethers repo. Thanks a lot.

daffl:
Is there a way to confirm this. Aside from [the compilation failing](https://github.com/feathersjs/feathers/actions/runs/13797915130/job/39138659083?pr=3574#step:8:1077) I just tested the existing setup with Google and it worked as expected.

francescochiapello:
Hi @daffl, I've tested it multiple times with this version `"@feathersjs/authentication-oauth": "^5.0.31"` and when I register a user it's working well, but after trying to sign in again with Google, this line
```
return this.entityService.patch(id, data, _.omit(params, 'query'))
```
on this method

```
  async updateEntity(entity: any, profile: OAuthProfile, params: Params) {
    const id = entity[this.entityId]
    const data = await this.getEntityData(profile, entity, params)

    debug(`updateEntity with id ${id} and data`, data)

    return this.entityService.patch(id, data, _.omit(params, 'query'))
  }
```
is returning an object that has inside a `value` object with the mongodb loggedin user object.

francescochiapello:
Hi @daffl , this is what I mean with `value` object returned by path service.

google-values

daffl:
It looks like you are using MongoDB aggregation in your `users` service. By default the strategy calls `userService.patch()`. I'd recommend first making sure that the `patch` method returns the format expected by `updateEntity` by default or [customizing the implementation](https://feathersjs.com/api/authentication/oauth.html#customization) accordingly.

daffl:
Closing this since I think it is related to the MongoDB aggregation and works as intended in the default case. Appreciate the PR though!

PR #3578, opened 2025-03-21
feat(knex): Add support for extended operators in query builder (fixes #3377)

### LINKED ISSUE #3377: Support additional postgres operators
I have a postgres database that makes use of array fields and fulltext indexes. I need to be able to perform queries on those fields using the following postgres operators:
* && - array overlap - https://www.postgresql.org/docs/current/functions-array.html#ARRAY-OPERATORS-TABLE
* @> - array contains - https://www.postgresql.org/docs/current/functions-array.html#ARRAY-OPERATORS-TABLE
* <@ - array contained by - https://www.postgresql.org/docs/current/functions-array.html#ARRAY-OPERATORS-TABLE
* @@ - fulltext match - https://www.postgresql.org/docs/current/functions-textsearch.html

I made a PR for the v4 feathers-knex module implementing these: https://github.com/feathersjs-ecosystem/feathers-knex/pull/236
However, it wasn't immediately clear to me how to implement this in the new knex core module.

Also, reading the migration guide, it seems like perhaps these could be defined in my service definitions using the filter/operator properties? The documentation there wasn't very clear how these are meant to function.

michalgm:
This change adds the ability to define custom operators via a new `extendedOperators` option for the knex service constructor. This allows the use of  database-specific operators that aren't included in the standard set.

- Modified the `knexify` method to use both built-in and extended operators
- Added documentation for the `extendedOperators` option
- Added tests to verify the functionality with custom operators

Example usage:
```typescript
const service = new KnexService({
  Model: db,
  name: 'people',
  extendedOperators: {
    $startsWith: 'LIKE',
    $fulltext: '@@'  // PostgreSQL full-text search
  }
})
```

This fixes #3377

### DIFF
--- a/docs/api/databases/knex.md
+++ b/docs/api/databases/knex.md
@@ -63,7 +63,7 @@ The Knex specific adapter options are:
 - `name {string}` (**required**) - The name of the table
 - `schema {string}` (_optional_) - The name of the schema table prefix (example: `schema.table`)
 - `tableOptions {only: boolean` (_optional_) - For PostgreSQL only. Argument for passing options to knex db builder. ONLY keyword is used before the tableName to discard inheriting tables' data. (https://knexjs.org/guide/query-builder.html#common)
-- `extendedOperators {[string]: string}` (_optional_) - A map defining additional operators for the query builder. Example: `{ $fulltext: '@@' }` for PostgreSQL full text search.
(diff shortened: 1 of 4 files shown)


michalgm:
I'm now realizing another approach would be to just extended `OPERATORS` to support everything the knex supports: https://github.com/knex/knex/blob/9bd12999907436c2ef51f786df09a9a7e8931cca/lib/formatter/wrappingFormatter.js#L9

daffl (approved):

daffl:
Thank you! This is a good step I think.

Issue #369, opened 2016-07-26
HTML5 HTTP/2 SSE Transport Support


ubergarm:
I love that Feathers builds out API endpoints for services with both REST and websockets. It may be of interest to additionally support HTTP/2 SSE transports for services.

I searched for this issue but didn't find it yet despite it being mentioned 4 months ago on a hacker news discussion (referenced below).

Thanks!

https://hn.algolia.com/story/11290577/feathers-2-0-a-minimalist-real-time-javascript-framework
```
N nailer  - 4 months ago
> exposes both a RESTful and real-time API automatically through HTTP/HTTPS and over websockets.
Why not both at once with HTML5 SSE?

M Matthias247 - 4 months ago
I think SSE has two limitations that might not make it too useful for some scenarios:
- Each SSE channel would require a separate TCP connection. And as the connections for a browser are limited, you can not support a lot of of different SSE channels. This is mainly a HTTP/1.1 problem, HTTP/2 fixes this. But as browser and server support for HTTP/2 is still limited you might not bet on it.
- SSE can only send chunks of text in each notification, where the possible payload is additionally limited by the SSE encoding (you can't send text with <br /><br /> in it). So if you need binary or abritrary (JSON?) text objects you would need some additional encoding. Like object -> json string -> byte-array representation -> base64 string. This works, but the overhead for your application (must do the encoding) and the browser (must to all the decoding and look for the <br /><br />) is not ideal.
I'm currently experimenting with using HTTP/2 and fetch API with readable streams to stream arbitrary objects on a bigger number of parallel channels from server to client. This seems to work and looks quite promising for the future, but due to missing browser and partly also missing server support the solution is definitely in it's infancy.
So the current safe bet for enabling lots of notification channels or enabling realtime queries would still be websockets.

D daffl - 4 months ago
That sounds like something worth keeping track of. Do you want to create an issue at https://github.com/feathersjs/feathers/issues for it? The nice thing is that you most likely won't have to change your services when adding a new transport layer.
```

daffl:
From what I could see, it does not seem as if Express will support HTTP2 any time soon so this will be very dependent on #258.

ekryski:
I did do some spelunking in this area a couple weeks back. I guess you can get it to work with Express5 but who knows what is going on with that branch. [Here's an example from Stackoverflow](http://stackoverflow.com/questions/28639995/node-js-server-and-http-2-2-0-with-express-js).

[...]

PR #3601, opened 2025-08-15
feat: SSE real-time events

daffl:
With the web standard implementation of the HTTP transport and the JS generator support for service methods we can add a service that provides an endpoint for streaming service events as server sent events (SSE).

Closes https://github.com/feathersjs/feathers/issues/369


packages/feathers/src/client/sse.ts
```diff
@@ -0,0 +1,101 @@
...
+              }
+
+              try {
+                if (payload.path === options.path && payload.event === 'connected') {
+                  sseService.emit('connected', abortController)
+                } else {
+                  client.service(payload.path).emit(payload.event, payload.data)
+                }
+              } catch (error) {
+                console.error(error)
```
marshallswain: I think this is the only area you've left for improvement.

daffl: Good point. I was thinking we could also emit an `sseError` event on the SSE service.

marshallswain (commented):
**Issue: Library shouldn't log directly to console**

  **Current:**
  ```typescript
  try {
    client.service(payload.path).emit(payload.event, payload.data)
  } catch (error) {
    console.error(error) // ❌ Library shouldn't log directly
  }

  Suggested:
  try {
    client.service(payload.path).emit(payload.event, payload.data)
  } catch (error) {
    sseService.emit('error', error) // ✅ Let the app decide what to do
  }

  Why this matters:
  1. Library best practice - Applications should control their own logging/error
  reporting
  2. Event processing errors - This catches errors in user event handlers that apps may
  want to handle differently
  3. Consistency - SSE client already emits 'connected' and 'disconnected' events

  Unlike internal parsing errors (which can be silently handled), this error occurs when
  processing user events - something the application should be aware of and potentially
  handle (invalid service paths, malformed payload data, user handler exceptions, etc.).

PR #3604, opened 2025-09-01
feat: pluggable router architecture with 49% performance improvement

marshallswain:
# Router Architecture Overhaul & Performance Optimization

  ## 🚀 Performance Improvements
  - **48% performance improvement** in route lookup operations through
  optimized algorithms and caching
  - Enhanced object lookups and streamlined matching logic for better V8
  performance

  ## 🏗️ New Modular Architecture
  - **Pluggable router system** with `RouterInterface` - swap router
  implementations at runtime
  - **New @feathersjs/routing package** with Express and Koa-compatible routers
  - Clean separation between core routing logic and framework-specific
  implementations

  ## ✨ Enhanced Routing Features
  - **Catch-all routes** with `::param` syntax for handling wildcards (e.g.,
  `/docs/::path`)
  - Improved route matching with better placeholder and parameter handling
  - Full backward compatibility with existing v5 routing patterns

  ## 🔧 Code Quality
  - DRY refactoring eliminates code duplication between Express/Koa
  implementations
  - Comprehensive test coverage for all new features and performance
  optimizations
  - Type-safe router swapping with full TypeScript support

  This enables both better performance for existing apps and flexible routing
  architectures for complex use cases.


packages/feathers/src/router.ts
```diff
@@ -1,25 +1,51 @@
...
+   */
+  caseSensitive: boolean
+}
+
 export class RouteNode<T = any> {
   data?: T
-  children: { [key: string]: RouteNode } = {}
+  children: { [key: string]: RouteNode } = Object.create(null) // Optimize object lookup
   placeholders: RouteNode[] = []
+  catchAll?: RouteNode<T>
```
daffl, line 38: I'm not sure we need a catchAll?

packages/feathers/src/router.ts
```diff
@@ -63,71 +116,141 @@ export class RouteNode<T = any> {
...
-    const result = stripSlashes(path).split('/')
+    // Check cache first
+    if (this.pathCache[path]) {
+      return this.pathCache[path]
+    }
+
+    // Fast path for root
+    if (!path || path === '/') {
+      const result = ['']
+      this.pathCache[path] = result
```
daffl, line 223: `pathCache` is never cleaned up so anybody can crash the server by making a lot of unique URL requests.

marshallswain: Good catch.  I added a simple LRU mechanism that defaults to 1000 cached paths before booting old ones.

packages/feathers/src/router.ts
marshallswain: This one ironically came up twice in the last couple days.  While I was making an npm server with Feathers over the weekend, I found a route that legitimately needed to capture a dynamic number of params.  I was able to work around it, but the dynamic params would have made it cleaner.  Then today Dallin had another use case that made sense to me. A root-level route that handles whatever other paths you throw at it:

```
/actions
/actions/test
/actions/test/test1/tes2
```

And with the catchAll feature, you can put a catch all custom param at the end of the route, so you could do this:

```ts
app.use('actions/::args', ..., { routeParams: ['args'] })

Then you get an array of args to handle as you please.
```

It's the most likely edge case to handle, so we cover 95% of routing use cases without any sacrifice of speed.

Then I realized that we had an opportunity to improve the upgrade path because there are likely users who are stuck with their old Express and Koa routes who might like to upgrade.  You already kept the router decoupled, so I made an express router and koa router in a routing package.  That might help people upgrade, or motivate them to do so if they feel like the upgrade path is easier.

PR #3565, opened 2025-02-14
feat(typebox): Update to latest TypeBox

daffl:

Issue #3182, opened 2023-04-28
Query syntax problems with TypeBox 0.26


daffl:
The query syntax helper creates two different types between TypeBox 0.25:

<img width="312" alt="Screenshot_2023-04-20_at_20 46 18" src="https://user-images.githubusercontent.com/338316/235203064-f8b55bac-626d-4c86-b17e-f42e66617ac5.png">

And 0.26:

<img width="275" alt="Screenshot_2023-04-20_at_20 51 28" src="https://user-images.githubusercontent.com/338316/235203068-471d1578-4655-4782-be9e-089ef830386f.png">

While the local package tests are passing, in a generated app this means that any operator like

```ts
app.service('messages').find({
  query: {
    createdAt: {
      $gte: Date.now() - 24 * 60 * 60 * 1000
    }
  }
})
```

Will receive a TypeScript error like this:

```
Property '$something' is incompatible with index signature.
          Type '<type>' is not assignable to type 'never'.
```

The current solution is to revert to TypeBox 0.25 and probably add a better integration like

```ts
Type.QuerySyntax(queryProperties, ['text', 'id'])
```

To v5.1

daedalus28:
Is there any way to help move this along? The typebox dependency is very stale - nearly two years out of date.

I'm happy to attempt a PR for this, but I'm not sure I completely follow the issue being described here. Can you provide a unit test that passes on 0.25 but fails on the latest version?

Alternatively, is there a way we can use the latest typebox directly in our apps while feathers itself still relies on the older version?

daedalus28:
I also see https://github.com/feathersjs/feathers/pull/3532 was opened 5 months ago with no engagement on it.

@daffl Any thoughts?

daffl:
The update is probably still not possible without introducing a breaking change (as the breaking tests in #3532 showed before the logs expired).

I recommend installing your own Typebox dependency in the latest version and using that to explicitly define the schemas you need including the query syntax you actually use.

daedalus28:
So if I understand right, the recommended path is for users to install typebox themselves to construct schemas. Does that mean we should avoid using `getValidator` and `querySyntax` from `@feathersjs/typebox` as well? Or are they safe to mix?

Assuming this is true, it would be great to update the docs to reflect it.

daffl:
`getValidator` will work but `querySyntax` is where the compatibility issue is happening.

[...]

Issue #2985, opened 2023-01-12
Type instantiation is excessively deep and possibly infinite


digitalcortex:
### Steps to reproduce

Github repository: https://github.com/digitalcortex/feathers-dove-typebox-bug

### Expected behavior

No Typescript errors, properly merged types.

### Actual behavior

Typescript error `Type instantiation is excessively deep and possibly infinite.ts(2589)`

### System configuration

Please check package.json in the repo above.

**NodeJS version**: v18.11.0

**TypeScript version**: ^4.9.3

**Operating System**: Mac OS

howlettt:
Looks like the error is caused by having `moneyTransferServices` on both `userDataSchema` and `userSchema`. I was able to fix the error by changing the following line in `userDataSchema`
```
Type.Partial(userSchema),
```
to
```
Type.Omit(Type.Partial(userSchema), ["moneyTransferServices"]),
```

Issue #2927, opened 2022-12-14
bug(typebox): Typebox package should export the Value APIs


lvivier:
### Steps to reproduce

The `@feathersjs/typebox` package doesn't export the [super useful values module](https://github.com/sinclairzx81/typebox#values). 

Happy to submit a PR for this.

### Expected behavior

```ts
import { Value } from '@feathersjs/typebox/value';

Value.Cast(DataSchema, ctx.data); // for example
```

### Actual behavior

"Cannot find module"

### System configuration

Tell us about the applicable parts of your setup.

**Module versions** (especially the part that's not working):
- @feathersjs/feathers@5.0.0-pre.33
- - @feathersjs/typebox@5.0.0-pre.33

**NodeJS version**: 16.13.2

**Operating System**: macOS

**Browser Version**:

**React Native Version**:

**Module Loader**:

daffl:
This is strangely turning out to be more complicated than I thought. For some reason just declaring the `exports` in our `package.json` like it is done in TypeBox itself does not work (this is also part of the problem of ESM builds in #2665).

Since the dependency gets installed automatically anyway, I recommend loading it from `@sinclair/typebox` direclty.

FossPrime:
Another workaround would be to load from ./node_modules/ directly. You can go as far as to use subpath exports to make it easier to do this in several files.

@daffl  If a package inside node_modules that doesn't support ESM loads a package, any package, it will always pick the CommonJS version, even if the user-land package.json is set to type=module.

You could have also ran into an issue with conditional exports and how you have to the more specific conditions like "node", "deno" and "development" above vague ones like "default" or "browser".

I'm happy to help get these issues expedited... I was hoping to have a CJS problem free year... i've been dealing with this since 2019 as you know.

PR #3612, opened 2025-09-05
feat: Upgrade to latest TypeBox

daffl:
This will be a breaking change because of the changes in TypeBox `Type.Intersect` and `Type.Composite` and the removal of `$in` and `$nin` support for querying arrays in MongoDb which broke the TypeScript type inference mechanism.

- Related https://github.com/feathersjs/feathers/pull/3565
- Related https://github.com/feathersjs/feathers/issues/2985
- Related https://github.com/feathersjs/feathers/issues/3182
- Related https://github.com/feathersjs/feathers/issues/2927


daedalus28:
@daffl FYI - typebox 1.0 was just released. Notably, `Type.Composite` was removed (which this PR just moved to). Now that it's 1.0, I expect future releases will be less likely to include breaking changes. There are also some specific changes feathers might be interested to support different library and user versions of typebox (by moving off of symbols and onto `~` prefixed properties)

daffl:
I thought there wouldn't be yet another change around this in the final version. Follow up PR is in https://github.com/feathersjs/feathers/pull/3616 but even just trying to compile the package completely crashes the TypeScript compiler for me.

daedalus28:
That's unfortunate. Hopefully typebox's API is stabilized now. It would be really great to kick off feathers v6 with 1 of typebox, but I definitely feel your pain (and appreciate all the effort you've put into feathers over the years!).

PR #3616, opened 2025-09-24
feat: Update to Typebox v1

daffl:
Follow-up for https://github.com/feathersjs/feathers/pull/3612

### DIFF
--- a/packages/typebox/test/index.test.ts
+++ b/packages/typebox/test/index.test.ts
@@ -81,7 +81,7 @@ describe('@feathersjs/schema/typebox', () => {
   })
 
   it('defaultAppConfiguration', async () => {
-    const configSchema = Type.Composite([
+    const configSchema = Type.Intersect([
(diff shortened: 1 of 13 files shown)


packages/typebox/test/index.test.ts
```diff
@@ -81,7 +81,7 @@ describe('@feathersjs/schema/typebox', () => {
   })
 
   it('defaultAppConfiguration', async () => {
-    const configSchema = Type.Composite([
+    const configSchema = Type.Intersect([
```
daedalus28: I think the preferred version in v1 is `Interface` and `Evaluate`

https://github.com/sinclairzx81/typebox/blob/main/changelog/1.0.0-migration.md#typecomposite

daffl: I think I got it working with help from @sinclairzx81 in https://github.com/sinclairzx81/typebox/issues/1361

PR #3622, opened 2025-10-20
fix(generators): typebox generated schema resolver generic

AshotN:
### Summary

The JSON schema generator had the correct types for resolvers. TypeBox did not.

This led to `obj` in resolvers being typed incorrectly when using TypeBox.


daffl (approved):

daffl:
Good catch, thank you!

mvarendorff:
Is the first line of these really correct? My understanding was that `nameResolver` would not be related to the query necessarily. I am relatively new to the whole structure but it seems to be similar in the JSON schema generator:

https://github.com/feathersjs/feathers/blob/1d38e4a21e9ed5da31d524e5015f039b75668ec0/packages/generators/src/service/templates/schema.json.tpl.ts#L59

Which also does not use the `Query` type in that position. Additionally, the hook generated for that schema is `resolveResult` and not `resolveQuery`.

More than happy to open a PR if my understanding is correct here, I just wanted to make sure since I just started looking into Feathers the other day.

PR #3633, opened 2025-12-18
fix: Prioritize URL hash access tokens over stored tokens and clear s…

### LINKED ISSUE #3631: Authenticate twice using an access token in the URL might not work as expected
The authentication client is able to read an access token from the URL, this is typically used when login through OAuth2. If a token is not found in the local storage but is found in the URL then it is "consumed" (ie removed from it) and stored as usual in the local storage. 

The problem is that if you try to log again by using a new token in the URL Feathers will not use it as the previous one is found in the local storage. It seems to me that if you provide a new token in the URL you expect Feathers to consume it instead of using the previous one. So I wonder if it might not be more logical to check first if a token is available in the URL before checking the local storage as done by the authentication `getAccessToken` method by default.

Typically we are using this capability to create kind of "magic" links to an application where a user can open a specific page without requiring to login first because the token is part of the URL. However, the token contains a permission to underlying services/objects so that the user can only open this specific page with it and not others. It was not working fine when the user tried to open in the same browser two magic links due to the problem discussed previously, we had to update the authentication `getAccessToken` method like this to make it work better:
```
getAccessToken = async function () {
    let accessToken = await this.getFromLocation(window.location)
    if (!accessToken) {
      accessToken = await this.storage.getItem(this.options.storageKey)
    } else {
      // Ensure any previous token will not be used
      await this.removeAccessToken()
    }
    return accessToken || null
  }
```

Let me know what you think.

faizi-7:
Fix: Prioritize URL access tokens over storage when using magic links
Summary
This PR fixes an issue where getAccessToken prioritizes localStorage tokens over URL hash tokens, causing magic links to fail when users already have a stored token.

Problem: When implementing passwordless authentication via magic links, the getAccessToken method checks localStorage before examining the URL hash. This means users with a stale token in storage cannot authenticate via a fresh magic link - the old token is used instead of the new one from the URL.

Solution: Modified getAccessToken to:

Check the URL hash for an access token before checking storage
Return the URL token even if a different token exists in storage
Clear (remove) the stored token when a URL token is consumed
Related Issues
Fixes #3631
Other Information
This change aligns with the expected behavior described in the original issue - if a user provides a new token in the URL, Feathers should consume it instead of using the previous one from localStorage.

Use case: Magic links to applications where a user can open a specific page without requiring login first because the token is part of the URL. Previously, opening two different magic links in the same browser would fail because the first token was cached.

Changes:

packages/authentication-client/src/core.ts
 - Modified getAccessToken to prioritize URL tokens
Added test case to verify URL token priority behavior

### DIFF
--- a/packages/authentication-client/src/core.ts
+++ b/packages/authentication-client/src/core.ts
@@ -114,16 +114,24 @@ export class AuthenticationClient {
   }
 
   /**
-   * Returns the access token from storage or the window location hash.
+   * Returns the access token from the window location hash or storage.
    *
-   * @returns The access token from storage or location hash
+   * @returns The access token from location hash or storage
    */
   getAccessToken(): Promise<string | null> {
-    return this.storage.getItem(this.options.storageKey).then((accessToken: string) => {
-      if (!accessToken && typeof window !== 'undefined' && window.location) {
-        return this.getFromLocation(window.location)
-      }
+    if (typeof window !== 'undefined' && window.location) {
+      return this.getFromLocation(window.location).then((urlToken) => {
+        if (urlToken) {
+          return this.removeAccessToken().then(() => urlToken)
+        }
 
+        return this.storage.getItem(this.options.storageKey).then((storageToken: string) => {
+          return storageToken || null
+        })
+      })
+    }
+
+    return this.storage.getItem(this.options.storageKey).then((accessToken: string) => {
       return accessToken || null
     })
   }
--- a/packages/authentication-client/test/index.test.ts
+++ b/packages/authentication-client/test/index.test.ts
@@ -96,6 +96,32 @@ describe('@feathersjs/authentication-client', () => {
     }
   })
 
+  it('getAccessToken prefers URL token over storage token (#3631)', async () => {
+    const auth = app.authentication
+    const storageToken = 'old-storage-token'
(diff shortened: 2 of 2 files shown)


daffl:
Thank you for the pull request. I remember looking into this at some point but I can't find the related discussion. The issue with this precedence is that it might be possible for an attacker to get you to open a URL with one of their access tokens. I think if implementing this way, it should probably at least check if the subject is the same.

e-kibet:
Putting access tokens in the URL is generally discouraged. URLs can leak via: Browser history, Server logs and Analytics tools. For example, If getFromLocation() reads from window.location, you’re already in a risky zone.

daffl:
This is correct, however, location hashes usually won't show up in server logs or analytics since they are not sent with browser requests (which is why it was implemented that way here). Still not ideal but there isn't too many other options for doing oAuth in this case.

We do have an entirely new approach for user authentication for any platform or framework that is using the web crypto API with a much more secure, scalable and flexible access mechanisms that we are getting ready for people to test hopefully very soon.

As for this PR, I'm not sure which way is the better approach. @faizi-7 would it be acceptable for you to [customize the client](https://feathersjs.com/api/authentication/client.html#customization) with this functionality?

PR #3638, opened 2026-01-10
feat: Allow hooks on all methods and add externalMethods option (#3457)

### LINKED ISSUE #3457: Hooks should run on non-enabled methods
### Steps to reproduce

[Discord discussion](https://discord.com/channels/509848480760725514/1212264370470527017)

https://codesandbox.io/p/devbox/feathers-custom-service-hooks-8m28r3?file=%2Fsrc%2Fapp.ts%3A27%2C22

export const customMethods = ['enabledMethod'] as const
...
      //BEFORE ALL is called with this
      app.service('custom').enabledMethod()
      //But not this
      app.service('custom').notEnabledMethod()

### Expected behavior

Ideally hooks should run on all methods

### Actual behavior

Hooks are only ran on enabled methods

marshallswain:
## Summary

This PR implements the feature requested in #3457, allowing hooks to run on all service methods while providing fine-grained control over which methods are exposed externally.

#### Type Declarations (`declarations.ts`)
- Updated `methods` description to clarify it controls which methods `all` hooks apply to
- Added new `externalMethods` option to control HTTP transport exposure

#### Service (`service.ts`)
- Added `getExternalMethods()` function (defaults to `methods` for backwards compatibility)
- Updated `normalizeServiceOptions()` to include `externalMethods`

#### Hooks (`hooks.ts`)
- Added `allMethods` to hook store to track which methods receive `all` hooks
- Modified `collectHooks()` to only include `all` hooks for methods in `allMethods`
- Updated `hookMixin()` to allow lazy creation of hook managers for any service method

#### HTTP Transport (`http/index.ts`)
- Updated to use `getExternalMethods()` for HTTP access control

#### Tests
- Added tests for lazy hook registration on unconfigured methods
- Added tests for `all` hooks scoping
- Added tests for `externalMethods` HTTP behavior

### 1. Individual method hooks work without configuration

```typescript
// Can register hooks on ANY method, even if not in `methods`
service.hooks({
  helperMethod: [myHook]  // works!
})
```

`methods` controls which methods receive `all` hooks

```typescript
app.use('myService', new MyService(), {
  methods: ['find', 'get', 'create']
})

service.hooks({
  around: {
    all: [loggingHook]  // only runs on find, get, create
  }
})
```

### 3. `externalMethods` controls HTTP exposure

```typescript
app.use('myService', new MyService(), {
  methods: ['find', 'get', 'create', 'internalAction'],  // all get hooks
  externalMethods: ['find', 'get', 'create']  // only these exposed via HTTP
})
```

## Backwards Compatibility

- If `externalMethods` is not specified, it defaults to `methods`
- Existing code works without modification
- All 354 existing tests pass

## Related

- Fixes #3457
- Companion PR for dove (v5.x): #3637

### DIFF
--- a/packages/feathers/src/declarations.ts
+++ b/packages/feathers/src/declarations.ts
@@ -30,9 +30,16 @@ export interface ServiceOptions<MethodTypes = string> {
    */
   events?: string[] | readonly string[]
   /**
-   * A list of service methods that should be available __externally__ to clients
+   * A list of service methods that `all` hooks will apply to.
+   * Defaults to the standard service methods (find, get, create, update, patch, remove)
+   * that exist on the service.
    */
   methods?: MethodTypes[] | readonly MethodTypes[]
+  /**
+   * A list of service methods that should be available __externally__ to clients
+   * via transports like HTTP or WebSockets. Defaults to `methods` if not specified.
+   */
+  externalMethods?: MethodTypes[] | readonly MethodTypes[]
   /**
    * Provide a full list of events that this service should emit to clients.
    * Unlike the `events` option, this will not be merged with the default events.
--- a/packages/feathers/src/hooks.ts
+++ b/packages/feathers/src/hooks.ts
@@ -28,6 +28,11 @@ type HookStore = {
   error: { [method: string]: HookFunction[] }
   collected: { [method: string]: AroundHookFunction[] }
   collectedAll: { before?: AroundHookFunction[]; after?: AroundHookFunction[] }
+  /**
+   * Methods that should receive "all" hooks. Methods not in this set
+   * will only receive their specific hooks, not "all" hooks.
+   */
+  allMethods?: Set<string>
 }
 
 type HookEnabled = { __hooks: HookStore }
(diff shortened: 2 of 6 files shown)


daffl:
I'm wondering if we need to add a new API if we are promoting using decorators going forward. For internal methods you need to list the parameter names of the method anyway to make it useful (to have things like `context.message`).

This already works and lets you register hooks on any method:

```ts
import { hooks } from 'feathers'

class MyService {
  @hooks([]).params('message')
  internalMethod(message: string) {}
  
  @hooks([])
  find() {
  }
}
```

marshallswain:
There are still use cases where decorators will not work well, like when using built-in services from a database manager instance.  It's a bit verbose to do custom adapter classes just to be able to add hooks.

marshallswain:
I forgot about the .params() calls

daffl:
Wouldn't that work with [the object wrapper](https://github.com/feathersjs/hooks?tab=readme-ov-file#global-hooks-on-an-object)?

```ts
import { hooks } from 'feathers'
import { MemoryService } from 'feathers-memory'

const messageService = hooks(new MemoryService(), {
  myMethod: middleware([]).params('message')
})

app.use('messages', messageService)
```

PR #3642, opened 2026-01-28
feat: Enhanced custom methods with @method decorator

### LINKED ISSUE #1976: Custom methods the Feathers way

marshallswain:
This PR adds enhanced custom method support for Feathers v6, allowing custom methods to be first-class citizens with flexible signatures, proper HTTP verb mapping, and clean URLs.

## Features

- **`@method` decorator** for configuring custom service methods
- **Different argument signatures** (not just `data, params`)
- **Different HTTP verbs** (GET, POST, PUT, PATCH, DELETE)
- **Clean URL paths** (e.g., `/messages/:id/status` instead of `X-Service-Method` header)
- **Internal-only methods** with `external: false`
- **`clientMethods()`** helper for client configuration

## Example

```ts
import { feathers, method } from '@feathersjs/feathers'

class MessageService {
  @method({ args: ['id', 'params'], http: 'GET', path: ':id/status' })
  async status(id: Id, params: Params) {
    return { id, status: 'active' }
  }

  @method({ args: ['id', 'params'], http: 'POST', path: ':id/archive' })
  async archive(id: Id, params: Params) {
    return this.patch(id, { archived: true }, params)
  }

  @method({ args: ['data', 'params'], external: false })
  async internalProcess(data: any, params: Params) {
    return { processed: data }
  }
}

app.use('messages', new MessageService())
// GET /messages/123/status -> status('123', params)
// POST /messages/123/archive -> archive('123', params)
```

## Documentation

- Updated services documentation with full `@method` decorator guide
- Updated REST client documentation with custom method paths
- Added plan document at `docs/v6-custom-methods-plan.md`

## Tests

All 405 tests pass.

## Related

- Closes #1976
- Replaces #3638
- Similar #3339


marshallswain:
Update: Improved client configuration API**

Changed how method configuration is passed to the client to avoid bundling server-side code.

**Before (problematic):**
```ts
// This required importing service CLASSES, bundling server code into the client
import { buildMethodConfig, type InferServiceTypes } from '@feathersjs/feathers'
import { MessageService } from './services/messages.service'

const services = { messages: MessageService }
export const serviceMethods = buildMethodConfig(services)  // instantiated classes internally
export type ServiceTypes = InferServiceTypes<typeof services>
```

**After (secure):**
```ts
// Only imports the static methods CONFIG object, not the class implementation
import { clientMethods } from '@feathersjs/feathers'
import { MessageService } from './services/messages.service'

export const serviceMethods = clientMethods({
  messages: MessageService.methods  // just the static property
})

export type ServiceTypes = {
  messages: MessageService  // types defined separately
}
```

**What `clientMethods()` does:**
- Filters out `external: false` methods (internal-only, shouldn't be exposed)
- Filters out methods without `path` (standard CRUD doesn't need config)
- Strips server-only properties (`external`, `event`) — only keeps `args`, `http`, `path`

**Key insight:** Types come from the service class definition itself. The runtime config (`clientMethods()`) only tells the HTTP client which verb/path to use — it doesn't affect TypeScript types at all.

PR #3651, opened 2026-01-31
feat: Chainable @hooks().params() decorator with hookMixin support

marshallswain:
## Summary

Adds a chainable `@hooks().params()` decorator syntax that allows defining custom parameter names for hook context on service methods. When registered as a Feathers service, `hookMixin` respects these custom params instead of overriding them with defaults.

## Usage

```ts
class NotificationService {
  @(hooks([
    async (ctx, next) => {
      console.log(ctx.userId)   // Custom param
      console.log(ctx.message)  // Custom param  
      console.log(ctx.app)      // Feathers context still available
      await next()
    }
  ]).params('userId', 'message'))
  async notify(userId: string, message: string) {
    return { sent: true }
  }
}

const app = feathers().use('notifications', new NotificationService(), {
  methods: ['notify']
})

await app.service('notifications').notify('user123', 'Hello!')
```

## Chainable Methods

- `.params(...names)` - Define context parameter names
- `.props(properties)` - Add static properties to context
- `.defaults(initializer)` - Provide default values via function

All can be chained:
```ts
@(hooks([])
  .params('id', 'data')
  .props({ serviceName: 'messages' })
  .defaults(() => ({ timestamp: Date.now() })))
```

## Changes

- **`packages/feathers/src/hooks/base.ts`** - Added `clone()` method to `HookManager`
- **`packages/feathers/src/hooks/hooks.ts`** - Added `ChainableHookDecorator` interface and `createChainableDecorator()`
- **`packages/feathers/src/hooks.ts`** - Modified `hookMixin()` to detect and respect custom params from decorators
- **`packages/feathers/src/hooks/decorator.test.ts`** - Added 11 new tests
- **`website/content/api/hooks-chainable-params.md`** - Documentation (pending integration into docs rewrite)

Adds approximately 110 lines of code to the core (not counting comments).

## Why

Previously, custom service methods were limited to the default hook context params (`data`, `params`, `id`). If you had a method like `notify(userId, message)`, hooks would receive `ctx.data` and `ctx.params` instead of `ctx.userId` and `ctx.message`.

This change allows:
- Custom methods to have meaningful, descriptive context properties
- Hooks to work naturally with any method signature
- Better developer experience when building services with non-CRUD methods

Based on [this comment](https://github.com/feathersjs/feathers/pull/3638#issuecomment-2572569338).

### DIFF
--- a/packages/feathers/src/hooks.ts
+++ b/packages/feathers/src/hooks.ts
@@ -148,6 +148,51 @@ export function createContext(service: Service, method: string, data: HookContex
   return createContext(data) as HookContext
 }
 
+/**
+ * Creates property descriptors for Feathers-specific context properties.
+ * Used to add app, path, service, method, etc. to hook contexts.
+ */
+function createFeathersContextProps<A>(
+  app: A,
+  path: string,
+  method: string,
+  service: FeathersService<A>
+): PropertyDescriptorMap {
+  return {
+    app: { value: app, enumerable: true, writable: true },
+    path: { value: path, enumerable: true, writable: true },
+    method: { value: method, enumerable: true, writable: true },
+    service: { value: service, enumerable: true, writable: true },
+    event: { value: null, enumerable: true, writable: true },
+    type: { value: 'around', enumerable: true, writable: true },
+    statusCode: {
+      enumerable: true,
+      get(this: HookContext) {
+        return this.http?.status
+      },
+      set(this: HookContext, value: number) {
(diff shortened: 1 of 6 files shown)


marshallswain:
Neither `.props()` nor `.defaults()` solve a problem that users actually have when writing services.  I think there's a different solution to this.  Needs more deliberation.

daffl:
Maybe we can discuss this more in a new issue.

PR #3665, opened 2026-03-04
fix(mongodb): Block arbitrary update operators in _patch data (CWE-943)

marshallswain:
## Summary

Addresses NoSQL operator injection in `_patch()` where arbitrary MongoDB update operators pass through directly to `findOneAndUpdate()`.

## Approach

Adds a `disabledOperators` option (defaults to `['$rename']`) that drops specified MongoDB update operators from patch data.

`$rename` is blocked by default since it can expose internal fields. Other operators like `$push` and `$inc` remain allowed — the dangerous scenarios they enable (e.g. unsetting passwords, pushing roles) should be caught by schema validation in a properly configured app.

### Customizing the denylist

Per-service:
```ts
new MongoDBService({
  Model: db.collection('users'),
  disabledOperators: ['$rename', '$unset', '$inc']
})
```

Per-call:
```ts
service.patch(id, data, {
  adapter: { disabledOperators: ['$rename', '$inc'] }
})
```

## Related

- #3664 — fixes a separate NoSQL injection vector via object IDs


daffl:
Wasn't it always intended for the operators to be supported?

daffl (approved):

dorakemon:
Thank you so much for fixing

PR #3653, opened 2026-02-02
fix(oauth): Patch open redirect and origin validation

daffl:
This PR fixes three security issues in @feathersjs/authentication-oauth:
                                                                                                                         
  - Open redirect via URL authority injection - The redirect query parameter was concatenated to the base origin without 
  validation, allowing attackers to steal access tokens via URLs like https://target.com@attacker.com#access_token=secret
  - Origin validation bypass - The `startsWith()` check allowed attackers to bypass origin validation by registering       
  domains like target.com.attacker.com when origin configuration does not end with `/`
  - Internal headers exposure - All HTTP request headers were stored in the session cookie (which is signed but not      
  encrypted), could leak proxy headers like X-Forwarded-For or internal API keys

🤖 Co-created with https://claude.com/claude-code

PR #3676, opened 2026-03-19
fix(authentication-oauth): Use actual URL origin comparison for origin check

daffl:
This is a follow-up to https://github.com/feathersjs/feathers/pull/3653 and https://github.com/feathersjs/feathers/pull/3669 and uses the proper `URL` constructor to check the redirects instead of an arbitrary regex. Still passes all tests for existing vulnerabilities.

### DIFF
--- a/packages/authentication-oauth/src/strategy.ts
+++ b/packages/authentication-oauth/src/strategy.ts
@@ -144,11 +144,11 @@ export class OAuthStrategy extends AuthenticationBaseStrategy {
       return null
     }
 
-    const redirectUrl = `${redirect}${queryRedirect}`
-
     if (queryRedirect) {
       validateRedirectOrigin(redirect, queryRedirect)
     }
+
+    const redirectUrl = `${redirect}${queryRedirect}`
     const separator = redirectUrl.endsWith('?') ? '' : redirect.indexOf('#') !== -1 ? '?' : '#'
     const authResult: AuthenticationResult = data
     const query = authResult.accessToken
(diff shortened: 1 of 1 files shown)


packages/authentication-oauth/src/strategy.ts
```diff
@@ -105,18 +144,11 @@ export class OAuthStrategy extends AuthenticationBaseStrategy {
...
-    // - @attacker.com -> https://target.com@attacker.com (authority injection)
-    // - .attacker.com -> https://target.com.attacker.com (domain suffix attack)
-    // - -attacker.com -> https://target.com-attacker.com (domain suffix attack)
-    // - //attacker.com -> protocol-relative redirect
-    // - \attacker.com -> backslash redirect
-    if (queryRedirect && (!/^\//.test(queryRedirect) || /[@\\]|\/\//.test(queryRedirect))) {
-      throw new NotAuthenticated('Invalid redirect path.')
-    }
-
     const redirectUrl = `${redirect}${queryRedirect}`
```
marshallswain: I'd probably swap the order so we validate it then we use it:

```ts
    if (queryRedirect) {
      validateRedirectOrigin(redirect, queryRedirect)
    }

    const redirectUrl = `${redirect}${queryRedirect}`
```

marshallswain (approved):
One linty nit.  :shipit:

skinofstars:
Hi, I think this has caused an issue for us. We use a port version for local development, i.e. `http://localhost:3003`. This no longer works.

PR #3691, opened 2026-06-04
fix(mongodb): let objectid keyword fail validation

### LINKED ISSUE #3685: `keywordObjectId` throws instead of returning `false`, breaking `anyOf` / union validation
## Package

`@feathersjs/mongodb` — specifically `src/converters.ts`, the `keywordObjectId` export.

## Description

The `keywordObjectId` AJV keyword throws an `Error` when the value is not a valid ObjectId string:

```typescript
return function (value: string, obj: any) {
  const { parentData, parentDataProperty } = obj
  try {
    parentData[parentDataProperty] = new ObjectId(value)
    return true
  } catch (error) {
    throw new Error(`invalid objectid for property "${parentDataProperty}"`)
  }
}
```

This breaks AJV's `anyOf` / `oneOf` evaluation. When a schema uses a union type like:

```json
{
  "anyOf": [
    { "type": "string", "objectid": true },
    { "type": "null" }
  ]
}
```

AJV is supposed to try each branch and succeed if any passes. But because the keyword **throws** instead of returning `false`, AJV cannot gracefully fall through to the `{ "type": "null" }` branch — the exception propagates up the call stack.

## Steps to reproduce

1. Register `keywordObjectId` on an AJV instance with `coerceTypes: true` and `useDefaults: true`
2. Define a schema with a nullable objectid field:
   ```typescript
   const schema = Type.Object({
     refId: Type.Union([Type.String({ objectid: true }), Type.Null()], { default: null }),
   })
   ```
3. Validate an object where `refId` is missing (so the default `null` is applied)
4. **Expected:** Validation passes — `null` matches the `Type.Null()` branch
5. **Actual:** Throws `Error: invalid objectid for property "refId"`

## Root cause

With `coerceTypes: true`, AJV attempts to coerce `null` into a string when evaluating the first `anyOf` branch. The coerced value (empty string) reaches the `objectid` keyword, which calls `new ObjectId("")`, catches the error, and **re-throws** instead of returning `false`.

AJV keyword validators must return `false` (not throw) to signal failure, otherwise AJV cannot continue evaluating alternative branches.

## Suggested fix

```diff
 return function (value: string, obj: any) {
   const { parentData, parentDataProperty } = obj
   try {
     parentData[parentDataProperty] = new ObjectId(value)
     return true
   } catch (error) {
-    throw new Error(`invalid objectid for property "${parentDataProperty}"`)
+    return false
   }
 }
```

This allows AJV to produce a standard validation error (referencing the keyword) rather than an uncaught exception, and correctly supports `anyOf` / `oneOf` unions with nullable objectid fields.

puneetdixit200:
### Summary

`keywordObjectId` currently throws when a string cannot be converted to a MongoDB `ObjectId`. That prevents AJV from treating the keyword as a normal branch failure, so `anyOf` / `oneOf` schemas cannot continue to another valid branch such as `null`.

This changes the keyword validator to return `false` on invalid ObjectId conversion, which lets AJV report a standard validation failure and continue evaluating union branches.

Closes #3685.

No.

### Other Information

Added regression coverage for nullable union validation and updated the invalid ObjectId test to assert a normal AJV validation failure instead of an uncaught exception.

Verification:

```console
source ~/.nvm/nvm.sh && nvm use 18.20.8 >/dev/null && npm_config_cache=/tmp/feathers-107-npm-cache npm test --workspace @feathersjs/mongodb -- --grep "objectid keyword"
source ~/.nvm/nvm.sh && nvm use 18.20.8 >/dev/null && npm_config_cache=/tmp/feathers-107-npm-cache npm test --workspace @feathersjs/mongodb -- --grep "ObjectId resolvers|objectid keyword"
source ~/.nvm/nvm.sh && nvm use 18.20.8 >/dev/null && npm_config_cache=/tmp/feathers-107-npm-cache npm run compile --workspace @feathersjs/mongodb
source ~/.nvm/nvm.sh && nvm use 18.20.8 >/dev/null && npm_config_cache=/tmp/feathers-107-npm-cache npx prettier --check packages/mongodb/src/converters.ts packages/mongodb/test/converters.test.ts
source ~/.nvm/nvm.sh && nvm use 18.20.8 >/dev/null && npm_config_cache=/tmp/feathers-107-npm-cache npx eslint packages/mongodb/src/converters.ts packages/mongodb/test/converters.test.ts
git diff --check
```

### DIFF
--- a/packages/mongodb/src/converters.ts
+++ b/packages/mongodb/src/converters.ts
@@ -57,7 +57,7 @@ export const keywordObjectId = {
         parentData[parentDataProperty] = new ObjectId(value)
         return true
       } catch (error) {
-        throw new Error(`invalid objectid for property "${parentDataProperty}"`)
+        return false
       }
     }
   }
--- a/packages/mongodb/test/converters.test.ts
+++ b/packages/mongodb/test/converters.test.ts
@@ -89,7 +89,7 @@ describe('objectid keyword', () => {
     assert.equal(typeof data.otherId, 'string')
   })
 
-  it('fails on invalid objectids', async () => {
+  it('fails validation on invalid objectids', async () => {
     const schema = {
       type: 'object',
       properties: {
@@ -104,6 +104,28 @@ describe('objectid keyword', () => {
     }
     assert.equal(typeof data._id, 'string')
 
-    assert.throws(() => validate(data), /invalid objectid for property "_id"/)
+    assert.equal(validate(data), false)
+    assert.equal(validate.errors?.[0].keyword, 'objectid')
+  })
+
+  it('continues validating nullable unions when an objectid branch fails', async () => {
+    const nullableValidator = new Ajv({ coerceTypes: true, useDefaults: true })
+    nullableValidator.addKeyword(keywordObjectId)
+
+    const schema: any = {
+      type: 'object',
+      properties: {
+        refId: {
+          anyOf: [{ type: 'string', objectid: true }, { type: 'null' }],
+          default: null
+        }
+      },
+      additionalProperties: false
+    }
+    const validate = nullableValidator.compile(schema)
+    const data: { refId?: ObjectId | null } = {}
+
+    assert.equal(validate(data), true)
+    assert.equal(data.refId, null)
   })
 })


daffl (approved):

daffl:
Published in 5.0.46, thank you!

PR #3697, opened 2026-08-10
fix(typebox): support nested extension query keys

### LINKED ISSUE #3526: $sort from querySyntax function in @feathersjs/typebox does not consider the extensions param, so can't sort on extensions
### Steps to reproduce

(First please check that this issue is not already solved as [described
here](https://github.com/feathersjs/feathers/blob/dove/.github/contributing.md#report-a-bug) - if it is a general question or suggestion please start a [Discussion](https://github.com/feathersjs/feathers/discussions))

- [x] Tell us what broke. The more detailed the better.
I tried to add a nested sort field to the extensions for query synatx and it wouldn't work, I dug into the source code and saw that extensions are not considered, I'm not sure if this is expected behaviour or not?
image

- [x] If you can, please create a simple example that reproduces the issue and link to a gist, jrepo, etc. This makes it much easier for us to debug and issues that have a reproducible example will get higher priority.

I managed to solve it for my case by passing in an intersect type as the main argument:
image

### Expected behavior

Tell us what should happen

Ideally, it would be great if the properties passed in as extensions could be used on the `$sort` as well as the `$select`

### Actual behavior

These properties are currently ignored
Tell us what happens instead

### System configuration

`"@feathersjs/typebox": "^5.0.29"`

**NodeJS version**:
v20.14.0

**Operating System**:
MacOS

**Browser Version**:
N/A

**React Native Version**:
N/A

**Module Loader**:
npm

tonycoder-hub:
### Summary

`querySyntax` currently accepts nested extension fields for property queries, but `$sort` and `$select` only expose top-level schema keys. As a result, valid nested extension keys such as `profile.name` are rejected. This change derives sortable/selectable nested keys from non-operator extension entries while continuing to exclude `$`-prefixed query operators.

Fixes #3526.

No.

### Verification

- `npm run mocha --workspace @feathersjs/typebox` (10 passing)
- `tsc -p packages/typebox/tsconfig.json --noEmit`
- `git diff --check`

### Other Information

The implementation and review used AI assistance. The final diff, regression behavior, type checks, and test results were independently verified before submission.

### DIFF
--- a/packages/typebox/src/index.ts
+++ b/packages/typebox/src/index.ts
@@ -54,9 +54,35 @@ export function StringEnum<T extends string[]>(allowedValues: [...T], options?: 
   return Type.Unsafe<T[number]>({ type: 'string', enum: allowedValues, ...options })
 }
 
-const arrayOfKeys = <T extends TObject>(type: T) => {
-  const keys = Object.keys(type.properties)
-  return Type.Unsafe<(keyof T['properties'])[]>({
+type QueryExtensions<T extends TObject> = {
+  [K in keyof T['properties']]?: { [key: string]: TSchema }
+}
+
+type NoQueryExtensions = Record<never, never>
+
+type ExtensionPropertyKeys<X> = {
+  [K in Extract<keyof X, string>]: {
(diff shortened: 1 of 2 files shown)


tonycoder-hub:
Closing as stale — opened on or before 2026-08-17 and still unmerged.

PR #3700, opened 2026-08-13
fix(adapter-commons): validate query operators nested in arrays

marshallswain:
### Summary

- [x] `validateQueryProperty()` in `@feathersjs/adapter-commons` did not recurse into array values. `_.isObject()` in `@feathersjs/commons` excludes arrays, so `isPlainObject()` returned false and the value was returned unvalidated.
- [x] An extra array level around a sub-query bypassed the operator allow-list (`$or: [[{ $where: '1==1' }]]` was accepted; `$or: [{ $where: '1==1' }]` was correctly rejected). This is a v5 regression: v4 `cleanQuery` mapped over arrays.
- [x] The same allow-list also skipped **filter values**: non-array `$or`/`$and`, object `$select` (including `{ x: { $function: … } }` for MongoDB `$project`), and object `$sort` values.
- [x] `ObjectIdSchema` accepted any object (`additionalProperties: true`), so generated MongoDB query schemas treated `{ _id: { $where } }`, `{ $regex }`, `{ $ne: null }` as valid ids. Default `validateQuery` then skipped adapter sanitization.

`filterQuery()` is used by `AdapterBase.sanitizeQuery()`, so every adapter built on it is covered. `ObjectIdSchema` is tightened in both `@feathersjs/schema` and `@feathersjs/typebox`; the `objectid` keyword now accepts ObjectId instances and rejects other objects.

Not dependent on other PRs.

### Other Information

Reported privately by Khiêm Trần (k0m3g4) and Chien Huynh (chienhm). Affects `@feathersjs/adapter-commons` 5.0.0–5.0.48 (and `ObjectIdSchema` / `keywordObjectId` on generated Mongo apps). v4 is not affected by the array regression.

On `@feathersjs/memory` the array wrap allowed sift `$where` through sanitization. Please publish a patch release after merge.

### DIFF
--- a/docs/api/schema/typebox.md
+++ b/docs/api/schema/typebox.md
@@ -101,28 +101,61 @@ const messageQuerySchema = querySyntax(messageQueryProperties)
 type MessageQuery = Static<typeof messageQuerySchema>
 ```
 
-Additional special query properties [that are not already included in the query syntax](../databases/querying.md) like `$ilike` can be added like this:
+`querySyntax` already allows the [common operators](../databases/querying.md) on each property (`$gt`, `$gte`, `$lt`, `$lte`, `$ne`, `$in`, `$nin`, `$exists`). `$ne`, `$in`, and `$nin` also accept `null` (for example `{ userId: { $ne: null } }`). Anything else — SQL `LIKE`, Mongo `$regex`, geo, array operators — must be added per property. Only add operators your adapter actually supports.
 
 ```ts
-import { querySyntax } from '@feathersjs/typebox'
+import { querySyntax, Type, ObjectIdSchema } from '@feathersjs/typebox'
 
-// Schema for allowed query properties
 const messageQueryProperties = Type.Pick(messageSchema, ['id', 'text', 'createdAt', 'userId'], {
   additionalProperties: false
 })
+
 const messageQuerySchema = Type.Intersect(
   [
-    // This will additionally allow querying for `{ name: { $ilike: 'Dav%' } }`
     querySyntax(messageQueryProperties, {
+      text: {
+        $like: Type.String(),
+        $notlike: Type.String(),
+        $ilike: Type.String() // PostgreSQL
+      },
       name: {
-        $ilike: Type.String()
+        $regex: Type.String(),
+        $options: Type.String()
+      },
+      tags: {
+        $all: Type.Array(Type.String()),
+        $size: Type.Number(),
+        $elemMatch: Type.Object({}, { additionalProperties: true })
       }
     }),
-    // Add additional query properties here
-    Type.Object({})
+    Type.Object({}, { additionalProperties: false })
   ],
   { additionalProperties: false }
 )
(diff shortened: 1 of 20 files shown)


packages/mongodb/src/converters.ts
```diff
@@ -44,14 +44,24 @@ export async function resolveQueryObjectId(value: ObjectIdParam | IdQueryObject<
   return convertedObject
 }
 
+const isObjectId = (value: any): value is ObjectId =>
+  value != null && typeof value === 'object' && value._bsontype === 'ObjectId'
```
marshallswain: **[suggestion]** `isObjectId` treats any object with `_bsontype === 'ObjectId'` as a real BSON ObjectId. Client JSON can send `{ _bsontype: 'ObjectId' }` or `{ _bsontype: 'ObjectId', $where: '…' }`. With the keyword registered, that still passes `ObjectIdSchema` (`{ type: 'object', objectid: true }` has no other constraints). Docs claim operator documents are not valid ObjectIds. This does not restore sift `$where` RCE by itself: sift throws on mixed operator/property objects, and `{ $where: '…' }` alone fails `isObjectId`. It does mean the new “instance only” gate is spoofable and is not covered by the new tests.

**Suggestion:** Reject plain objects. Keep `instanceof ObjectId` for the same package copy, and for duplicate `mongodb`/`bson` copies require `_bsontype === 'ObjectId'` plus a non-`Object` constructor (or `toHexString`/`equals` functions). Add a test that `{ _bsontype: 'ObjectId', $where: '1==1' }` fails when the keyword is registered.

docs/api/schema/typebox.md
```diff
@@ -101,30 +101,40 @@ const messageQuerySchema = querySyntax(messageQueryProperties)
...
 const messageQuerySchema = Type.Intersect(
   [
-    // This will additionally allow querying for `{ name: { $ilike: 'Dav%' } }`
     querySyntax(messageQueryProperties, {
+      text: {
+        $like: Type.String(),
+        $notlike: Type.String(),
+        $ilike: Type.String() // PostgreSQL
+      },
       name: {
```
marshallswain: **[suggestion]** The querySyntax example picks `['id', 'text', 'createdAt', 'userId']` but attaches `$regex` / `$options` to `name`. `querySyntax` only applies extensions for keys that exist on the definition, so that `$regex` block is dropped. The following sentence claims `{ name: { $regex: 'feathers', $options: 'i' } }` is allowed; it is not, and TypeScript should reject the `name` extension key.

**Suggestion:** Put `$regex` / `$options` on `text` (or add `name` to the Pick) and update the “That allows …” sentence to match the properties that are actually in the schema.

marshallswain (commented):
Suggestions addressed in b06311d7: isObjectId now rejects spoofed `{ _bsontype: 'ObjectId' }` JSON, and the TypeBox example attaches `$regex`/`$options` to `text`.

PR #3701, opened 2026-08-27
fix(transport-commons): map HEAD requests like GET for REST services

### LINKED ISSUE #3464: Calls to feathers services with HEAD method are throwing an error since v5.x.
Hello,

### Steps to reproduce
Create a simple service :

info.service.js
```
import Info from './info.class';
import hooks from './info.hooks';

export default function (app) {

  app.use(
    '/v1/info',
    new Info({}, app)
  );

  const service = app.service('/v1/info');
  service.hooks(hooks);
}
```

info.class.js :
```
import { MethodNotAllowed } from '@feathersjs/errors';
import fse from 'fs-extra';

class Info {
  async setup(app) {
    this.app = app;
  }

  async find() {
    const { version } = await fse.readJson('./package-lock.json');
    return { version };
  }

  async get()
  {
    throw new MethodNotAllowed();
  }
}

export default Info;
```

- [ ] Since I recently updated feathers to v5.x, calls to a service with HEAD method are throwing an error (500).
`curl --location --head 'http://localhost:12030/api/v1/info'`

### Expected behavior

To return a 200 code.

### Actual behavior

Throwing a 500 error.

### System configuration

**Module versions** (especially the part that's not working): 5.0.21

**NodeJS version**: 20.11.1

**Operating System**: Ubuntu 22.04.3

**Module Loader**: commonJs

rome-xi:
Fixes #3464

Map HEAD like GET in getServiceMethod so REST HEAD probes do not throw MethodNotAllowed. Unit tests cover HEAD/head with and without an id.

### DIFF
--- a/packages/transport-commons/src/http.ts
+++ b/packages/transport-commons/src/http.ts
@@ -38,7 +38,9 @@ export function getServiceMethod(_httpMethod: string, id: unknown, headerOverrid
     return mappedMethod
   }
 
-  if (httpMethod === 'get') {
+  // HEAD is semantically a GET without a body (RFC 9110). Map it the same
+  // way so Express/Koa REST endpoints do not 405/500 on HEAD probes.
+  if (httpMethod === 'get' || httpMethod === 'head') {
     return id === null ? 'find' : 'get'
   }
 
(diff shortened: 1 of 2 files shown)


daffl (approved):
This makes sense, thank you!

PR #3705, opened 2026-09-14
fix(schema): recheck external dispatch after resolving

### LINKED ISSUE #3476: Cannot redefine property: Symbol(@feathersjs/schema/dispatch)
I've encountered an issue where `resolveExternal` will throw `Cannot redefine property: Symbol(@feathersjs/schema/dispatch)`. However, I can't reproduce this reliably. Sometimes the guard introduced with #3380 catches it, other times it passes the guard and the property is defined by `runResolvers`, then `setDispatch` attempts to redefine it. I don't know what influences the different behaviour, as I'll get either result, seemingly at random, rerunning the server without changes. The resolver is just as it was generated:

```ts
export const messageExternalResolver = resolve<Message, HookContext<MessageService>>({})
```

Along with the `around` hooks:
```ts
around: {
  all: [authenticate('jwt'), schemaHooks.resolveExternal(messageExternalResolver), schemaHooks.resolveResult(messageResolver)]
}
```

I think the guard might need to be moved to immediately before `setDispatch`, or perhaps a second guard could be added. 

### System configuration

**Module versions**:
@feathers/feathers:5.0.24, @feathers/schema:5.0.23

**NodeJS version**:
NodeJS v18.12.1 with tsx v4.7.1

**Operating System**:
Manjaro Linux

**Module Loader**:
ESM

KayanoLiam:
### Summary

`resolveExternal` can encounter dispatch data created by a nested service call or a concurrent resolver after its initial cache check. Defining the same non-configurable dispatch property again then throws `Cannot redefine property`.

Check the cache again immediately before each item-level and result-level write, reusing dispatch data already produced while awaiting resolution. This covers repeated result objects and concurrent calls sharing the same array or paginated container.

- [x] Are there any open issues that are related to this? Fixes #3476.
- [x] Is this PR dependent on PRs in other repos? No.

### Other Information

Added five service-level regressions, all failing with the reported TypeError on the original code. Tests also verify that filtered dispatch data is reused while the original service data remains intact.

`npm test --workspace=@feathersjs/schema` passes all 42 tests, including compilation and package checks. TypeScript, ESLint, Prettier, and diff checks pass for this change.


daffl (approved):

daffl:
This looks good, thank you for getting that in! It will go out with the next release!

PR #3706, opened 2026-09-15
fix(authentication): reject leftover codegen params for user (#3599)

### LINKED ISSUE #3412: Minor issue with authentication generator if service name is not "user"
Minor issue - running the feathers authentication generator appears to scaffold an authentication service which assumes the service is always called user. 

To reproduce the error generate authentication per the documenatation:

```> npx feathers generate authentication
? Which authentication methods do you want to use? Other methods and providers can be added at any time. Email + Password
? What is your authentication service name? admin
? What path should the service be registered on? admins
? What database is the service using? SQL
? Which schema definition format do you want to use? Schemas allow to type, validate, secure and populate data TypeBox
(recommended)```


The file admins.schema.ts will be generated with the following minor errors:

```export const adminQueryResolver = resolve<AdminQuery, HookContext<AdminService>>({
  // If there is a user (e.g. with authentication), they are only allowed to see their own data
  id: async (value, user, context) => {
    if (context.params.user) {    <-- user does not exist
      return context.params.user.id   <-- user does not exist
    }

    return value
  }
})```

Replacing the context.params.user with context.params.admin in two places resolves the issue. 

Perhaps the authentication generator should read the service schema id dynamically to improve this?

### LINKED ISSUE #3599: Bug in Codegen of Authentication generation
I stumbled on a bug in the `npx feathers generate authentication` command.

I ran `npx feathers generate authentication`  with the following options:
Image

Pay attention to the value for `What is your authentication service name?`. It is set to `users` and not `user`.

This led to me getting the following errors in `src\services\users\users.schema.ts`
Image

My guess is, that the `user` part in `context.params.user` is hardcoded, but it should be inferred from the answer to the `What is your authentication service name?` question.

This bug can be reproduced by:
- running `npx feathers generate authentication` 
- entering anything other than "user" for the `What is your authentication service name?` question

For completeness' sake, this is the config I used to setup feathers with:
Image

tonycoder-hub:
## Summary
Reject leftover / partial integer parses in authentication codegen params path related to feathersjs/feathers#3599.

## Test plan
- [ ] Existing authentication tests
- [ ] Manual check of codegen params parsing

Signed-off-by: Tony Coder <407243179@qq.com>

- Closes https://github.com/feathersjs/feathers/issues/3599
- Closes https://github.com/feathersjs/feathers/issues/3412


daffl (approved):

daffl:
Great, thank you! This will go out with the next release!

